PR #46, opened 2022-03-14
Add notices page

lpsinger:
Add a Notices page with links to the GCN Classic notice archives. You can filter the archive using an auto-completing search box.

https://user-images.githubusercontent.com/728407/158174801-f04405ab-c72e-433a-a686-3beff8b5a4fb.mov

### DIFF
--- a/app/routes/notices.tsx
+++ b/app/routes/notices.tsx
@@ -1,0 +1,411 @@
+import {
+  Button,
+  CardBody,
+  CardFooter,
+  CardGroup,
+  CardHeader,
+  IconClose,
+  Tag,
+} from '@trussworks/react-uswds'
+import { ReactNode, useEffect, useRef, useState } from 'react'
+import ReactTags, {
+  SuggestionComponentProps,
+  Tag as ReactTag,
+  TagComponentProps,
+} from 'react-tag-autocomplete'
+import { MetaFunction } from 'remix'
+
+export const meta: MetaFunction = () => ({
+  title: 'GCN - Notices',
+})
+
+function NoticeCard({
+  children,
+  name,
+  tags,
+  selectedTags,
+  href,
+}: {
+  children: ReactNode
+  name: string
+  tags: string[]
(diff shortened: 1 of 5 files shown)


app/routes/notices.tsx
```diff
@@ -0,0 +1,412 @@
...
+        >
+          GRBs detected by the GBM instrument on Fermi.
+        </NoticeCard>
+        <NoticeCard
+          name="Fermi LAT"
+          href="https://gcn.gsfc.nasa.gov/fermi_lat_mon_trans.html"
+          tags={['gamma']}
+          selectedTags={tagNames}
+        >
+          GRBs and gamma-ray transients detected by the LAT instrument on Fermi.
```
jracusin, line 194: LAT_mon_trans is actually all non-GRB sources.

app/routes/notices.tsx
```diff
@@ -0,0 +1,412 @@
...
+        </NoticeCard>
+        <NoticeCard
+          name="Swift GRBs"
+          href="https://gcn.gsfc.nasa.gov/swift_grbs.html"
+          tags={['gamma', 'x-ray', 'uv', 'optical']}
+          selectedTags={tagNames}
+        >
+          GRBs detected by the Neil Gehrels Swift Observatory.
```
jracusin, line 202: You spell out the full name of Swift here, but not the full name of other observatories.  We should be consistent, either way.

app/routes/notices.tsx
```diff
@@ -0,0 +1,412 @@
...
+        </NoticeCard>
+        <NoticeCard
+          name="Swift Sub-Threshold"
+          href="https://gcn.gsfc.nasa.gov/swift_sub_sub_archive.html"
```
jracusin, line 206: These are specifically only BAT, not XRT, so only gamma.

lpsinger:
I think that I addressed all of your comments. Ready for another review.

PR #49, opened 2022-04-06
Update to remix 1.3.5

lpsinger:
This allows us to drop our workaround for remix-run/remix#1536.


lpsinger:
Note: we should wait to merge this until after remix 1.3.5 has been released.

PR #56, opened 2022-05-03
Homepage: explain Kafka better with graphics and tables

lpsinger:
On the home page, explain the advantages of Kafka better using graphics to depict data flow and a table to compare features.
Screen Shot 2022-05-03 at 12 59 45


lpsinger:
Merging with verbal approval from @jracusin in Teams.

jracusin:
I like the graphic and the idea behind the table, but could we reword it to not sound disparaging of GCN Classic?  Maybe it could be just changing the No/Yes to X/checkmark?

jracusin:
We also probably need a little primer on "what is kafka".

PR #79, opened 2022-05-25
Remove workaround for remix-run/remix#1828

lpsinger:
Seems to have been fixed upstream.


lpsinger:
All of the files containing this workaround have now been removed.

PR #97, opened 2022-06-09
Handle 403 Forbidden errors, provide redirect

lpsinger:
https://user-images.githubusercontent.com/728407/172897719-b4fdde74-2f99-49e6-a2ed-891034533c96.mov

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -20,14 +20,15 @@
  */
 
 import {
+  Link,
   Links,
   LiveReload,
   Meta,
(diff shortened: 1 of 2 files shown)


lpsinger:
@sergiodxa, this is what I had in mind when I was pinging you on Discord about using catch boundaries to handle 403 errors. Does this make sense?

kiliman:
Looks good to me. Also, that's awesome that you're using Remix at NASA.

PR #101, opened 2022-06-13
Missions

eburnsastro:
Final pull to complete missions pages


app/routes/missions/fermi.md
```diff
@@ -31,19 +31,19 @@ https://fermi.gsfc.nasa.gov/ssc/data/
...
+| `FERMI_GBM_FLT_POS`   | Flight localization, classification                   | ~10 seconds    |
+| `FERMI_GBM_GND_POS`   | Updated ground localization using finer lookup tables | 20–300 seconds |
+| `FERMI_GBM_FIN_POS`   | Final trigger localization                            | 15 minutes     |
+| `FERMI_GBM_SUBTHRESH` | Ground pipeline detected signals                      | 0.5–6 hours    |
+| `FERMI_LAT_POS_INI`   | Onboard LAT detection, initial position               | 2–4 seconds    |
+| `FERMI_LAT_POS_UPD`   | Updated onboard localization with more data           | 2–32 seconds   |
+| `FERMI_LAT_GND`       | Ground localization of onboard trigger                | 8–12 hours     |
+| `FERMI_LAT_OFFLINE`   | Ground-detected GRB localization                      | 8–12 hours     |
+| `FERMI_LAT_TRANS`     | Unknown source transient                              | 1 days         |
+| `FERMI_LAT_MONITOR`   | Known source flare                                    | 1 days         |
```
lpsinger, line 46: Do we want to switch everywhere from SI abbreviations to fully spelled-out units?

app/routes/missions/agile.md
```diff
@@ -0,0 +1,55 @@
...
+
+**GCN Notice Types in GCN Classic and GCN Classic Over Kafka:**
+[Detailed Descriptions and Examples](https://gcn.gsfc.nasa.gov/agile.html)
+
+| Type      | Contents                               | Latency        |
+| --------- | -------------------------------------- | -------------- |
+| `WAKEUP`  | On-board processing                    | ~30 minutes    |
+| `REFINED` | Ground processing                      | 1–2 hours      |
+| `Refined` | Humans in the loop                     | ~2–5 hours     |
+| `MCAL`    | Ground processing of the MCAL triggers | 40–130 minutes |
```
lpsinger, line 40: Use notice type names from https://github.com/lpsinger/pygcn/blob/main/gcn/notice_types.py. Same for all other files.

lpsinger (changes requested):

app/routes/missions/fermi.md
eburnsastro: I've switched a few files so they are all now fully-spelled out units. I think this makes it easier to read without having to think about it. If you agree then my latest push (9a769a9) resolves this as well

lpsinger: It makes sense, but in the future, it would be preferable to separate something like that into a different PR, to keep each PR to a single concern.

app/routes/missions/maxi.md
```diff
@@ -0,0 +1,53 @@
...
+| [Gas Slit Camera (GSC)](https://iss.jaxa.jp/en/kiboexp/ef/maxi/)         | 2 keV–30 keV   | 160°x1.5°     | &leq;1° (statistical + systematic, 90%) |
+
+**GCN Notice Types in GCN Classic and GCN Classic Over Kafka:**
+[Detailed Descriptions and Examples](https://gcn.gsfc.nasa.gov/maxi.html)
+
+| Type           | Contents                             | Latency     |
+| -------------- | ------------------------------------ | ----------- |
+| `MAXI_Unknown` | Signal with no catalogued source     | 200 minutes |
+| `MAXI_Known`   | Signal with a matched catalog source | 200 minutes |
+| `MAXI_Test`    | Test notices                         | n/a         |
```
lpsinger, line 38:
```suggestion
| `MAXI_UNKNOWN` | Signal with no catalogued source     | 200 minutes |
| `MAXI_KNOWN`   | Signal with a matched catalog source | 200 minutes |
| `MAXI_TEST`    | Test notices                         | n/a         |
```

lpsinger (approved):

PR #108, opened 2022-06-22
Use slim size for dev banner

lpsinger:
The full-size site alert was too distracting.

## Before
Screen Shot 2022-06-22 at 13 12 45

## After
Screen Shot 2022-06-22 at 13 12 57


dakota002 (approved):
Beautiful, simple but needed

PR #109, opened 2022-06-22
Add hero to landing page

lpsinger:
Place the getting started link front and center on the lading page, and move most of the text to the "What is GCN?" section in the docs.

Note: we need the "Getting started" UI implemented before this is merged.

## Before
Screen Shot 2022-06-22 at 14 49 07

## After
Screen Shot 2022-06-22 at 14 48 29

### DIFF
--- a/app/routes/docs/index.md
+++ b/app/routes/docs/index.md
@@ -3,13 +3,16 @@ meta:
   title: GCN - About
 ---
 
-# About GCN
+# What is GCN?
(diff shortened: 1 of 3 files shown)


dakota002:
This new layout looks great. Its more eye catching and guides the user more effectively

jracusin:
I like it!  We just need to get the new graphic in there.

PR #125, opened 2022-06-29
Step indicator update

dakota002:
This introduces the step process for the user. Once logged in, they can access these steps from the user page.

### DIFF
--- a/app/components/ClientCredentialForm.tsx
+++ b/app/components/ClientCredentialForm.tsx
@@ -3,13 +3,15 @@ import { useLoaderData } from '@remix-run/react'
 import { Label, TextInput, Dropdown, Button } from '@trussworks/react-uswds'
 import { useState } from 'react'
 import ReCAPTCHA from 'react-google-recaptcha'
+import { getEnvOrDieInProduction } from '~/lib/env'
 import { ClientCredentialVendingMachine } from '~/routes/user/client_credentials.server'
 import type { ClientCredentialData } from './ClientCredentialCard'
 
 export async function loader({ request }: DataFunctionArgs) {
   const machine = await ClientCredentialVendingMachine.create(request)
   const client_credentials = await machine.getClientCredentials()
-  return { client_credentials }
+  const recaptchaSiteKey = getEnvOrDieInProduction('RECAPTCHA_SITE_KEY')
+  return { client_credentials, recaptchaSiteKey }
 }
 
 export default function ClientCredentialForm({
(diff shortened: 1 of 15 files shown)


lpsinger:
Would you please add some screen shots?

lpsinger (changes requested):
I'm a bit confused about how this works. Where is `setActiveSteps` defined?

dakota002:
Screen Shot 2022-06-30 at 12 43 50 PM
Screen Shot 2022-06-30 at 12 44 03 PM
Screen Shot 2022-06-30 at 12 44 39 PM
Screen Shot 2022-06-30 at 12 44 48 PM

dakota002:
Screen Shot 2022-06-30 at 12 44 39 PM

dakota002:
Screen Shot 2022-06-30 at 12 44 48 PM

app/routes/user/streaming_steps/credentials.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const client_credentials = await machine.getClientCredentials()
+  return { client_credentials }
+}
+
+export default function Credentials() {
+  //Data passed between steps
+  const clientData = useClient()
+  // Data loaded for options
+  const { client_credentials } =
+    useLoaderData<Awaited<ReturnType<typeof loader>>>() ?? []
```
lpsinger, line 23: Why would this ever return null?

app/components/ClientCredentialForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+  setItems: React.Dispatch<React.SetStateAction<ClientCredentialData[]>>
+}) {
+  const { client_credentials } =
+    useLoaderData<Awaited<ReturnType<typeof loader>>>() ?? []
+  const [items] = useState<ClientCredentialData[]>(client_credentials)
+  const defaultName = ''
+  const [name, setName] = useState(defaultName)
+  const defaultScope = 'gcn.nasa.gov/kafka-public-consumer'
+  const [scope, setScope] = useState(defaultScope)
+  const siteKey = 'site-key'
```
lpsinger, line 27: Would you please restore the logic from main that gets this from the environment?

lpsinger (changes requested):

app/routes/user/streaming_steps/credentials.tsx
dakota002: It shouldn't, I believe that was from testing. Fixing now

app/components/ClientCredentialForm.tsx
dakota002: Updated!

lpsinger (approved):

lpsinger (changes requested):
Looks great, but needs to be rebased.

lpsinger (approved):
I rebased it.

PR #126, opened 2022-06-30
Show nested checkboxes as a tree rather than an accordion

lpsinger:
This is much more compact.

https://user-images.githubusercontent.com/728407/176735840-5fb321ef-18d6-4b98-b983-b0cd5d57f370.mov


dakota002 (approved):
Looks great!

PR #131, opened 2022-07-01
Quick start: redo lifecycle stuff using hidden inputs, eliminate extra clicks

lpsinger:
Redo the lifecycle management of the quick start steps by saving state from other steps in hidden forms. This is simpler than using React state and it's fast using Remix forms. Party like it's 1998!

Also, eliminate unnecessary clicks from the UX by making the client credentials themselves links rather than radio buttons.

Note, I haven't implemented the new client credential bit yet. So this isn't quite yet ready to merge.

https://user-images.githubusercontent.com/728407/176909257-5f9406aa-bb14-4a82-a945-31956a51f910.mov


lpsinger:
This is ready for review, though there is some duplication of code now between the quick start and the client credentials page, so a refactor will be needed in the near future.

Uploading Screen Recording 2022-07-01 at 12.36.45.mov…

dakota002 (approved):

PR #132, opened 2022-07-01
Fixes delete issue and updates the client credential form to be more …

dakota002:
…usable and less state dependant


dakota002:
Redundant, already fixed in [another PR](https://github.com/tachgsfc/gcn.nasa.gov/pull/133)

PR #143, opened 2022-07-07
Replace GCN diagram with new artwork by Britt Griswold

lpsinger:


lpsinger:
@dakota002, the image's aspect ratio is messed up. Can you fix it please? Maybe some CSS flux incantation.

Screen Shot 2022-07-06 at 21 30 04

PR #164, opened 2022-07-15
Updates the code sample buttons

dakota002:
Use only icons and tooltips to save space. Copy buttons tooltip text is on a 3 second timer:
"Copy" => click => "Copied!" => 3 sec... => "Copy"


lpsinger:
💯  amazing!

PR #171, opened 2022-07-18
Styling updates to make UI more responsive in the missions pages incl…

dakota002:
…uding word-breaks on links, image scaling, and horizontal scrolling on tables

### DIFF
--- a/app/routes/missions.tsx
+++ b/app/routes/missions.tsx
@@ -12,7 +12,7 @@ import { NavLink, Outlet } from '@remix-run/react'
 export default function Missions() {
   return (
     <div className="grid-row grid-gap">
-      <div className="desktop:grid-col-4">
+      <div className="desktop:grid-col-4 ">
         <SideNav
           items={[
             <NavLink key="." to="." end>
@@ -60,7 +60,7 @@ export default function Missions() {
           ]}
         />
       </div>
-      <div className="desktop:grid-col-8 usa-prose">
+      <div className="desktop:grid-col-8 usa-prose grid-col-12">
         <Outlet />
       </div>
     </div>
--- a/app/routes/missions.tsx
+++ b/app/routes/missions.tsx
@@ -12,7 +12,7 @@ import { NavLink, Outlet } from '@remix-run/react'
 export default function Missions() {
   return (
     <div className="grid-row grid-gap">
-      <div className="desktop:grid-col-4 ">
+      <div className="desktop:grid-col-4">
         <SideNav
           items={[
             <NavLink key="." to="." end>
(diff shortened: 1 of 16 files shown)


app/routes/missions.tsx
```diff
@@ -12,7 +12,7 @@ import { NavLink, Outlet } from '@remix-run/react'
 export default function Missions() {
   return (
     <div className="grid-row grid-gap">
-      <div className="desktop:grid-col-4">
+      <div className="desktop:grid-col-4 ">
```
lpsinger, line 15: Unnecessary whitespace change.

lpsinger (changes requested):
What about https://designsystem.digital.gov/components/table/#responsive-stacked-table for the tables?

dakota002:
Fixed the whitespace issue, do we want to use those stacked tables? I don't see any immediate benefit, but I am not opposed to using them

dakota002:
That's doable, but is there any reason we are keeping these tables as markdown? We will need to define the tables as elements anyway to add the uswds utility class thing. Just thinking the files are looking less and less like markdown. Are you cool with them being turned into tsx files instead? They feel repetitive enough that we might be able to break them into components or something

lpsinger:
> That's doable, but is there any reason we are keeping these tables as markdown? We will need to define the tables as elements anyway to add the uswds utility class thing. Just thinking the files are looking less and less like markdown. Are you cool with them being turned into tsx files instead? They feel repetitive enough that we might be able to break them into components or something

These particular files are maintained by scientists who can write Markdown but probably are not comfortable with React.

Note that we can globally set all tables in markdown to be responsive by editing the rehype settings in remix.config.js.

dakota002:
Ah good to know, in trying that approach I think it makes some of the tables look good, albeit somewhat like nutrition fact labels
image
vs 
image

But others look a little off, with the empty cells such as this 
image
vs
image

lpsinger:
😆 

The ones that look funky, have some colspans and rowspans missing.

dakota002:
haha yep playing with some colspans now, is there a rehype-correct way to do colspans?

lpsinger:
It looks like there is a mix of the two table formats. Is that intended?

Screen Shot 2022-07-18 at 16 39 14

lpsinger:
Yeah, the tables are all responsive now, but there is something funny going on with the last row of many of the tables. For example:
Screen Shot 2022-07-19 at 11 03 35

app/routes/missions/fermi.md
```diff
@@ -30,20 +30,19 @@ https://fermi.gsfc.nasa.gov/ssc/data/
 
 **GCN Notice Types in GCN Classic and GCN Classic Over Kafka:**
 [Detailed Descriptions and Examples](https://gcn.gsfc.nasa.gov/fermi.html)
-
-| Type                  | Contents                                              | Latency        |
+| Type | Contents | Latency |
```
lpsinger, line 33: Undo unnecessary whitespace changes here.

lpsinger (changes requested):

lpsinger (changes requested):

lpsinger (approved):

PR #180, opened 2022-07-22
updated gcn-diagram with recent changes from Britt

jracusin:
Screen Shot 2022-07-22 at 4 08 46 PM


jracusin:
Changes include medium/large missions look like Swift and Fermi rather than TESS and Roman.  Also the jet is no longer striped.

lpsinger:
Image size went from 272K to 976K. Would you please ask Britt to rasterize the image to a similar resolution and compression ratio to the jpg we currently use? Point him to the file on the main branch.

jracusin:
Sorry, that was actually my fault when I converted from a png.  It should now be 327kB.

lpsinger:
Did you downsample and convert it from a jpg? We'll get the best results by rasterizing directly from the source Illustrator document.

PR #190, opened 2022-07-26
Fix disclosure icon for webinar announcement banner

lpsinger:
The arrow icon in the "Show more"/"Show less" button for the announcement banner was not updating when the button was toggled.


dakota002 (approved):
Much neater

PR #208, opened 2022-08-02
Email backend

dakota002:
These were tested locally, and all seem to work. I'm not super familiar with the proper use of the arctable instance methods, so let me know if and how I can make improvements.


lpsinger (changes requested):
Looks pretty good, but:

1. [ ] This code needs to populate the inverse table mapping topics to subscribed emails.
2. [x] I would prefer for the backend class to know only about topics and not about notice types and formats.

app/lib/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+export function mapFormatAndNoticeTypesToTopics(
+  noticeFormat: string,
+  noticeTypes: string[]
+) {
+  return noticeTypes.map(
+    (noticeType) => `gcn.classic.${noticeFormat}.${noticeType}`
+  )
+}
```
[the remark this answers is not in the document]
dakota002: Implemented in latest commit

app/lib/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function mapTopicsToFormatAndNoticeTypes(topics: string[]): {
```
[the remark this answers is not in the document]
dakota002: Implemented in latest commit

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,154 @@
...
+import {
+  mapFormatAndNoticeTypesToTopics,
+  mapTopicsToFormatAndNoticeTypes,
+} from '~/lib/utils'
+
+export type EmailNotification = {
+  name: string
+  recipient: string
+  format: string
+  noticeTypes: string[]
```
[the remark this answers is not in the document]
dakota002: I dont believe these would be known to the db, since they are not included in the put, but I have updated the file to have a view model version that includes these fields for the purpose of the eventual Edit page, and updated the methods to accept and return the appropriate models based on function

lpsinger (changes requested):
Still needs the materialized inverse table...

dakota002:
The latest works for successfully doing crud stuff to the materialized views, but I think the way it is structured may cause issues when trying to query the email_notification_subscription table by topic, since the UUID is the partition key in this case. I can switch these around, but it will probably just mean that previous topics will need to be included the update function and the delete

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+    })
+
+    // Materialized view
+    for (const topic of notification.topics) {
+      await db.email_notification_subscription.put({
+        uuid,
+        topic,
+        recipient: notification.recipient,
+      })
+    }
```
lpsinger, line 72: Use `await Promise.all()` to submit all of these SDK requests in parallel?

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+export type EmailNotification = {
+  name: string
+  recipient: string
+  created: number
+  active: boolean
+  uuid?: string
+  topics: string[]
+}
+
+// view model
```
lpsinger, line 22: What do you mean by view model?

```suggestion
// inverse table
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+    for (const notice of emailNotifications) {
+      let formats = []
+      notice.noticeTypes = []
+      for (const topic of notice.topics) {
+        let mappedData = mapTopicsToFormatAndNoticeType(topic)
+        formats.push(mappedData.noticeFormat)
+        notice.noticeTypes.push(mappedData.noticeType)
+      }
+      notice.format = [...new Set(formats)][0]
+    }
```
lpsinger, line 103: I don't know about you, but I prefer `array.map(...)` to `for (... of ...)` because it avoids having to dynamically grow the result and it lets the TypeScript compiler do more type inference for you.

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+    const emailNotifications = results.Items as EmailNotificationVM[]
+    for (const notice of emailNotifications) {
+      let formats = []
+      notice.noticeTypes = []
+      for (const topic of notice.topics) {
+        let mappedData = mapTopicsToFormatAndNoticeType(topic)
+        formats.push(mappedData.noticeFormat)
+        notice.noticeTypes.push(mappedData.noticeType)
+      }
+      notice.format = [...new Set(formats)][0]
```
lpsinger, line 102: I think that we can just let it be an error if there are multiple formats.

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+      uuid,
+      ...notification,
+    }
+  }
+
+  // Update
+  async updateEmailNotification(email_notification: EmailNotification) {
+    if (!email_notification.uuid) return null
+    const db = await tables()
+    db.email_notification.update(
```
lpsinger, line 134:
```suggestion
    await db.email_notification.update(
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+          ':name': email_notification.name,
+          ':recipient': email_notification.recipient,
+          ':topics': email_notification.topics,
+          ':active': email_notification.active,
+        },
+      },
+      (err, data) => {
+        if (err) console.log(err)
+        else console.log(data)
+      }
```
lpsinger, line 155: Use try/catch and async/await rather than error callbacks.

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,208 @@
...
+  name: string
+  recipient: string
+  created: number
+  active: boolean
+  uuid?: string
+  topics: string[]
+}
+
+// view model
+export type EmailNotificationVM = EmailNotification & {
```
lpsinger, line 23: Prefer interface over type intersections (see [TypeScript wiki](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections)).

lpsinger (changes requested):
Looks good. General comments:

1. [x] Use `await Promise.all()` to parallelize the AWS SDK calls.
2. [ ] Prefer `array.map()` over `for (... of ... )` due to type inference.

lpsinger:
> The latest works for successfully doing crud stuff to the materialized views, but I think the way it is structured may cause issues when trying to query the email_notification_subscription table by topic, since the UUID is the partition key in this case. I can switch these around, but it will probably just mean that previous topics will need to be included the update function and the delete

Hmm, you're right, the partition key needs to be the topic. Maybe we should try out the 2-table-and-index method:

```
@tables
email_subscription
  sub *String
  uuid **String
  # other fields: created, email, name

email_subscription_topic
  uuid *String
  # other fields: email, topic

@tables-indexes
email_subscription_topic
  topic *String
```

dakota002:
Looks like we had the same thought, I added that in the latest commit along with a name, which would allow usage like:
  async geAllSubsWithSpcificTopic(topic:string) {
    const db = await tables()
    const results = await db.email_notification_subscription.query({
      IndexName:'byTopic',
      KeyConditionExpression:'topic = :topic',
      ExpressionAttributeValues:{
        ':topic' : topic
      },
      ExpressionAttributeNames:{
        '#uuid':'uuid'
      },
      ProjectionExpression:'topic, recipient, #uuid'
    })
    return results.Items
  }

lpsinger (changes requested):
See unresolved comments above.

app/routes/user/email_notifications.server.ts
dakota002: This is probably a less applicable term in the context of what we are going, but the idea is that the View Model is what the frontend works with; the model that the 'view' page is using. 

In this case, the fields like format and notice type are not 'real' in the sense of how the tables are storing data, but are needed as distinct for the UI.

app/routes/user/email_notifications.server.ts
dakota002: I intentionally left the await out because it gives this message:
image

lpsinger: That's because you have a callback. Try removing the callback.

dakota002: That did it!

app/routes/user/email_notifications.server.ts
dakota002: Updated in latest

app/routes/user/email_notifications.server.ts
dakota002: I have removed this altogether, it was mainly just for debugging

app/routes/user/email_notifications.server.ts
dakota002: In the latest commit I am just assuming there will be one type. I just take the first

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+    })
+    const items = results.Items as EmailNotification[]
+    const emailNotifications: EmailNotificationVM[] = items.map(
+      (notification) => {
+        return {
+          format: topicToFormatAndNoticeType(notification.topics[0])
+            .noticeFormat,
+          noticeTypes: notification.topics.map((topic) => {
+            return topicToFormatAndNoticeType(topic).noticeType
+          }),
```
lpsinger, line 104:
```suggestion
          noticeTypes: notification.topics.map((topic) => (
            topicToFormatAndNoticeType(topic).noticeType
          )),
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+  async getEmailNotification(uuid: string): Promise<EmailNotificationVM> {
+    const db = await tables()
+    const item = (await db.email_notification.get({
+      sub: this.#sub,
+      uuid,
+    })) as ({ sub: string } & EmailNotificationVM) | null
+    if (!item) throw new Response(null, { status: 404 })
+    item.noticeTypes = item.topics.map((topic) => {
+      return topicToFormatAndNoticeType(topic).noticeType
+    })
```
lpsinger, line 127:
```suggestion
    item.noticeTypes = item.topics.map((topic) => (
      topicToFormatAndNoticeType(topic).noticeType
    ))
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+      active: true,
+      recipient: notification.recipient,
+    })
+    const subscriptionPromises = notification.topics.map((topic) => {
+      return db.email_notification_subscription.put({
+        uuid,
+        topic,
+        recipient: notification.recipient,
+      })
+    })
```
lpsinger, line 71:
```suggestion
    const subscriptionPromises = notification.topics.map((topic) => (
      db.email_notification_subscription.put({
        uuid,
        topic,
        recipient: notification.recipient,
      ))
    })
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+        ':uuid': uuid,
+      },
+    })
+    await Promise.all(
+      subscriptions.Items.map((sub) => {
+        return db.email_notification_subscription.delete({
+          uuid: sub.uuid,
+          topic: sub.topic,
+        })
+      })
```
lpsinger, line 213:
```suggestion
      subscriptions.Items.map((sub) => (
        db.email_notification_subscription.delete({
          uuid: sub.uuid,
          topic: sub.topic,
        ))
      })
```

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+    )
+    if (email_notification.active) {
+      await Promise.all(
+        email_notification.topics.map((topic) => {
+          return db.email_notification_subscription.put({
+            uuid: email_notification.uuid,
+            topic,
+            recipient: email_notification.recipient,
+          })
+        })
```
lpsinger, line 184:
```suggestion
        email_notification.topics.map((topic) => (
          db.email_notification_subscription.put({
            uuid: email_notification.uuid,
            topic,
            recipient: email_notification.recipient,
          ))
        })
```

lpsinger (changes requested):
Eliminate some unnecessary `return` statements in one-line anonymous functions.

app/routes/user/email_notifications.server.ts
dakota002: fixed in commit, the end parenthesis is mismatched

app/routes/user/email_notifications.server.ts
```diff
@@ -0,0 +1,216 @@
...
+          noticeTypes: notification.topics.map(
+            (topic) => topicToFormatAndNoticeType(topic).noticeType
+          ),
+          name: notification.name,
+          recipient: notification.recipient,
+          created: notification.created,
+          active: notification.active,
+          topics: notification.topics,
+        }
+      }
```
lpsinger, line 111:
```suggestion
      (notification) => (
        {
          format: topicToFormatAndNoticeType(notification.topics[0])
            .noticeFormat,
          noticeTypes: notification.topics.map(
            (topic) => topicToFormatAndNoticeType(topic).noticeType
          ),
          name: notification.name,
          recipient: notification.recipient,
          created: notification.created,
          active: notification.active,
          topics: notification.topics,
        }
      )
```

lpsinger (changes requested):

lpsinger (approved):

lpsinger:
Looks good to me. Merge whenever you are ready.

PR #235, opened 2022-08-23
Adds a direct reference to crypto

dakota002:
This works locally


app/routes/user/email_notifications.server.ts
```diff
@@ -11,6 +11,7 @@ import type { SendEmailCommandInput } from '@aws-sdk/client-ses'
 import { SendEmailCommand, SESClient } from '@aws-sdk/client-ses'
 import { topicToFormatAndNoticeType } from '~/lib/utils'
 import { getUser } from '~/routes/__auth/user.server'
+import crypto from 'node:crypto'
```
lpsinger, line 14: Let's try:

```suggestion
import crypto from 'crypto'
```

lpsinger (changes requested):

PR #239, opened 2022-08-24
Removes an unused css breakpoint to fix the overflow issue

dakota002:
Fixes this overflow issue:
image


lpsinger:
Is there no way to do this with USWDS utility classes?

dakota002:
There is, however they are bound to some parts in files we have ignored and which, I believe, are automatically built by gulp. The _uswds-theme-utilities.scss file has a width setting where responsive needs to be set to true

lpsinger:
Which utility classes would do this?

dakota002:
This one should be all we need: $width-settings

PR #247, opened 2022-08-25
Test confirmation

dakota002:


app/routes/user/email_notifications.server.ts
```diff
@@ -252,6 +252,6 @@ export class EmailNotificationVendingMachine {
       },
     })
 
-    await client.send(command)
+    return await client.send(command)
```
lpsinger, line 255: I think that this will just raise an exception if it fails, right? No need to check the status code.

lpsinger (changes requested):

PR #251, opened 2022-08-26
Sets the default email address to be the current user when creating a…

dakota002:
… new email notification


lpsinger (commented):
Does this populate the default email address visibly, in the form?

dakota002:
Yep!

PR #272, opened 2022-08-30
Adds a counter and total counts on the email alerts and noticetypeche…

dakota002:
…ckbox

### DIFF
--- a/app/components/NestedCheckboxes.tsx
+++ b/app/components/NestedCheckboxes.tsx
@@ -15,6 +15,7 @@ type CheckboxProps = CheckboxArgs[0]
 interface NestedCheckboxProps extends CheckboxProps {
   nodes: CheckboxProps[]
   link?: string
+  childoncheckhandler?: (arg: HTMLInputElement) => void
 }
 
 function allTrue(values: boolean[]) {
@@ -28,6 +29,7 @@ function allSame([first, ...rest]: any[]) {
 function NestedCheckboxNode({
   nodes,
   link,
+  childoncheckhandler = () => null,
   ...topLevelNodeProps
 }: NestedCheckboxProps) {
   const [expanded, setExpanded] = useState(false)
@@ -48,7 +50,14 @@ function NestedCheckboxNode({
   }
 
   useEffect(updateParent)
-
+  useEffect(() => {
+    for (const ref in childRefs.current) {
+      let childRef = childRefs.current[ref]
+      if (childRef != null) {
+        childoncheckhandler(childRef)
+      }
+    }
+  })
   return (
     <li
       role="treeitem"
@@ -105,11 +114,23 @@ function NestedCheckboxNode({
   )
 }
 
-export function NestedCheckboxes({ nodes }: { nodes: NestedCheckboxProps[] }) {
+interface NestedCheckboxesProps {
+  nodes: NestedCheckboxProps[]
+  childoncheckhandler?: (arg: HTMLInputElement) => void
+}
+
+export function NestedCheckboxes({
+  nodes,
+  childoncheckhandler,
+}: NestedCheckboxesProps) {
   return (
     <ul role="tree" aria-multiselectable>
       {nodes.map((node, index) => (
-        <NestedCheckboxNode key={index} {...node} />
+        <NestedCheckboxNode
+          key={index}
+          {...node}
+          childoncheckhandler={childoncheckhandler}
+        />
       ))}
     </ul>
   )
--- a/app/components/NoticeTypeCheckboxes.tsx
+++ b/app/components/NoticeTypeCheckboxes.tsx
@@ -6,6 +6,8 @@
  * SPDX-License-Identifier: NASA-1.3
  */
 
+import { useState } from 'react'
+import { triggerRate } from '~/lib/prometheus'
 import { NestedCheckboxes } from './NestedCheckboxes'
 
 const NoticeTypes = {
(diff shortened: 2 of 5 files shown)


app/components/NestedCheckboxes.tsx
```diff
@@ -13,6 +13,7 @@ type CheckboxArgs = Parameters<typeof Checkbox>
 type CheckboxProps = CheckboxArgs[0]
 interface NestedCheckboxProps extends CheckboxProps {
   nodes: CheckboxProps[]
+  counterhandler?: (arg: any) => void
```
lpsinger, line 16: Is there a way to move this counter handler stuff to NoticeTypeCheckboxes?

dakota002: The function is defined on the NoticeTypeCheckboxes component level, and passed through the interface. I think this is necessary, since we are trying to have it act in response to a checkbox being clicked, a couple levels down component wise. I could try to make the interface's property more abstract, maybe something like 'childoncheckhandler' or something, if we think there is a need for that, but otherwise, I don't think the lines specifically in this file can be moved out.

dakota002:
Making this a draft, noticed a bug on load that doesnt update the numbers correctly

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,161 @@ const NoticeTypes = {
...
+  selectedFormat = 'text',
+  showTotalSelected = false,
+}: NoticeTypeCheckboxProps) {
+  const [metrics, setMetrics] = useState<FormattedData>()
+  const [userSelected, setUserSelected] = useState<{ [key: string]: boolean }>(
+    {}
+  )
+  const [selectedCounter, setSelectedCounter] = useState(0)
+  const [alertEstimate, setAlertEstimate] = useState(0)
+  useEffect(() => {
```
lpsinger, line 194: The `useEffect()` hook should not be necessary any more because the trigger data need not be fetched asynchronously. Move this out of the hook so that the component can be fully pre-rendered.

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,161 @@ const NoticeTypes = {
...
-}: NoticeTypeCheckboxeProps) {
+  selectedFormat = 'text',
+  showTotalSelected = false,
+}: NoticeTypeCheckboxProps) {
+  const [metrics, setMetrics] = useState<FormattedData>()
+  const [userSelected, setUserSelected] = useState<{ [key: string]: boolean }>(
+    {}
+  )
+  const [selectedCounter, setSelectedCounter] = useState(0)
+  const [alertEstimate, setAlertEstimate] = useState(0)
```
lpsinger, line 193: Initial state may be incorrect if any of the checkboxes are initially checked.

app/components/NestedCheckboxes.tsx
```diff
@@ -13,6 +13,7 @@ type CheckboxArgs = Parameters<typeof Checkbox>
 type CheckboxProps = CheckboxArgs[0]
 interface NestedCheckboxProps extends CheckboxProps {
   nodes: CheckboxProps[]
+  childoncheckhandler?: (arg: any) => void
```
lpsinger, line 16: Specify the actual type of `arg` (a React ref).

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,161 @@ const NoticeTypes = {
...
+          })),
+        voevent: response.result
+          .filter(
+            (item: { metric: { topic: string } }) =>
+              item.metric.topic.indexOf('.voevent.') != -1
+          )
+          .map((item: { metric: { topic: any }; values: string | any[] }) => ({
+            topic: item.metric.topic.split('.')[3],
+            latestRate: Math.ceil(item.values[item.values.length - 1][1]),
+          })),
```
lpsinger, line 223: Refactor to avoid repetition...

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,161 @@ const NoticeTypes = {
...
+      setAlertEstimate(getSum(metrics))
+    }
+  }
+
+  function getSum(metrics: FormattedData) {
+    return metrics[selectedFormat]
+      .filter(
+        (item) =>
+          userSelected[item.topic] !== undefined &&
+          userSelected[item.topic] == true
```
lpsinger, line 250: Shouldn't that be equivalent to this?

```suggestion
          userSelected[item.topic]
```

or this?

```suggestion
          !!userSelected[item.topic]
```

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,161 @@ const NoticeTypes = {
...
+          })),
+        }))}
+        childoncheckhandler={counterfunction}
+      />
+      {showTotalSelected ? (
+        <div>
+          {selectedCounter} selected alerts for an estimated total of{' '}
+          {alertEstimate} alerts per day
+        </div>
+      ) : null}
```
lpsinger, line 316: 1. [ ] Display "0 selected alerts for an estimated total of 0 alerts per day" if nothing is selected
2. [ ] Display correct singular/plural form

lpsinger (changes requested):

app/lib/prometheus.ts
```diff
@@ -0,0 +1,1514 @@
...
+          topic: 'gcn.classic.voevent.UNKNOWN',
+        },
+        values: [
+          [1661953926.298, '7.002431399791594'],
+          [1661953927.298, '7.002431399791594'],
+        ],
+      },
+    ],
+  }
+}
```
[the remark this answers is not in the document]
dakota002: This is not an extra bracket, it is the closing bracket of the function

app/components/NoticeTypeCheckboxes.tsx
dakota002: I recently worked through this, have it working correctly in [this](https://github.com/nasa-gcn/gcn.nasa.gov/pull/272/commits/15cbefd151a9fc16a570337f0a9edcb07f0b3918) commit

app/components/NoticeTypeCheckboxes.tsx
dakota002: Can do, will be moved in next commit

app/components/NestedCheckboxes.tsx
```diff
@@ -25,6 +26,9 @@ function allSame([first, ...rest]: any[]) {
 
 function NestedCheckboxNode({
   nodes,
+  childoncheckhandler = () => {
+    return null
+  },
```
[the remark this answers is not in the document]
dakota002: Implemented this in latest commit

app/components/NestedCheckboxes.tsx
dakota002: Updated

app/components/NoticeTypeCheckboxes.tsx
dakota002: refactored into getMetricsByFormat function

app/components/NoticeTypeCheckboxes.tsx
dakota002: yup! updated in latest!

app/components/NoticeTypeCheckboxes.tsx
```diff
@@ -157,29 +159,137 @@ const NoticeTypes = {
...
+type FormattedData = {
+  text: DataRow[]
+  binary: DataRow[]
+  voevent: DataRow[]
+}
+
+interface NoticeTypeCheckboxProps {
   defaultSelected?: string[]
+  selectedFormat?: 'binary' | 'text' | 'voevent'
+  showTotalSelected?: boolean
```
lpsinger, line 176: I don't think that there is a situation where we wouldn't want to show the total selected.

lpsinger (changes requested):

app/components/NoticeTypeCheckboxes.tsx
dakota002: Do we want to show it in the quickstart? I added the option so that it could be off for that section, but I can remove the option and display it there too

theme/custom.scss
```diff
@@ -521,3 +521,23 @@ a {
...
+
+.pill-med-rate {
+  border-color: color($theme-color-warning);
+  background-color: color($theme-color-warning-lighter);
+}
+
+.pill-high-rate {
+  border-color: color($theme-color-error);
+  background-color: color($theme-color-error-light);
+}
```
lpsinger: Dead code:

```suggestion
```

lpsinger (approved):

PR #276, opened 2022-09-02
Adds link to unsubscribe from old form

dakota002:


jracusin (approved):

lpsinger:
I think that this needs more explanation.

jracusin:
Ok, how about:
Signing up here does not edit prior subscriptions to GCN Classic.  When you would like to unsubscribe from GCN Classic, please contact us.

PR #277, opened 2022-09-02
Server side validation

dakota002:


lpsinger (changes requested):
Why not here where all of the other validation is done? https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/user/email_notifications.server.ts#L57-L66

dakota002:
image
 wait is this already just being handled then?

lpsinger:
Yeah it was. Oops!

PR #307, opened 2022-09-15
Peer endorsements

dakota002:


dakota002:
A new page under users dropdown to manage permissions
image


New users have an 'unrequested' status by default
image

Clicking the request endorsement button opens a modal
image


Pending:
image


For an existing approved user, the requests from others for endorsement show up here
image


I assumed 3 options, Approve, Reject, and Reject and Report. The main case for the distinction between the 'Reject' and 'Reject and Report' is to be able to allow us to potentially track spam. This may depend on how we go about trying to enforce restrictions.

@lpsinger and @jracusin  Let me know what you think of this so far.

lpsinger:
We merged all these changes already, in separate and smaller PRs.

PR #364, opened 2022-10-24
Peer Endorsement Breakout PR 1, Endorsement Backend

dakota002:


app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,236 @@
...
+    )
+    if (!potentialEndorsers) return ''
+
+    const userSub = potentialEndorsers.Attributes?.find(
+      (attr) => attr.Name == 'sub'
+    )?.Value
+    if (!userSub) return ''
+    return userSub
+  }
+}
```
lpsinger, line 236: There may, in general, be multiple matching subs for a given email address.

lpsinger (changes requested):
- [ ] In general, there is a one-to-many mapping between email addresses and subs for potential endorsers. How do we deal with this?
- [x] This backend business logic not to concern itself with parsing form data.
- [ ] We need FIXME comments to add checks of the user's scopes before we approve endorsements.
- [ ] Function names are a bit confusing. For example, what does `updateEndorsementRequestedEndorser` do?

app/routes/user/endorsements.server.ts
dakota002: True, I am going to remove this from this layer. If we want to add a dropdown that fills with options as the user enters emails, maybe we could show multiple results in the dropdown along with the idp? like the dropdown text could read:
dakota@example.com (Facebook)
dakota@example.com (Google)
dakota@example.com (Email)
What are your thoughts?

app.arc
```diff
@@ -25,11 +25,19 @@ email_notification_subscription
   uuid *String
   topic **String
 
+circular_permission_requests
```
lpsinger, line 28: Terser:

```suggestion
circular_endorsements
```

app.arc
```diff
@@ -25,11 +25,19 @@ email_notification_subscription
...
+
 @tables-indexes
 email_notification_subscription
   topic *String
   name byTopic
-  
+
+circular_permission_requests
+  endorserSub *String
+  name byEndorserSub
```
lpsinger, line 39:
```suggestion
  name circularEndorsementsByEndorserSub
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,151 @@
...
+        status: 403,
+      })
+    return new this(user.sub)
+  }
+
+  // FIXME: Requests should probably only be created with endorsers as users who are in the submitters group
+  async createEndorsementRequest(request: EndorsementRequest) {
+    const db = await tables()
+    await db.circular_permission_requests.put(request)
+  }
```
lpsinger, line 45: Add Cognito group membership check based on access token claims here.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  // Allows the current user to change the requested endorser on their previously made request
+  async updateRequestedEndorser(
+    existingEndorserSub: string,
+    newEndorserSub: string,
+    newEndorserEmail: string
+  ) {
+    const db = await tables()
```
lpsinger, line 53: Check that:

- [ ] The endorsement exists
- [ ] The endorsement is pending

lpsinger (changes requested):

app/routes/user/endorsements.server.ts
dakota002: I added a group check, but its checking against other users. I dont use the claims though, let me know what you think

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,151 @@
...
+        ':endorserSub': this.#sub,
+        ':status': 'pending',
+      },
+      ProjectionExpression:
+        '#requestorSub, #requestorEmail, #endorserSub, #endorserEmail, #status, #created',
+    })
+
+    return result.Items as EndorsementRequest[]
+  }
+}
```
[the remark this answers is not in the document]
dakota002: I squashed them into a single method and use the role to construct the queryParams

app/lib/utils.ts
```diff
@@ -31,3 +36,13 @@ export function getFormDataString(formData: FormData, key: string) {
...
+
+export async function getSubmitterUsersFromCognito() {
+  const command = new ListUsersInGroupCommand({
+    GroupName: 'gcn.nasa.gov/circular-submitter',
+    UserPoolId: process.env.COGNITO_USER_POOL_ID,
+  })
+  const cognitoIdentityProviderClient = new CognitoIdentityProviderClient({})
+  const result = await cognitoIdentityProviderClient.send(command)
+  return result.Users
+}
```
lpsinger, line 48: Move this to endorsements.server.ts

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  private constructor(sub: string) {
+    this.#sub = sub
+  }
+
+  static async create(request: Request) {
+    const user = await getUser(request)
+    if (!user)
+      throw new Response('not signed in', {
+        statusText: 'not signed in',
```
lpsinger, line 36: I don't think you have to supply both `status` and `statusText`:

```suggestion
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+        status: 403,
+      })
+    return new this(user.sub)
+  }
+
+  async createEndorsementRequest(request: EndorsementRequest) {
+    // Checking endorser sub:
+    if (!this.requestedEndorserIsValid(request.endorserSub))
+      throw new Response('Requested Endorser is not a verified submitter', {
+        statusText: 'Bad Request',
```
lpsinger, line 46:
```suggestion
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+        status: 400,
+      })
+    return validUsers?.some(
+      (user) =>
+        user.Attributes?.find((attr) => attr.Name == 'sub')?.Value ==
+        endorserSub
+    )
+  }
+
+  validateStateAndExistanceOfRequest(existingEndorsement: EndorsementRequest) {
```
lpsinger, line 199: Make this function private.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+    if (!validUsers)
+      throw new Response('No submitter users exist yet', {
+        statusText: 'Bad Request',
+        status: 400,
+      })
+    return validUsers?.some(
+      (user) =>
+        user.Attributes?.find((attr) => attr.Name == 'sub')?.Value ==
+        endorserSub
+    )
```
lpsinger, line 196: Use Cognito's [ListUsers](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListUsers.html) API with a Filter to find the user with the matching sub, then check that they belong to the submitters group. That way you don't get the entire Cognito database back!

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+    const user = await getUser(request)
+    if (!user)
+      throw new Response('not signed in', {
+        statusText: 'not signed in',
+        status: 403,
+      })
+    return new this(user.sub)
+  }
+
+  async createEndorsementRequest(request: EndorsementRequest) {
```
lpsinger, line 42: How about we just take the requested `sub` as the argument? That way we don't have to validate that the status is pending and that the sub matches the current user.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+    // Checking endorser sub:
+    if (!this.requestedEndorserIsValid(request.endorserSub))
+      throw new Response('Requested Endorser is not a verified submitter', {
+        statusText: 'Bad Request',
+        status: 400,
+      })
+    const db = await tables()
```
lpsinger, line 49: We need to check that if an endorsement request with matching subs already exists, then it is in the pending state. That way, the user cannot overwrite a request that was already made, unless doing so would be idempotent.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,214 @@
...
+      },
+    })
+  }
+
+  // Allows the current user to update the status of a request they have been listed on as an endorser
+  async updateEndorsementRequestStatus(
+    status: EndorsementStates,
+    requestorSub: string
+  ) {
+    if (!this.requestedEndorserIsValid(this.#sub))
```
lpsinger, line 91: Please check if the user is authorized by looking at the group claims in their access token.

lpsinger (changes requested):
Overall, needs more validation of requests against existing DynamoDB records.

app/routes/user/endorsements.server.ts
dakota002: I don't think thats quite right for this case, the ListUsers method doesnt return info about the groups even when you pass the filter, just an object like the following:
```
{
  Attributes: [
    { Name: 'sub', Value: 'sub-string' },
    {
      Name: 'identities',
      Value: '[{"userId":"stuff","providerName":"stuff","providerType":"type","issuer":null,"primary":true,"dateCreated":DateTime}]'
    },
    { Name: 'email_verified', Value: 'false' },
    { Name: 'email', Value: 'example@example.com' }
  ],
  Enabled: true,
  MFAOptions: undefined,
  UserCreateDate: DateTime,
  UserLastModifiedDate: DateTime,
  UserStatus: 'EXTERNAL_PROVIDER',
  Username: 'usernamestring'
}
```
getSubmitterUsersFromCognito is just using the ListUsersInGroup. 

The docs also say:
You can only search for the following standard attributes:

username (case-sensitive)

email

phone_number

name

given_name

family_name

preferred_username

cognito:user_status (called Status in the Console) (case-insensitive)

status (called Enabled in the Console) (case-sensitive)

sub

lpsinger: Right. So use ListUsers to find the user by sub, then AdminListGroupsForUser to get the user's groups. That way you avoid ever having a huge API response that contains the entire Cognito user database.

dakota002: Ahhh okay that makes sense. Going to implement it now

lpsinger: Also, remember to add those actions to the arc plugin that grants the Lambda the IAM permissions to execute those API calls.

app/lib/utils.ts
lpsinger: Still here in utils...

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,212 @@
...
+export type EndorsementRequest = {
+  endorserEmail: string
+  endorserSub: string
+  requestorEmail: string
+  requestorSub: string
+  created: number
+  status: EndorsementStates
+}
+
+export type EndorsementStates = 'approved' | 'pending' | 'rejected' | 'reported'
```
lpsinger, line 23:
```suggestion
export type EndorsementState = 'approved' | 'pending' | 'rejected' | 'reported'
```

app/routes/user/endorsements.server.ts
dakota002: Yep, pushed an empty commit, should be addressed

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,314 @@
...
+    const command = new ListUsersInGroupCommand({
+      GroupName: 'gcn.nasa.gov/circular-submitter',
+      UserPoolId: process.env.COGNITO_USER_POOL_ID,
+    })
+    const cognitoIdentityProviderClient = new CognitoIdentityProviderClient({})
+    const result = await cognitoIdentityProviderClient.send(command)
+    return result.Users
+  }
+
+  private async buildNewRequestFromEndorserSub(
```
lpsinger, line 277:
```suggestion
  async #buildNewRequestFromEndorserSub(
```

lpsinger (changes requested):
I still have the feeling that there is some missing validation. Would you please add JSDocs for the public methods, explaining the access checks and validation that each method does?

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "include": ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
   "compilerOptions": {
-    "lib": ["DOM", "DOM.Iterable", "ES2019"],
+    "lib": ["DOM", "DOM.Iterable", "ES2019", "ES2021.String"],
```
lpsinger, line 4: Will this affect both the browser build and the server build? Will it be polyfilled correctly for the browser?

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,313 @@
...
+  static async create(request: Request) {
+    const user = await getUser(request)
+    if (!user)
+      throw new Response('not signed in', {
+        status: 403,
+      })
+    return new this(user.sub, user.email, user.groups)
+  }
+
+  async createEndorsementRequest(endorserSub: string) {
```
lpsinger, line 52: I'm looking for comments like this:

```suggestion
  /**
   * Request a pending endorsement.
   *
   * Throws an HTTP error if:
   *   - The current user already belongs to the GCN Circulars submitters group
   *   - User `endorserSub` does not exist
   *   - User `endorserSub` is not in the GCN Circulars submitters group
   *   - There is an existing endorsement with the same requester and submitter sub but the state is not `pending`
   */
  async createEndorsementRequest(endorserSub: string) {
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,352 @@
...
+   * the provided endorserSub.
+   * @param endorserSub - the sub of the listed endorser on the existing request
+   */
+  async deleteEndorsementRequest(endorserSub: string) {
+    const db = await tables()
+    await db.circular_endorsements.delete({
+      requestorSub: this.#sub,
+      endorserSub,
+    })
+  }
```
lpsinger, line 182: Only permit deleting "pending" requests.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,352 @@
...
+    status: EndorsementState,
+    requestorSub: string
+  ) {
+    if (
+      this.#currentUserGroups.indexOf('gcn.nasa.gov/circular-submitter') == -1
+    )
+      throw new Response(
+        'The user is not a verified submitter, and therefore can not approve requests',
+        {
+          statusText: 'Bad Request',
```
lpsinger, line 142:
```suggestion
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,352 @@
...
+   *
+   * @param endorserSub
+   * @returns true if all checks succeed
+   */
+  async requestedEndorserIsValid(endorserSub: string): Promise<boolean> {
+    if (endorserSub === this.#sub)
+      throw new Response(
+        'Users cannot request themselves as the endorser of their own request',
+        {
+          statusText: 'Bad Request',
```
lpsinger, line 242:
```suggestion
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,352 @@
...
+      await cognitoIdentityProviderClient.send(
+        new ListUsersCommand({
+          UserPoolId: process.env.COGNITO_USER_POOL_ID,
+          Filter: `sub = "${escapedEndorserString}"`,
+        })
+      )
+    )?.Users?.[0]
+    if (!user)
+      throw new Response('Requested user does not exist', {
+        statusText: 'Bad Request',
```
lpsinger, line 259:
```suggestion
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,292 @@
...
+    const result = await cognitoIdentityProviderClient.send(command)
+    return result.Users
+  }
+
+  async #buildNewRequestFromEndorserSub(
+    endorserSub: string
+  ): Promise<EndorsementRequest> {
+    const getEndorserDataCommand = new ListUsersCommand({
+      UserPoolId: process.env.COGNITO_USER_POOL_ID,
+      Filter: `sub = "${endorserSub}"`,
```
lpsinger, line 263: endorserSub must be escaped. Isn't this code duplicated?

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,292 @@
...
+      endorserSub: endorserSub,
+    })) as EndorsementRequest
+
+    if (existingEndorsement)
+      throw new Response(
+        'An endorsement already exists for the selected endorser',
+        {
+          status: 400,
+        }
+      )
```
lpsinger, line 82: Only an error if the existing endorsement's status is not pending.

Check this using a ConditionExpression.

app/routes/user/endorsements.server.ts
lpsinger: Check with a ConditionExpression

lpsinger (changes requested):

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,292 @@
...
+    const getGroupsCommand = new AdminListGroupsForUserCommand({
+      Username: user.Username,
+      UserPoolId: process.env.COGNITO_USER_POOL_ID,
+    })
+    const groupsResponse = await cognitoIdentityProviderClient.send(
+      getGroupsCommand
+    )
+
+    if (
+      !groupsResponse.Groups ||
```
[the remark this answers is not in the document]
dakota002: Yes, that is part of the command
image

app/routes/user/endorsements.server.ts
dakota002: will be handled in related change from one of your comments above

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,292 @@
...
+      !groupsResponse.Groups.find(
+        (x) => x.GroupName == 'gcn.nasa.gov/circular-submitter'
+      )
+    )
+      throw new Response('User is not in the submitters group', {
+        statusText: 'Bad Request',
+        status: 400,
+      })
+
+    return true
```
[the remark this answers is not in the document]
dakota002: Resolved in that the method it belongs to no longer exists

dakota002:
I updated the response objects, the statusText property is what can be read out, and the body can just be null

app/routes/user/endorsements.server.ts
dakota002: Do you have docs or examples of what you are talking about using?

lpsinger: `(await tables())._db` ...

app/routes/user/endorsements.server.ts
lpsinger: I think that statusText is normally something like "Forbidden" or "Gateway Error". It's the canonical text representation of the HTTP status code. I don't think it's typically used to report the detail of the error.

Unless I'm mistaken, please return the explanatory text in the body of the response, not in statusText.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,293 @@
...
+    })
+    const groupsResponse = await cognitoIdentityProviderClient.send(
+      getGroupsCommand
+    )
+
+    if (
+      !groupsResponse.Groups ||
+      !groupsResponse.Groups.find(
+        ({ GroupName }) => GroupName === 'gcn.nasa.gov/circular-submitter'
+      )
```
lpsinger, line 97:
```suggestion
      !groupsResponse.Groups?.find(
        ({ GroupName }) => GroupName === 'gcn.nasa.gov/circular-submitter'
      )
```

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    const response = await cognitoIdentityProviderClient.send(
+      getEndorserDataCommand
+    )
+
+    if (!response.Users)
+      throw new Response(null, {
+        statusText: 'Requested user does not exist',
+        status: 400,
+      })
```
lpsinger, line 117: But you already retrieved the user above.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,293 @@
...
+        },
+        FilterExpression: '#status = :status',
+      })
+    ).Items as EndorsementRequest[]
+
+    if (existingEndorsement.length > 0)
+      throw new Response(null, {
+        statusText: 'An endorsement already exists for the selected endorser',
+        status: 400,
+      })
```
lpsinger, line 152: Use a ConditionExpression so that DynamoDB checks this for you.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,293 @@
...
+      },
+      ExpressionAttributeValues: {
+        ':pendingStatus': 'pending',
+      },
+    }
+
+    db._doc.delete(param, (err, data) => {
+      if (err) throw new Response(null, { statusText: err.message })
+      return data
+    })
```
lpsinger, line 231: Really? It doesn't support async/await? You have to use the old callback style?

lpsinger (changes requested):

app/routes/user/endorsements.server.ts
dakota002: duplicated code from moving stuff around. It will be removed

app/routes/user/endorsements.server.ts
dakota002: I guess? 
image

There appears to be a promise method on the delete function though. I'll try that out and see whats up. My implementation was just from copying the usage I came across in the docs.

lpsinger: How long has it been since you rebased this branch? arc.codes recently migrated to AWS SDK v3 for the Node.js 18.x runtime. It looks like you might be using the AWS SDK v2.

app/routes/user/endorsements.server.ts
dakota002: When debugging I wasn't getting any info out of the errors unless there was something in the status text

I moved it for this one, but since the others are all returning otherwise the same type of error I am leaving the string in the statusText

lpsinger: What do you mean by "not getting any info"? Where were you expecting to see this message?

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+      ExpressionAttributeValues: {
+        ':status': 'pending',
+        ':endorserEmail': endorserEmail,
+        ':requestorEmail': this.#currentUserEmail,
+        ':created': Date.now(),
+        ':endorserSub': escapedEndorserString,
+        ':requestorSub': this.#sub,
+      },
+      ConditionExpression:
+        'NOT (#endorserSub = :endorserSub and #requestorSub = :requestorSub)',
```
lpsinger, line 134: What? No. This should be checking that if the endorsement request exists, then its status is "pending".

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+        '#created': 'created',
+        '#requestorSub': 'requestorSub',
+        '#endorserSub': 'endorserSub',
+      },
+      ExpressionAttributeValues: {
+        ':status': 'pending',
+        ':endorserEmail': endorserEmail,
+        ':requestorEmail': this.#currentUserEmail,
+        ':created': Date.now(),
+        ':endorserSub': escapedEndorserString,
```
lpsinger, line 130: No, here it should be the endorser sub itself, not the escaped version.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+    const db = await tables()
+    await db.circular_endorsements.update({
+      Key: {
+        requestorSub: requestorSub,
+        endorserSub: this.#sub,
+      },
+      UpdateExpression: 'set #status = :status',
+      ExpressionAttributeNames: {
+        '#status': 'status',
+      },
```
lpsinger, line 171: Probably not a DynamoDB reserved word. No need for ExpressionAttributeNames.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+    const param = {
+      Key: {
+        requestorSub: this.#sub,
+        endorserSub: endorserSub,
+      },
+      TableName: tableName,
+      ConditionExpression: '#status = :pendingStatus',
+      ExpressionAttributeNames: {
+        '#status': 'status',
+      },
```
lpsinger, line 197: Again, no need for ExpressionAttributeNames.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+          : '#endorserSub = :endorserSub',
+      FilterExpression: role == 'requestor' ? undefined : '#status = :status',
+      ExpressionAttributeNames: {
+        '#requestorSub': 'requestorSub',
+        '#requestorEmail': 'requestorEmail',
+        '#endorserSub': 'endorserSub',
+        '#endorserEmail': 'endorserEmail',
+        '#status': 'status',
+        '#created': 'created',
+      },
```
lpsinger, line 232: No need for ExpressionAttributeNames.

app/routes/user/endorsements.server.ts
```diff
@@ -0,0 +1,265 @@
...
+      UpdateExpression:
+        'set #status = :status, #endorserEmail = :endorserEmail, #requestorEmail = :requestorEmail, #created = :created',
+      ExpressionAttributeNames: {
+        '#status': 'status',
+        '#endorserEmail': 'endorserEmail',
+        '#requestorEmail': 'requestorEmail',
+        '#created': 'created',
+        '#requestorSub': 'requestorSub',
+        '#endorserSub': 'endorserSub',
+      },
```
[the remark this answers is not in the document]
dakota002: Status is a key word, but i can update the others

app/routes/user/endorsements.server.ts
dakota002: Why are we checking pending?  This is correct, since it is true if there isnt an existing one with a matching endorser and requestor sub.

lpsinger: So that creating an endorsement is idempotent.

lpsinger (approved):

PR #365, opened 2022-10-25
Peer Endorsement Breakout PR 2, Auth token refresh

dakota002:


dakota002:
Auth token refresh and adding users to new group

lpsinger (changes requested):
The token refreshing facility should be very general and independent from the circulars machinery. Please split out the token refreshing into a self-contained PR.

app/routes/__auth/login.ts
```diff
@@ -13,23 +13,23 @@ import { generators } from 'openid-client'
...
   const sub = claims.sub
   const email = claims.email as string
-
+  const refreshToken = tokenSet.refresh_token as string
   const idp =
     claims.identities instanceof Array && claims.identities.length > 0
       ? (claims.identities[0].providerName as string)
       : null
-
+  const cognitoUserName = claims['cognito:username'] as string
```
lpsinger, line 27: Why are we grabbing the username?

app/routes/__auth/login.ts
```diff
@@ -13,23 +13,23 @@ import { generators } from 'openid-client'
...
   const groups = ((claims['cognito:groups'] ?? []) as string[]).filter(
-    (group) => group.startsWith('gcn.nasa.gov')
+    (group) => group.split('/')[0] == 'gcn.nasa.gov'
```
lpsinger, line 29: Equivalent, but simpler:

```suggestion
    (group) => group.startsWith('gcn.nasa.gov/')
```

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +15,18 @@ export async function getUser({ headers }: Request) {
   const email = session.get('email') as string
   const groups = session.get('groups') as string[]
   const idp = session.get('idp') as string | null
-
+  const refreshToken = session.get('refreshToken') as string
+  const cognitoUserName = session.get('cognitoUserName') as string
   if (!sub) return null
+  const user = { sub, email, groups, idp, refreshToken, cognitoUserName }
+  await getLatestUserGroups(user)
```
lpsinger, line 22: Won't this _always_ submit the refresh token on each page load?

We should only submit the refresh token if (1) the access token is absent or (2) the access token has expired.

lpsinger (changes requested):

app/routes/__auth/login.ts
dakota002: Does this change comply with githubs security checks? I had previously used a startswith, but the automated checks were failing. Does the slash at the end get around this?

app/routes/__auth/login.ts
dakota002: Ya, this makes more sense when the other function is in mind, but this command needs it:

const command = new AdminAddUserToGroupCommand({
      Username: user.cognitoUserName,
      UserPoolId: process.env.COGNITO_USER_POOL_ID,
      GroupName: 'gcn.nasa.gov/circular-submitter',
    })

app/routes/__auth/user.server.ts
dakota002: Just to clarify on 1, if the access token is absent or if the permission is absent?

The access token should be there either way, right?

lpsinger: Doesn't matter, it's a false alarm. We aren't using this as a URL. I had previously dismissed the alert.

app/routes/__auth/login.ts
lpsinger: How is the Cognito username related to the sub?

app/routes/__auth/login.ts
lpsinger: Doesn't matter, it's a false alarm. We aren't using this as a URL. I had previously dismissed the alert.

app/routes/__auth/login.ts
dakota002: I assume it is tied to the user?  its just how the command expects arguments.

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminAddUserToGroup.html

app/routes/__auth/login.ts
dakota002: Good enough for me then!

app/routes/__auth/login.ts
lpsinger: Right, but how will having that in the session help us? A user won't be adding _themselves_ to a group.

lpsinger: Never mind, I understand now.

dakota002:
@lpsinger I think I got it going in the direction we talked about on Friday. The latest version adds the access token to the session and I added a method to clear it, which will be used when an endorsement is approved

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
 }
+
+export async function clearUserToken(sub: string) {
+  const db = await tables()
+  const targetSessionResults = await db.sessions.query({
+    IndexName: 'bySub',
+    KeyConditionExpression: '#sub = :sub',
+    ExpressionAttributeNames: {
+      '#sub': 'sub',
+      '#token': 'token',
```
lpsinger, line 45:
```suggestion
```

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
+      '#idx': '_idx',
+      '#ttl': '_ttl',
```
lpsinger, line 47:
```suggestion
```

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
+      ':sub': sub,
+    },
+    ProjectionExpression: '#token, #sub, #idx, #ttl',
+  })
+
+  if (!targetSessionResults.Items) return
+
+  const targetSession = targetSessionResults.Items[0]
+  console.log(targetSession['_idx'])
+  console.log(targetSession['_ttl'])
```
lpsinger, line 59:
```suggestion
```

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
+    ExpressionAttributeNames: {
+      '#sub': 'sub',
+      '#token': 'token',
+      '#idx': '_idx',
+      '#ttl': '_ttl',
+    },
+    ExpressionAttributeValues: {
+      ':sub': sub,
+    },
+    ProjectionExpression: '#token, #sub, #idx, #ttl',
```
lpsinger, line 52:
```suggestion
    ProjectionExpression: '#idx',
```

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
+  })
+
+  if (!targetSessionResults.Items) return
+
+  const targetSession = targetSessionResults.Items[0]
```
lpsinger, line 57: Need to loop over all matching sessions.

app/routes/__auth/user.server.ts
```diff
@@ -30,3 +34,37 @@ export async function updateSession(user: any) {
...
+  console.log(targetSession['_idx'])
+  console.log(targetSession['_ttl'])
+  await db.sessions.update({
+    Key: { _idx: targetSession['_idx'] as string },
+    UpdateExpression: 'set #token = :nullToken',
```
lpsinger, line 62: How about you just remove the token, rather than setting it to the empty string?

lpsinger (changes requested):

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +16,55 @@ export async function getUser({ headers }: Request) {
   const email = session.get('email') as string
   const groups = session.get('groups') as string[]
   const idp = session.get('idp') as string | null
-
+  const refreshToken = session.get('refreshToken') as string
+  const cognitoUserName = session.get('cognitoUserName') as string
+  const token = session.get('token') as string
   if (!sub) return null
+  const user = { sub, email, groups, idp, refreshToken, cognitoUserName, token }
+  if (!token) {
```
lpsinger, line 24: or if token has expired...

lpsinger (changes requested):

app/lib/utils.ts
```diff
@@ -31,3 +36,14 @@ export function getFormDataString(formData: FormData, key: string) {
...
+export async function refreshUser(
+  user: NonNullable<Awaited<ReturnType<typeof getUser>>>
+) {
+  const client = await getOpenIDClient()
+  const refreshedTokenSet = await client.refresh(user.refreshToken)
+  const user_new = userFromTokenSet(refreshedTokenSet)
+  await updateSession(user_new)
+  user.groups = user_new.groups
+  user.token = user_new.token
+}
```
lpsinger, line 49: I thought that JS is generally pass-by-value, not pass-by-reference. Does this actually update the user?

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +16,59 @@ export async function getUser({ headers }: Request) {
...
   const idp = session.get('idp') as string | null
-
+  const refreshToken = session.get('refreshToken') as string
+  const cognitoUserName = session.get('cognitoUserName') as string
+  const token = session.get('token') as string
   if (!sub) return null
+  const user = { sub, email, groups, idp, refreshToken, cognitoUserName, token }
+  if (
+    !token ||
+    JSON.parse(Buffer.from(token.split('.')[1], 'base64').toString()).exp *
```
lpsinger, line 26: I think that we already have a dependency on jose. Use it to parse the JWT.

app/routes/__auth/user.server.ts
dakota002: Removed in latest commit

app/routes/__auth/user.server.ts
dakota002: removed in latest commit

app/routes/__auth/user.server.ts
dakota002: removed in latest commit

app/routes/__auth/user.server.ts
dakota002: updated in latest commit

app/routes/__auth/user.server.ts
dakota002: Latest version takes all of items and uses Promise.all to resolve however many there may be

app/routes/__auth/user.server.ts
dakota002: Added a check for expiration in latest commit

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +16,59 @@ export async function getUser({ headers }: Request) {
...
+
+export async function updateSession(user: any) {
+  const session = await storage.getSession()
+  Object.entries(user).forEach(([key, value]) => {
+    session.set(key, value)
+  })
+  await storage.commitSession(session)
+}
+
+export async function clearUserToken(sub: string) {
```
lpsinger, line 43: Add a comment to explain what this function does. Something like:

```
// Clear the access tokens from all sessions belonging to the user with the given sub, to force them to get new access tokens.
```

Several of the other functions would benefit from some comments.

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +16,59 @@ export async function getUser({ headers }: Request) {
...
+    },
+    ProjectionExpression: '#idx',
+  })
+
+  if (!targetSessionResults.Items) return
+
+  const promises = targetSessionResults.Items.map((item) =>
+    db.sessions.update({
+      Key: { _idx: item['_idx'] as string },
+      UpdateExpression: 'set #token = :nullToken',
```
lpsinger, line 63: Just delete the access token. (Don't just set it to null.)

app/routes/__auth/login.ts
```diff
@@ -105,6 +106,7 @@ export const loader: LoaderFunction = async ({ request: { headers, url } }) => {
 
     const user = userFromTokenSet(tokenSet)
     const session = await sessionPromise
+    session.set('token', tokenSet.access_token)
```
lpsinger, line 109: Might as well call this field `accessToken` or `access_token`, to distinguish it from the many other kinds of tokens (access, ID, refresh).

```suggestion
    session.set('accessToken', tokenSet.access_token)
```

app/routes/__auth/login.ts
```diff
@@ -13,23 +13,24 @@ import { generators } from 'openid-client'
...
 
-function userFromTokenSet(
+export function userFromTokenSet(
   tokenSet: TokenSet
 ): NonNullable<Awaited<ReturnType<typeof getUser>>> {
   const claims = tokenSet.claims()
   const sub = claims.sub
   const email = claims.email as string
-
+  const token = tokenSet.access_token as string
```
lpsinger, line 22:
```suggestion
  const accessToken = tokenSet.access_token as string
```

lpsinger (changes requested):

app/routes/__auth/user.server.ts
dakota002: Is there a specific method that handles this with jose? I've never heard of that package

lpsinger: jose parses JWTs. It is a dependency of node-openid-client.

app/lib/utils.ts
dakota002: Yea, it works. I threw a console log to check, but if i go into cognito and change the groups and just call the refresh user method I get the new groups 
image

app/routes/__auth/user.server.ts
dakota002: I didnt realize there was a command to just REMOVE an attribute, trying that now

app/routes/__auth/user.server.ts
dakota002: Is this what you're talking about using? https://github.com/panva/jose/blob/HEAD/docs/functions/jwt_verify.jwtVerify.md#readme
If so, where can I find the secret, issuer, and audience values?

dakota002: Or are we able to just do this through one of the issue or client classes from node-openid-client?

dakota002: Possibly disregard, I updated it by parsing the JSON and using the TokenSet class's expired method

app/routes/__auth/user.server.ts
dakota002: I added that and one other, are there specific ones you think need comments?

app.arc
```diff
@@ -30,6 +30,10 @@ email_notification_subscription
   topic *String
   name byTopic
   
+sessions
+  sub *String
+  name bySub
```
lpsinger, line 35:
```suggestion
  name sessionsBySub
```

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +17,64 @@ export async function getUser({ headers }: Request) {
...
+    session.set(key, value)
+  })
+  await storage.commitSession(session)
+}
+
+// Clear the access tokens from all sessions belonging to the user with the given sub, to force them to get new access tokens.
+export async function clearUserToken(sub: string) {
+  const db = await tables()
+  const targetSessionResults = await db.sessions.query({
+    IndexName: 'bySub',
```
lpsinger, line 55:
```suggestion
    IndexName: 'sessionsBySub',
```

app/lib/utils.ts
```diff
@@ -31,3 +36,15 @@ export function getFormDataString(formData: FormData, key: string) {
...
+export async function refreshUser(
+  user: NonNullable<Awaited<ReturnType<typeof getUser>>>
+) {
+  const client = await getOpenIDClient()
+  const refreshedTokenSet = await client.refresh(user.refreshToken)
+  const user_new = userFromTokenSet(refreshedTokenSet)
+  await updateSession(user_new)
+  user.groups = user_new.groups
+  user.accessToken = user_new.accessToken
+}
```
lpsinger, line 50: Move this to somewhere in `~/routes/__auth`? Perhaps `~/routes/__auth/user.server.ts`?

app/routes/__auth/login.ts
```diff
@@ -105,6 +106,7 @@ export const loader: LoaderFunction = async ({ request: { headers, url } }) => {
 
     const user = userFromTokenSet(tokenSet)
     const session = await sessionPromise
+    session.set('accessToken', tokenSet.access_token)
     Object.entries(user).forEach(([key, value]) => {
       session.set(key, value)
     })
```
lpsinger, line 112: This looks like a copy of `updateSession`.

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +17,64 @@ export async function getUser({ headers }: Request) {
   const email = session.get('email') as string
   const groups = session.get('groups') as string[]
   const idp = session.get('idp') as string | null
-
+  const refreshToken = session.get('refreshToken') as string
+  const cognitoUserName = session.get('cognitoUserName') as string
+  const accessToken = session.get('accessToken') as string
```
lpsinger, line 22: `as string` isn't right here; it might be undefined.

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +17,64 @@ export async function getUser({ headers }: Request) {
...
+
+  if (!targetSessionResults.Items) return
+
+  const promises = targetSessionResults.Items.map((item) =>
+    db.sessions.update({
+      Key: { _idx: item['_idx'] as string },
+      UpdateExpression: 'REMOVE #token',
+      ExpressionAttributeNames: {
+        '#token': 'token',
+      },
```
lpsinger, line 75: I think we renamed this field:

```suggestion
      UpdateExpression: 'REMOVE accessToken',
```

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +17,64 @@ export async function getUser({ headers }: Request) {
...
+  const refreshToken = session.get('refreshToken') as string
+  const cognitoUserName = session.get('cognitoUserName') as string
+  const accessToken = session.get('accessToken') as string
   if (!sub) return null
+  const user = {
+    sub,
+    email,
+    groups,
+    idp,
+    refreshToken,
```
lpsinger, line 29: We may want to not expose the refreshToken in the returned user object, because the application itself will never have a need for it and it is very sensitive.

app/routes/__auth/user.server.ts
```diff
@@ -14,8 +17,64 @@ export async function getUser({ headers }: Request) {
...
+    cognitoUserName,
+    accessToken,
+  }
+  if (!accessToken) await refreshUser(user)
+
+  const tokenSet = new TokenSet(
+    JSON.parse(Buffer.from(accessToken.split('.')[1], 'base64').toString())
+  )
+  if (tokenSet.expired()) await refreshUser(user)
```
lpsinger, line 38: Please parse this using jose.

lpsinger (changes requested):

app/lib/utils.ts
dakota002: Sounds good, will be moved in next commit

app/routes/__auth/login.ts
dakota002: Yea, originally the two looked a bit different, but I refactored a bit and I think it is a bit cleaner now. Also added a simple comment on the method to give it some clarity in use.

app/routes/__auth/user.server.ts
dakota002: We do use it to call the refresh though, and the user object is what we use to set the session. Any suggestions for how to not include it?

app/routes/__auth/user.server.ts
dakota002: Do you have docs or a reference for how to do this? I came across this, https://github.com/panva/jose/blob/cfc63778b133b8c6731080a033c7701dfd9cc9ca/docs/functions/jwe_general_decrypt.generalDecrypt.md#readme is this what we want? 

If so, where do i get the key parameter?

dakota002: I don't know why I didn't find this sooner, https://github.com/panva/jose/blob/cfc63778b133b8c6731080a033c7701dfd9cc9ca/docs/functions/util_decode_jwt.decodeJwt.md#readme

app/routes/__auth/user.server.ts
dakota002: Actually I was able to do some rearranging and got it out of the getUser method, let me know what you think

lpsinger (approved):

PR #367, opened 2022-10-27
Peer Endorsement Breakout PR 3, Circulars backend

dakota002:
A class for reading data which does not require a user's sub, and another which allows for writing, but requires the sub.

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -1,0 +1,212 @@
+/*!
+ * Copyright © 2022 United States Government as represented by the Administrator
+ * of the National Aeronautics and Space Administration. No copyright is claimed
+ * in the United States under Title 17, U.S. Code. All Other Rights Reserved.
+ *
+ * SPDX-License-Identifier: NASA-1.3
+ */
+
+import { tables } from '@architect/functions'
+import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
+import { dynamoDBAutoIncrement } from '~/lib/dynamoDBAutoIncrement'
+import { getUser } from '../__auth/user.server'
+
+export type CircularModel = {
+  circularId?: number
+  sub?: string
+  createdOn: number
+  subject: string
+  body: string
+  submitter: string
+}
+
+export class CircularsServer {
+  #sub?: string
+  #email?: string
+  #groups?: string[]
+  #name?: string
+  #affiliation?: string
+  #isLocal: boolean
+
+  private constructor(
+    isLocal: boolean,
+    sub?: string,
+    email?: string,
+    groups?: string[],
+    name?: string,
+    affiliation?: string
+  ) {
+    this.#affiliation = affiliation
+    this.#email = email
+    this.#groups = groups
+    this.#name = name
+    this.#sub = sub
+    this.#isLocal = isLocal
+  }
+
+  static async create(request: Request) {
+    const user = await getUser(request)
+    const isLocal = new URL(request.url).hostname == 'localhost'
+    return new this(
+      isLocal,
+      user?.sub,
+      user?.email,
+      user?.groups,
+      user?.name,
+      user?.affiliation
+    )
+  }
+
+  /**
+   * Gets all available circulars
+   * @param searchString - a string that will be checked against subjects,
+   * bodies, and authors of saved circulars. A blank value returns all.
+   * @returns an array of CircularModel objects
+   */
+  async getCirculars(searchString?: string) {
+    let result = await this.#scanCircularsTable(searchString)
+    let items: CircularModel[] = result.Items
+    while (result.LastEvaluatedKey) {
+      result = await this.#scanCircularsTable(
+        searchString,
+        result.LastEvaluatedKey
+      )
+      items = [...items, ...result.Items]
+    }
+    return items.sort((a, b) => (b.circularId ?? 0) - (a.circularId ?? 0))
+  }
+
+  async #scanCircularsTable(searchString?: string, lastEvaluatedKey?: any) {
+    const db = await tables()
+    const result = await db.gcn_circulars.scan({
+      ExclusiveStartKey: lastEvaluatedKey,
+      ExpressionAttributeValues: searchString
+        ? {
+            ':searchTerm': searchString,
+          }
+        : undefined,
+      FilterExpression: searchString
+        ? 'contains(subject, :searchTerm) OR contains(body, :searchTerm) OR contains(submitter, :searchTerm)'
+        : undefined,
+    })
+    return result
+  }
+
+  /**
+   * Get a specified Circular by Id.
+   *
+   * Throws an HTTP error if:
+   *  - the provided Id does not match any existing circulars
+   *
+   * @param id - the number Id (ex. 12345) of a specific Circular
+   * @returns a CircularModel object
+   */
+  async get(circularId: number): Promise<CircularModel> {
+    const db = await tables()
+    const result = await db.gcn_circulars.get({
+      circularId: circularId,
+    })
+    if (!result)
+      throw new Response('The requested circular does not exist', {
+        status: 404,
+      })
+    return result
+  }
+
+  /**
+   * Adds a new entry into the GCN Circulars table
+   *
+   * Throws an HTTP error if:
+   *  - The current user is not signed in, verified by the class's #sub and #groups properties
+   *  - The current user is not in the submitters group
+   *  - Body or Subject are blank
+   * @param body - main content of the Circular
+   * @param subject - the title/subject line of the Circular
+   */
+  async createNewCircular(body: string, subject: string) {
+    if (!this.#sub || !this.#groups)
+      throw new Response('User is not signed in', { status: 403 })
+    if (!this.#groups.includes('gcn.nasa.gov/circular-submitter'))
+      throw new Response('User is not in the submitters group', {
+        status: 403,
+      })
+    if (!body || !subject)
+      throw new Response('Subject and Body cannot be blank', { status: 400 })
+
+    const db = await tables()
+    const doc = db._doc as unknown as DynamoDBDocument
+
+    const tableName = db.name('gcn_circulars')
+    const indexTableName = db.name('auto_increment_metadata')
+
+    const submitter =
+      this.#name +
+      (this.#affiliation ? ` at ${this.#affiliation} ` : ' ') +
+      this.#email
+
+    await dynamoDBAutoIncrement({
+      doc,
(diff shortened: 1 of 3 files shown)


dakota002:
This PR should include some validation too, but I might want to reuse some of that which is being used in #364

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+   * @param subject - the title/subject line of the Circular
+   * @param correctedCircularId - an optional parameter, the Id of an existing circular that this new one updates
+   */
+  async createNewCircular(
+    body: string,
+    subject: string,
+    correctedCircularId?: number
+  ) {
+    if (
+      !this.#groups.find((group) => group === 'gcn.nasa.gov/circular-submitter')
```
lpsinger, line 139:
```suggestion
      !this.#groups.includes('gcn.nasa.gov/circular-submitter')
```

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Deletes the circular corresponding to the given circularId
+   *
+   * Throws an HTTP error if:
+   *  - The current user is not in the moderator group
+   * @param circularId - the ID of the circular to be deleted
+   */
+  async deleteCircular(circularId: number) {
+    if (
+      !this.#groups.find((group) => group === 'gcn.nasa.gov/circular-moderator')
```
lpsinger, line 239:
```suggestion
      !this.#groups.includes('gcn.nasa.gov/circular-moderator')
```

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+
+export type CircularModel = {
+  circularId: number
+  sub: string
+  createdOn: number
+  subject: string
+  body: string
+  submitter?: string
+  linkedEvent?: string
+  correctedCircularId?: number
```
lpsinger, line 20: Remove these for now; they are for future enhancement.

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * SPDX-License-Identifier: NASA-1.3
+ */
+
+import { tables } from '@architect/functions'
+import { getUser } from '../__auth/user.server'
+
+export type CircularModel = {
+  circularId: number
+  sub: string
```
lpsinger, line 14:
```suggestion
  sub?: string
```

because legacy users won't have Cognito accounts.

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+  createdOn: number
+  subject: string
+  body: string
+  submitter?: string
```
lpsinger, line 18:
```suggestion
  submitter: string
```

This should never be null.

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+  linkedEvent?: string
+  correctedCircularId?: number
+}
+
+export class PublicCircularsServer {
```
lpsinger, line 23: Why two classes?

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Get a specified Circular by Id.
+   *
+   * Throws an HTTP error if:
+   *  - the provided Id does not match any existing circulars
+   *
+   * @param id - the number Id (ex. 12345) of a specific Circular
+   * @returns a CircularModel object
+   */
+  async getCircularById(circularId: number): Promise<CircularModel> {
```
lpsinger, line 72: How about just:

```suggestion
  async get(circularId: number): Promise<CircularModel> {
```

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+    )
+      throw new Response('User is not in the submitters group', {
+        status: 403,
+      })
+    if (!body || !subject)
+      throw new Response('Subject and Body cannot be blank', { status: 400 })
+
+    const db = await tables()
+    const result = await db.gcn_circulars.scan({
+      FilterExpression: 'attribute_exists(circularId)',
```
lpsinger, line 149: How couldn't this attribute exist? It's the partition key.

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,248 @@
...
+    })
+
+    let itemId =
+      (result.Items[0]
+        ? Math.max(...result.Items.map((x) => x['circularId']))
+        : 0) + 1
+    const submitter =
+      this.#name +
+      (this.#affiliation ? ` at ${this.#affiliation} ` : ' ') +
+      this.#email
```
lpsinger, line 160: This is a very expensive (as in both cycles and $'s) to get the next circular ID. Please follow the optimistic locking example in https://bitesizedserverless.com/bite/reliable-auto-increments-in-dynamodb/, which has the following essential features:

1. A special row in a table that stores the last assigned ID.
2. A query that retrieves that last assigned ID.
3. A transaction that attempts to increment the last assigned ID and put the new record.
4. A loop that retries steps 2 and 3 until the transaction succeeds.

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
dakota002: I like it, changing now

app/routes/user/index.tsx
```diff
@@ -23,7 +23,6 @@ export const handle = { breadcrumb: 'Profile', getSitemapEntries: () => null }
 export async function loader({ request }: DataFunctionArgs) {
   const user = await getUser(request)
   if (!user) throw new Response(null, { status: 403 })
```
dakota002, line 25:
```suggestion
  if (!user) throw new Response(null, { status: 403 })
  
```

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,212 @@
...
+    let result = await this.#scanCircularsTable(searchString)
+    let items: CircularModel[] = result.Items
+    while (result.LastEvaluatedKey) {
+      result = await this.#scanCircularsTable(
+        searchString,
+        result.LastEvaluatedKey
+      )
+      items = [...items, ...result.Items]
+    }
+    return items.sort((a, b) => (b.circularId ?? 0) - (a.circularId ?? 0))
```
lpsinger, line 76: I don't think that circularID can be undefined, can it?

```suggestion
    return items.sort((a, b) => b.circularId - a.circularId)
```

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,212 @@
...
+    this.#email = email
+    this.#groups = groups
+    this.#name = name
+    this.#sub = sub
+    this.#isLocal = isLocal
+  }
+
+  static async create(request: Request) {
+    const user = await getUser(request)
+    const isLocal = new URL(request.url).hostname == 'localhost'
```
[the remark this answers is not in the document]
dakota002: I see we use the process.env.ARC_ENV in sandboxOidcIdp.js file, would that work here as well? Would be be able to infer it from a testing environment, or does that not work for this case?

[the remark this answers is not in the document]
lpsinger: I use it from time to time. Best not to break it if we can help it. I don't see anything in ARC_ENV that would tell us if we are connected to the real DynamoDB; it has ports.tables set regardless of the value of ARC_ENV.

One option would be to look at the DynamoDB client's endpoint URL.

Have you tried asking about this on the Architect discord?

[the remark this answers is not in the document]
dakota002: Not yet, i've been reading through the architect docs to try to get a better idea of what is/isn't available already. I didn't see anything that explicitly states it either, but it did get me thinking that maybe it would make sense to just add a variable into our .env? 

I'll ask there shortly though, just to be safe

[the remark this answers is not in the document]
dakota002: From ryan on discord:

> Assuming this is a Lambda, you can use the ARC_SANDBOX env var to know whether you are in sandbox

I'll explore that option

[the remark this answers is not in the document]
dakota002: locally, that object looks like this

```
{
  apiType: 'http',
  cwd: 'C:\\Repos\\gcn.nasa.gov',
  ports: { http: 3333, tables: 5555, _arc: 2222 },
  staticPath: 'C:\\Repos\\gcn.nasa.gov\\public',
  version: '5.4.1',
  lambdaSrc: 'C:\\Repos\\gcn.nasa.gov\\server'
}
```

@lpsinger do you know if any of these fields are useful for us to use? I don't know what the values are on the hosted env, but I imagine we could use cwd?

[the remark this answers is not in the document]
lpsinger: Try this...

```ts
(await (db._db as unknown as DynamoDB).config.endpoint?.())?.hostname
```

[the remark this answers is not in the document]
dakota002: Nice, I updated it to use that

app/routes/circulars/circulars.server.ts
```diff
@@ -0,0 +1,214 @@
...
+    const db = await tables()
+    const doc = db._doc as unknown as DynamoDBDocument
+
+    const tableName = db.name('circulars')
+    const indexTableName = db.name('auto_increment_metadata')
+
+    const submitter = formatAuthor({
+      name: this.#name,
+      affiliation: this.#affiliation,
+      email: this.#email ?? '',
```
lpsinger, line 142: How can the email be undefined?

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
dakota002: In the case where the getter methods are only being used. If the user is not signed in, but just reading/searching for circulars

lpsinger: I don't understand. The purpose of this function is to put a new circular into the database. How would email be undefined?

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
dakota002: I just updated the check, been a little since I looked at the actual function. This should be resolved in the latest commit

lpsinger (approved):

PR #378, opened 2022-11-03
Hero update

dakota002:


lpsinger (changes requested):
I think that you forgot to update the image...
Screen Shot 2022-11-11 at 12 57 03

lpsinger:
What is the purpose of SectionWrapper?

dakota002:
> I think that you forgot to update the image... Screen Shot 2022-11-11 at 12 57 03

No, I definitely had it working before as I showed in the screenshots on the linked issue. Are you seeing console errors or anything that would suggest that there is something missing on your end?

As for the SectionWrapper, each page was originally a single section, this just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere

lpsinger:
> This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere

Are we actually making use of granular margins anywhere right now?

lpsinger (changes requested):
At large screen sizes, the "Transient" text in the image gets cut off.

https://user-images.githubusercontent.com/728407/201409216-4219ea64-6c18-4284-9a72-75b8d428dfdc.mov

dakota002:
> > This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere
> 
> Are we actually making use of granular margins anywhere right now?

Yes, the background image is specific to this section, otherwise it would take up the whole page.

What size are you working with that is causing that break? like resolution. I can't make that happen in chrome on my laptop

lpsinger:
> > > This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere
> > 
> > 
> > Are we actually making use of granular margins anywhere right now?
> 
> Yes, the background image is specific to this section, otherwise it would take up the whole page.
> 
> What size are you working with that is causing that break? like resolution. I can't make that happen in chrome on my laptop

window.innerWidth is 1907 pixels.

Can you reproduce it on your machine if you zoom out?

dakota002:
> > > > This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere
> > > 
> > > 
> > > Are we actually making use of granular margins anywhere right now?
> > 
> > 
> > Yes, the background image is specific to this section, otherwise it would take up the whole page.
> > What size are you working with that is causing that break? like resolution. I can't make that happen in chrome on my laptop
> 
> window.innerWidth is 1907 pixels.
> 
> Can you reproduce it on your machine if you zoom out?

Ah yep that did it, looks like it should be resolved by changing the background position style to only specify the x. Changing now

lpsinger:
> > This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere
> 
> Are we actually making use of granular margins anywhere right now?

Ping...

dakota002:
> > > This change just allows it to be broken up so that the individual sections can be distinct, and so we can have more granular control over space, like if we want to make more use of side margins elsewhere
> > 
> > 
> > Are we actually making use of granular margins anywhere right now?
> 
> Ping...

lpsinger:
> Ah yep that did it, looks like it should be resolved by changing the background position style to only specify the x. Changing now

That looks better. However the people are getting cut off:

Screen Shot 2022-11-11 at 14 18 41

I think that the problem here is that as the screen grows larger, the image expands in all directions. At larger screen sizes we shouldn't enlarge the image. What we need, I think, are tiling images for the left and right margin. We can ask Britt to prepare the tiling images. Meanwhile, can you figure out the code to make it work?

dakota002:
I think it makes more sense to just expand the image itself out to the max width. I dont think tiling is needed, just full width for screens with resolution up to 4k

lpsinger:
Can we mock up the other option without coding it? (I know I could do a mockup in Photoshop; do you have any similar software that would let you do it?)

PR #399, opened 2022-11-14
Peer Endorsement Breakout PR 4, Shared components

dakota002:
Adds some shared components. Lifted from big PR,


dakota002:
New permission tab

Dropdown populates with users who have the submitters permission

image


New request(s) appear below the dropdown:
image


The approver (a user who is in the submitters group) sees requests like this, similar format to the credential cards
image

lpsinger (changes requested):
Some initial feedback:

- There's no need for permissions.tsx and permissions/index.tsx because there is only one route underneath permissions/. Please combine into permissions.tsx.
- EndorsementReqeustCard and EndorsementRequestForm are unlikely to be reused. Please put them in permissions.tsx.

app/routes/__auth/user.server.ts
```diff
@@ -156,3 +161,47 @@ export async function refreshUser(refreshToken: string, session: Session) {
...
   return parsedTokenSet.user
 }
+
+/**
+ * Gets another user from cognito
+ *
+ * @param sub - the sub of another user
+ * @returns a user if found, otherwise undefined
+ */
+export async function getOtherUser(sub: string) {
```
lpsinger, line 171: This is a confusing name for this function, because it doesn't return something that has the same type as `getUser`. A better name would be `getCognitoUsernameForSub`. Also this function doesn't have much use outside of the endorsements server; can it go in endorsements.server.ts?

app/routes/__auth/user.server.ts
```diff
@@ -156,3 +161,47 @@ export async function refreshUser(refreshToken: string, session: Session) {
...
+/**
+ * Adds a user to a group
+ *
+ * Throws an HTTP error if:
+ *  - The provided sub does not correspond to an existing user
+ *
+ * @param sub - sub of another user
+ * @param group - group to which the user corresponding to the given sub should be added
+ */
+export async function addUserToGroup(sub: string, group: string) {
```
lpsinger, line 193: For now, can we keep this in endorsements.server.ts?

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+}
+
+export async function loader({ request }: DataFunctionArgs) {
+  const endorsementServer = await EndorsementsServer.create(request)
+  const user = await getUser(request)
+
+  // Maps the retrived users to an object that can be used in the ComboBox
+  const validUsers = await endorsementServer
+    .getSubmitterUsers()
+    .then((response) => {
```
lpsinger, line 38: Use `await` rather than old-style promise chaining style.

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+      const mappedUsers = response?.map((user) =>
+        user.Attributes?.filter((attr) =>
+          ['sub', 'email', 'name'].includes(attr.Name as string)
+        ).reduce(
+          (obj, item) =>
+            Object.assign(obj, { [item.Name as string]: item.Value }),
+          {}
+        )
```
lpsinger, line 46: I prefer `Object.fromEntries(blahBlahBlah.map((...) => [key, value])` which we use in other places throughout the code.

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+  Label,
+} from '@trussworks/react-uswds'
+import { useState } from 'react'
+import SegmentedCards from '~/components/SegmentedCards'
+import { getFormDataString } from '~/lib/utils'
+import { getUser } from '../__auth/user.server'
+import type { EndorsementRequest } from './endorsements.server'
+import { EndorsementsServer } from './endorsements.server'
+
+type intermediateMappedUser = {
```
lpsinger, line 25: What does "intermediate" mean here?

```suggestion
type IntermediateMappedUser = {
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+    <Form method="post">
+      <h2 id="modal-request-heading">Request Endorsement</h2>
+      <div className="usa-prose">
+        Requesting an endorsement from another user will share your email with
+        that individual. Please keep this in mind when submitting. Enter the
+        email of the individual you want to be endorsed by. They will receive a
+        notification alerting them to this request.
+      </div>
+      <Label htmlFor="endorserSub">Endorser</Label>
+      <input type="hidden" name="endorserSub" value={endorserSub} />
```
lpsinger, line 282: Is the hidden field necessary? Does the ComboBox not populate anything in the form data?

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+  return (
+    <Form method="post">
+      <h2 id="modal-request-heading">Request Endorsement</h2>
+      <div className="usa-prose">
+        Requesting an endorsement from another user will share your email with
+        that individual. Please keep this in mind when submitting. Enter the
+        email of the individual you want to be endorsed by. They will receive a
+        notification alerting them to this request.
+      </div>
+      <Label htmlFor="endorserSub">Endorser</Label>
```
lpsinger, line 281:
```suggestion
      <Label htmlFor="endorserSubSelect">Endorser</Label>
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+      <input type="hidden" name="endorserSub" value={endorserSub} />
+      <ComboBox
+        id="endorserSubSelect"
+        name="endorserSubSelect"
+        options={validUsers}
+        onChange={(e) => setEndorserSub(e ?? '')}
+        defaultValue={''}
+      />
+      <Button
+        className="usa-button margin-top-2"
```
lpsinger, line 291: The usa-button style is baked into this component, I think.

```suggestion
        className="margin-top-2"
```

Why is it necessary to control the margin?

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,300 @@
...
+        ).reduce(
+          (obj, item) =>
+            Object.assign(obj, { [item.Name as string]: item.Value }),
+          {}
+        )
+      ) as intermediateMappedUser[]
+
+      return mappedUsers.map((user) => ({
+        value: user.sub,
+        label: `${user.name} - ${user.email}`,
```
lpsinger, line 51: My preference would be to return the email, name, and sub from the loader, and let the React component do the formatting.

lpsinger (changes requested):
- [x] Are all buttons that, if you clicked them, would produce an error (like deleting a reported endorsement) disabled or invisible?
- [ ] I am uncomfortable with sending the entire list of users to the browser. Can we do a server-side autocomplete combo box?

app/routes/user/endorsements.server.ts
```diff
@@ -190,8 +194,7 @@ export class EndorsementsServer {
         ':pendingStatus': 'pending',
       },
     }
-
-    await db._doc.delete(param).promise()
+    db._doc.delete(param)
```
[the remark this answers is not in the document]
dakota002: I thought so too, this one was strange. Originally the await db._doc.delete(param).promise() was working, then I started getting errors that _doc.delete(...).promise() was not a function, so I tried as await db._doc.delete(param), which gave an error in the IDE that the await was unneeded.

app/routes/user/permissions.tsx
dakota002: Yea, another one where thats how I thought it should work too, but I found that as I tried the form it wasn't updating the value. The ComboBox has a required onChange, but not a value property

app/routes/user/permissions.tsx
dakota002: I'll remove that class, but I included the margin to give a bit of spacing. The way the buttons layout otherwise had it touching the bottom of the ComboBox and it looked a little too squished

app/routes/user/permissions.tsx
dakota002: I can make that work, I'll have that in the next update

app/routes/__auth/user.server.ts
dakota002: Sounds good, moving it and the addUserToGroup function in the next commit

app/routes/user/endorsements.server.ts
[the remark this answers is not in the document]
lpsinger: That's because the type signature from @types/architect__functions is wrong. See DefinitelyTyped/DefinitelyTyped#63652

app/routes/user/permissions.tsx
lpsinger: Please show screen shots.

dakota002: Without the margin specified:
image

With:
image

app/routes/user/endorsements.server.ts
```diff
@@ -8,11 +8,12 @@
 
 import { tables } from '@architect/functions'
 import {
+  AdminAddUserToGroupCommand,
```
lpsinger, line 11: Remember to add this to the arc.codes plugin that sets the Lambda's IAM actions.

app/routes/user/endorsements.server.ts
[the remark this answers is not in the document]
lpsinger: Please rebase, to pick up the type fix from main.

app/routes/user/endorsements.server.ts
```diff
@@ -67,14 +68,7 @@ export class EndorsementsServer {
       )
 
     const escapedEndorserString = endorserSub.replaceAll('"', '\\"')
```
lpsinger, line 70: Move the escaping not the `#getCognitoUsernameForSub` method.

app/routes/user/endorsements.server.ts
```diff
@@ -67,14 +68,7 @@ export class EndorsementsServer {
...
-        new ListUsersCommand({
-          UserPoolId: process.env.COGNITO_USER_POOL_ID,
-          Filter: `sub = "${escapedEndorserString}"`,
-        })
-      )
-    )?.Users?.[0]
+    const user = await this.#getCognitoUsernameForSub(escapedEndorserString)
     if (!user)
       throw new Response('Requested user does not exist', {
         status: 400,
```
lpsinger, line 74: Move the exception into the `#getCognitoUsernameForSub` method, so that on success it always returns a valid user and never undefined?

app/routes/user/endorsements.server.ts
```diff
@@ -248,4 +248,49 @@ export class EndorsementsServer {
...
     return result.Users
   }
+
+  /**
+   * Gets another user from cognito
+   *
+   * @param escapedEndorserString - the sub of another user
+   * @returns a user if found, otherwise undefined
+   */
+  async #getCognitoUsernameForSub(escapedEndorserString: string) {
```
lpsinger, line 258:
```suggestion
  async #getCognitoUsernameForSub(sub: string) {
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+      if (!requestorSub)
+        throw new Response('The user was not found', { status: 400 })
+      await endorsementServer.updateEndorsementRequestStatus(
+        intent,
+        requestorSub
+      )
+      break
+    case 'delete':
+      if (!endorserSub)
+        throw new Response('The user was not found', { status: 400 })
```
lpsinger, line 101: Doesn't `endorsementServer.deleteEndorsementRequest` do this check?

```suggestion
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+    case 'create':
+      if (!endorserSub)
+        throw new Response('The user was not found', { status: 400 })
+      await endorsementServer.createEndorsementRequest(endorserSub)
+      break
+    case 'approved':
+    case 'rejected':
+    case 'reported':
+      if (!requestorSub)
+        throw new Response('The user was not found', { status: 400 })
```
lpsinger, line 93: Doesn't `endorsementServer.updateEndorsementRequestStatus` do this check?

```suggestion
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+export async function action({ request }: DataFunctionArgs) {
+  const data = await request.formData()
+  const intent = getFormDataString(data, 'intent')
+  const endorserSub = getFormDataString(data, 'endorserSub')
+  const requestorSub = getFormDataString(data, 'requestorSub')
+  const endorsementServer = await EndorsementsServer.create(request)
+  switch (intent) {
+    case 'create':
+      if (!endorserSub)
+        throw new Response('The user was not found', { status: 400 })
```
lpsinger, line 86: Doesn't the `endorsementServer.createEndorsementRequest` method do this check?

```suggestion
```

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+        requestorSub
+      )
+      break
+    case 'delete':
+      if (!endorserSub)
+        throw new Response('The user was not found', { status: 400 })
+      await endorsementServer.deleteEndorsementRequest(endorserSub)
+      break
+    default:
+      break
```
lpsinger, line 105: Should not be reached. Raise an exception.

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+export async function loader({ request }: DataFunctionArgs) {
+  const endorsementServer = await EndorsementsServer.create(request)
+  const user = await getUser(request)
+  const parsedUrl = new URL(request.url)
+  const filterString = parsedUrl.searchParams.get('filter')
+
+  // Maps the retrived users to an object that can be used in the ComboBox
+  const validUsers = filterString
+    ? await endorsementServer.getSubmitterUsers().then((response) => {
```
lpsinger, line 39: Use `async`/`await`, not the old `then` callback style.

lpsinger (changes requested):

app/routes/user/permissions.tsx
lpsinger: I think that's because you have it all wrapped in a `ButtonGroup`. Try removing the `ButtonGroup`.

dakota002: Nope, no button groups in that form, they are used in the cards

src/plugins/lambdaCognitoPermissions.js
```diff
@@ -24,7 +24,8 @@ module.exports.deploy = {
           {
             Effect: 'Allow',
             Action: [
-              'cognito-idp:AdminListGroupsForUser',
+              'cognito-idp:AdminAddUserToGroupCommand',
+              'cognito-idp:AdminListGroupsForUserCommand',
```
lpsinger, line 28:
```suggestion
              'cognito-idp:AdminAddUserToGroup',
              'cognito-idp:AdminListGroupsForUser',
```

lpsinger (changes requested):

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+export function EndorsementRequestForm() {
+  const { validUsers } = useLoaderData<typeof loader>()
+  const [options, setOptions] = useState([])
+  const [endorserSub, setEndorserSub] = useState('')
+  const fetcher = useFetcher()
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target
+
+    if (value && validUsers.findIndex((f) => f.value === value) < 0) {
```
lpsinger, line 290:
```suggestion
    if (value && !validUsers.includes(value)) {
```

app/routes/user/endorsements.server.ts
dakota002: Where should the escaping get moved to? It is used a few lines lower in addition to the get method

lpsinger:
Where did we leave this PR?

app/routes/user/permissions.tsx
```diff
@@ -0,0 +1,338 @@
...
+      <ComboBox
+        id="endorserSubSelect"
+        name="endorserSubSelect"
+        options={options}
+        onChange={(e) => setEndorserSub(e ?? '')}
+        defaultValue={''}
+        inputProps={{ onChange: handleInputChange }}
+      />
+      <Button
+        className="margin-top-2"
```
lpsinger, line 328: If you put this in a `<ButtonGroup>`, then you won't need the manual styling.

app/components/Header.tsx
```diff
@@ -82,6 +82,13 @@ export function Header({ email }: { email?: string }) {
                       >
                         Profile
                       </Link>,
+                      <Link
+                        key="permissions"
+                        to="/user/permissions"
```
lpsinger, line 87: We need a more descriptive name for this page, because "Credentials" and "Permissions" sound too similar.

app/routes/user/endorsements.server.ts
lpsinger: I'm OK with repeating it here and also inside `#getCognitoUsernameForSub`. What I want to avoid is future bugs related to confusion about which functions expect sub itself versus which functions expect escaped sub.

app/components/Header.tsx
lpsinger: Sounds good.

app/routes/user/endorsements.tsx
```diff
@@ -0,0 +1,339 @@
...
+      break
+    case 'delete':
+      if (!endorserSub)
+        throw new Response('Valid endorser is required', { status: 403 })
+      await endorsementServer.deleteEndorsementRequest(endorserSub)
+      break
+    default:
+      throw new Response('An error has occured, unreachable code', {
+        status: 404,
+      })
```
lpsinger, line 106:
```suggestion
      throw new Response('unknown intent', {
        status: 400,
      })
```

lpsinger (changes requested):

lpsinger (approved):

PR #420, opened 2022-12-02
Sets the stored sub to really be subiss

dakota002:
This looks like it will be all that is needed code wise after a migration

### DIFF
--- a/app/routes/__auth/login.ts
+++ b/app/routes/__auth/login.ts
@@ -12,6 +12,7 @@ import type { TokenSet } from 'openid-client'
 import { generators } from 'openid-client'
 import { getOpenIDClient, oidcStorage } from './auth.server'
 import type { getUser } from './user.server'
+import { subiss } from './user.server'
 import { updateSession } from './user.server'
 
 export function userFromTokenSet(tokenSet: TokenSet): {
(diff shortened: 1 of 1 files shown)


dakota002:
@lpsinger I'm checking out some of the places it is used, should we update every instance of it (Client credential table, email notifications, w/e else) or just the name in the method being edited?

lpsinger:
Let's update it everywhere. A database migration will be unavoidable because we will need to replace the "sub" data with "subiss" data.

app/routes/__auth/login.ts
```diff
@@ -12,14 +12,15 @@ import type { TokenSet } from 'openid-client'
...
+import { subiss } from './user.server'
 import { updateSession } from './user.server'
 
 export function userFromTokenSet(tokenSet: TokenSet): {
   user: NonNullable<Awaited<ReturnType<typeof getUser>>>
   refreshToken: string
 } {
   const claims = tokenSet.claims()
-  const sub = claims.sub
+  const subIss = subiss({ sub: claims.sub, iss: claims.iss })
```
lpsinger, line 23: Would it be more convenient if we renamed this function `getSubiss` so that we don't need to do case gymnastics to avoid collisions?

lpsinger (approved):

lpsinger:
We need to write 2 little scripts to do the migration:

1. [ ] Populate `subiss` field of all tables by looking up `sub` from DynamoDB and `iss` from Cognito. In the process of running this script, we will find out if we have any collisions (unlikely).
2. [ ] Delete the old `sub` field.

lpsinger (changes requested):
The code looks ready to go.

I have experimentally shown that changing the key schema of an existing DynamoDB table defined by a CloudFormation template will result in the table being deleted and a new one created.

Please design a step-by-step migration strategy and add as comments on this issue.

PR #466, opened 2022-12-12
Display descriptions of available scopes

lpsinger:


lpsinger:
@dakota002, can you tell why TypeScript is complaining about this and how to make it happy? Why can't TypeScript tell that the filter() method ensures that the resulting GroupNames are all defined?

lpsinger:
Screen Shot 2022-12-12 at 15 00 42

PR #488, opened 2022-12-20
Add JSON format alerts to notice type control

dakota002:


app/components/NoticeFormat.tsx
```diff
@@ -61,11 +63,26 @@ export function NoticeFormat({
...
   ]
 
   return (
     <>
-      <input type="hidden" name={name} value={currentValue} />
+      <input
+        type="hidden"
+        name={name}
+        value={currentValue}
+        onChange={(e) => (stateFunction ? stateFunction(e) : null)}
```
lpsinger, line 84: Wouldn't this result in multiple invocation of the state function? Since the value of the hidden field only every changes programmatically, when we set it.

```suggestion
```

app/components/NoticeFormat.tsx
```diff
@@ -82,7 +99,10 @@ export function NoticeFormat({
               aria-checked={currentValue === value}
               aria-describedby={`${value}-label`}
               outline={currentValue !== value}
-              onClick={() => setCurrentValue(value)}
+              onClick={() => {
+                setCurrentValue(value)
+                if (stateFunction) stateFunction(value)
```
lpsinger, line 104: Use optional chaining:

```suggestion
                stateFunction?.(value)
```

app/routes/quickstart/alerts.tsx
```diff
@@ -21,6 +21,9 @@ export default function () {
   const [params] = useSearchParams()
   const [alertsValid, setAlertsValid] = useState(false)
   const clientId = params.get('clientId') || undefined
+  const [format, setFormat] = useState<'text' | 'voevent' | 'binary' | 'json'>(
```
lpsinger, line 24: We already have this type defined elsewhere... let's use it.

app/components/NestedCheckboxes.tsx
```diff
@@ -57,6 +57,9 @@ function NestedCheckboxNode({
       }
     }
   })
+
+  const isExternalLink = link && link.startsWith('http')
```
lpsinger, line 61: Technically, there are many href values that start with http that would _not_ be external links: `http`, `http/foo/bar`, etc. Take a look at how [react-router](https://github.com/remix-run/react-router)'s Link component or [rehype-external-links](https://github.com/rehypejs/rehype-external-links) do it.

lpsinger (changes requested):

app/components/NoticeFormat.tsx
```diff
@@ -8,14 +8,16 @@
 import { Button, ButtonGroup } from '@trussworks/react-uswds'
 import { useState } from 'react'
 
-type ValueType = 'text' | 'voevent' | 'binary'
+type ValueType = 'text' | 'voevent' | 'binary' | 'json'
 
 export function NoticeFormat({
   name,
   value,
+  stateFunction,
```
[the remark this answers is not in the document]
dakota002: Yea thats much better, changing now

app/components/NoticeFormat.tsx
dakota002: Good catch, who knows what past-Dakota was thinking. 

Cleaning this up in the next commit

app/components/NestedCheckboxes.tsx
dakota002: I updated this in the latest commit to compare the provided path with our origin. I'm reading over the react-router link now though, that might have a better implementation

app/components/NoticeFormat.tsx
```diff
@@ -61,11 +63,26 @@ export function NoticeFormat({
         </>
       ),
     },
+    {
+      value: 'json' as ValueType,
+      label: 'JSON',
+      description: (
+        <>
+          Preliminary new notice types available only via JSON (other notice
+          types under development).
```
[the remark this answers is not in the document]
dakota002: Maybe something that links to the schema repo/schema browser and the LVK definition? something like:

"JSON formatted notices defined by the [gcn-schema](nasa-gcn/gcn-schema) and [LVK](https://emfollow.docs.ligo.org/userguide/content.html#notice-formats)"

app/components/NestedCheckboxes.tsx
```diff
@@ -57,6 +59,10 @@ function NestedCheckboxNode({
       }
     }
   })
+  const origin = useOrigin()
+  const isExternalLink =
+    link && !link.startsWith('/') && new URL(link).origin !== origin
```
lpsinger, line 64: Suggest factoring isExternalLink into a pure function.

app/components/NestedCheckboxes.tsx
```diff
@@ -70,20 +76,25 @@ function NestedCheckboxNode({
...
-                    e.stopPropagation()
-                  }}
-                >
-                  Details
-                </Link>
-                )
-              </>
-            )}
+            {link &&
+              (isExternalLink ? (
```
lpsinger, line 80: Change the condition in the ternary operator to isExternalLink(link) to avoid checking if link is defined twice.

app/components/NoticeFormat.tsx
```diff
@@ -61,6 +66,20 @@ export function NoticeFormatInput({
         </>
       ),
     },
+    ...(location.pathname !== '/user/email/edit'
```
lpsinger, line 69: Add a FIXME comment to explain why we are checking the location here.

Better yet… just add a showJson prop to the component to explicitly control whether it is visible.

lpsinger (changes requested):

lpsinger:
@dakota002, did you address all of these issues? Is it time for me to review it again?

dakota002:
Almost, I think i need to finish some final changes on the link checking. I'm on another branch at the moment, but I can pause on that and switch back to this

app/components/NoticeFormat.tsx
```diff
@@ -61,6 +66,20 @@ export function NoticeFormatInput({
...
     },
+    ...(showJson
+      ? [
+          {
+            value: 'json' as NoticeFormat,
+            label: 'JSON',
+            description: (
+              <>
+                Preliminary new notice types available only via JSON (other
+                notice types under development).
```
lpsinger, line 77:
```suggestion
                New notice types in JSON format defined using <a href="https://json-schema.org" rel="external">JSON schema</a>
```

app/components/NestedCheckboxes.tsx
```diff
@@ -25,6 +25,15 @@ function allSame([first, ...rest]: any[]) {
   return rest.every((value) => value === first)
 }
 
+function isExternalLink(link: string): boolean {
+  try {
+    const url = new URL(link)
+    return url.hostname !== window.location.hostname
```
lpsinger, line 31: So, this will only work on the client side. A simpler test is, if the URL starts with http:// or https://, then we should treat it as absolute.

lpsinger (changes requested):

app/components/NestedCheckboxes.tsx
dakota002: That makes sense,  what are your thoughts on this implementation:
```
function isExternalLink(link: string): boolean {
  try {
    return ['https:', 'http:'].includes(new URL(link).protocol)
  } catch (error) {
    return false
  }
}
```

lpsinger: `new URL()` doesn't work on URLs without schemes, so this is more complicated than it needs to be. I suggest:

```ts
function isExternalLink(link: string): boolean {
  return link.startsWith('https://') || link.startsWith('http://')
}
```

lpsinger (approved):

PR #489, opened 2022-12-21
Add unit tests for dynamoDBAutoIncrement, and fix some bugs

lpsinger:
Note that this comes with some extra dev dependencies for mocking up DynamoDB that aren't necessary for the web site as a whole, so this is a prime candidate for factoring into a separate package once we are confident that we are going to use it.


lpsinger:
@dakota002, the tests are passing now. This is ready for review.

dakota002 (approved):

PR #490, opened 2022-12-22
Circular UI

dakota002:


dakota002:
Archive/Search page:
image

Expanded example:
image

Info(previously just the circulars page):
image

Submit:
image

lpsinger:
Please use the USWDS [search box component](https://designsystem.digital.gov/components/search/).

lpsinger (changes requested):
- [x] Please use the USWDS [search box component](https://designsystem.digital.gov/components/search/).
- [ ] I don't think that the accordion is appropriate here. Let's just have each line be a link.

dakota002:
V2:
* URL format updated for simple nav to specific circulars: ex: `http://localhost:3333/circulars/1`
* Links following the {id} - {subject} pattern
image

* UI for read page:
image

lpsinger:
This is no longer a mockup, is it?

app/routes/circulars/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+    <GridContainer>
+      <div className="usa-prose">
+        <h1>GCN Circulars</h1>
+      </div>
+      <p>
+        GCN Circulars are rapid astronomical bulletins submitted by and
+        distributed to community members worldwide. For more information, see{' '}
+        <Link to="">docs</Link>
+      </p>
+      <div id="sticky-header" className="sticky margin-bottom-1 padding-top-1">
```
lpsinger, line 89: The ID is unused:

```suggestion
      <div className="sticky margin-bottom-1 padding-top-1">
```

app/routes/circulars/submit.tsx
```diff
@@ -0,0 +1,124 @@
...
+  if (!user || !user.groups.includes('gcn.nasa.gov/circular-submitter'))
+    throw new Response('', { status: 403 })
+  return { user }
+}
+
+export async function action({ request }: DataFunctionArgs) {
+  const data = await request.formData()
+  const body = getFormDataString(data, 'body')
+  const subject = getFormDataString(data, 'subject')
+  if (!body || !subject) throw new Response('', { status: 403 })
```
lpsinger, line 27:
```suggestion
  if (!body || !subject) throw new Response('Body and subject are required', { status: 400 })
```

app/routes/circulars/submit.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </div>
+        <Label htmlFor="subject">
+          Subject
+          <abbr title="required" className="usa-label--required">
+            *
+          </abbr>
+        </Label>
+        <small className="text-base maxw-full">
+          The subject line should start with the name of the transient which
+          must be one of the <Link to="">known keywords</Link>. (Contact us for
```
lpsinger, line 66: Fill in link.

"Contact us" should be a link to the contact form.

theme/custom.scss
```diff
@@ -490,3 +490,10 @@ a {
   white-space: nowrap;
   overflow: hidden;
 }
+
+.sticky {
```
lpsinger, line 494: Is this class needed? There are a variety of USWDS utility classes that you could use here, like `.position-sticky`.

lpsinger (changes requested):

dakota002:
Mobile s :
image

app/routes/circulars/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <h1>GCN Circulars</h1>
+      </div>
+      <p>
+        GCN Circulars are rapid astronomical bulletins submitted by and
+        distributed to community members worldwide. For more information, see{' '}
+        <Link to="">docs</Link>
+      </p>
+      <div className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1">
+        <div className="usa-search">
+          <Search defaultValue={searchTerm ?? ''} onSubmit={() => {}} />
```
lpsinger, line 84: Wrap this in a Form so that the search box is functional.

package.json
```diff
@@ -39,6 +39,7 @@
     "memoizee": "^0.4.15",
     "openid-client": "^5.3.1",
     "react": "^18.2.0",
+    "react-autocomplete-hint": "^2.0.0",
```
lpsinger, line 42: Unused dependency.

lpsinger (changes requested):

app/routes/circulars/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+      </div>
+      <p>
+        GCN Circulars are rapid astronomical bulletins submitted by and
+        distributed to community members worldwide. For more information, see{' '}
+        <Link to="">docs</Link>
+      </p>
+      <div className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1">
+        <div className="usa-search">
+          <Form>
+            <Search defaultValue={searchTerm ?? ''} onSubmit={() => {}} />
```
lpsinger, line 85:
```suggestion
            <Search defaultValue={searchTerm ?? ''} name="search" onSubmit={() => {}} />
```

so that clicking "search" actually does a search.

lpsinger (changes requested):
Please add a feature flag environment variable to enable/disable the entire circulars UI so that we can merge this into main without showing the partial functionality in production.

The circulars UI should 404 if the environment variable is unset.

app/routes/circulars/index.tsx
dakota002: I think this is under the hood of the Search component already, this does currently behave as expected

dakota002: Actually it worked even without being wrapped in a Form, which is fairly strange

lpsinger: Hmm, OK.

dakota002:
Should I remove these nav options for now, or add a loader to the index to get the flag?
image

app/routes/circulars/index.tsx
```diff
@@ -19,6 +19,7 @@ import { getUser } from '../__auth/user.server'
 import { CircularsServer } from './circulars.server'
 
 export async function loader({ request }: DataFunctionArgs) {
+  if (!process.env['FEATURE']) throw new Response('', { status: 404 })
```
lpsinger, line 22: Please call this feature flag something more descriptive, like `GCN_CIRCULARS_ENABLE`.

Note that we already have a route /circulars on the main branch, which should not 404 error.

lpsinger (changes requested):

lpsinger:
> Should I remove these nav options for now, or add a loader to the index to get the flag?
> image

The latter. Show them if the feature flag is set.

app/routes/circulars/index.tsx
dakota002: True, I have updated the file names to keep the /circulars route behaving as it should

lpsinger (approved):

PR #505, opened 2023-01-03
Factor dynamodb-autoincrement into a separate package

lpsinger:


dakota002:
I think this is all good, but build might need to be updated to remove the `npm run test` command

dakota002 (approved):

PR #507, opened 2023-01-05
Draft: Add Markdown Circular editor

lpsinger:


lpsinger:
It turns out that bytemd is written by Bytedance, creator of TikTok. Although it's a very nice library, there is no way that we would be permitted to use it due to supply chain risk management concerns.

PR #508, opened 2023-01-06
Api backend

dakota002:


dakota002:
This update will include adding some validation to the subject line input, which applies to the UI and API

Invalid:
image

Valid:
image

app/routes/circulars/circulars.server.ts
```diff
@@ -63,8 +63,10 @@ export async function list({
 
   // Calculate pagination assuming that last === the number of records.
   const totalPages = Math.ceil(last / limit)
-  const firstIdOnPage = Math.max(last - page * limit + 1, 1)
-
+  const firstIdOnPage = Math.max(
```
dakota002, line 66: This change was mainly for an issue locally where the `limit` far exceeds the `last`, and was causing an issue where no circulars would show up

app/routes/circulars/index.mdx
```diff
@@ -41,36 +42,11 @@ Please follow the [style guidelines](/docs/styleguide). In particular, the email
...
-  <li className="grid-col-2">SDSS</li>
-  <li className="grid-col-2">SGR</li>
-  <li className="grid-col-2">Swift</li>
-  <li className="grid-col-2">SWIFT</li>
-  <li className="grid-col-2">Virgo</li>
-  <li className="grid-col-2">VLA</li>
-  <li className="grid-col-2">VLBI</li>
-  <li className="grid-col-2">XRB</li>
-  <li className="grid-col-2">XTR</li>
+  {validSubjectKeywords?.map((keyword) => (
```
dakota002, line 45: any new keywords can just be added to the `validSubjectKeywords` list and this page will automatically update

sandbox-seed.json
```diff
@@ -1,109 +1,130 @@
 {
   "circulars": [
     {
+      "dummy": 0,
```
lpsinger, line 4: I fixed this on the main branch. Please rebase.

app/routes/circulars/circulars.server.ts
lpsinger: OK, would you please split that out as a separate PR?

app/routes/circulars/index.mdx
lpsinger: Please split the keyword validation out to a separate PR (separate from adding the API endpoints).

app/routes/circulars/circulars.server.ts
dakota002: I think you actually solved this in a previous commit, my branch was just out of date

sandbox-seed.json
dakota002: Sweet, rebased

app/routes/circulars/index.mdx
dakota002: Okay, I originally have that in the api/circulars file, should I keep the validation in that file, or move them into another file like utils?

dakota002: Broke it out here https://github.com/nasa-gcn/gcn.nasa.gov/pull/523

lpsinger (changes requested):
I don't think that we are ready to add the API yet. We first need to figure out how users are going to authenticate for API access.

PR #547, opened 2023-01-23
Convert architect plugins to mjs

lpsinger:
Down with require.


lpsinger:
Blocked by https://github.com/architect/architect/issues/1401.

PR #564, opened 2023-01-30
Pulls out the extract methods for reuse in later commit

dakota002:


lpsinger (changes requested):
Since these are not general utilities but only related to Cognito, please put them in a new file. I suggest the path app/lib/cognito.ts.

Also, since we are moving the definition of these functions out of the context where they are called, please add a JSDoc description of each of the two functions.

lpsinger (approved):

Issue #291, opened 2022-09-09
Onboarding of GCN Circulars submitters from GCN Classic


lpsinger:
There are many existing submitters of GCN Circulars, which essentially take the form of a list of email addresses, names, and institutions. How do we onboard them into the new site? When someone signs up on gcn.nasa.gov, how can we permit them to "claim" the submission privileges of their old GCN Classic registration?

lpsinger:
Option 1.

dakota002:
@lpsinger  Any thoughts on the cognito attributes vs dynamodb table options for adding user info like name and association? 
I have linked a PR above that would allow users to add that info with the dynamo option

jracusin:
I would like to see a method that is sort of both.  I think we should ingest the list of name, email, institution, and submit/receive preferences from the gcn3.list file, which @Vidushi-GitHub can write a script to parse appropriately.  We do email all of those people asking them to make an account, and that they cannot submit unless they make an account after a specified date.  However, I would like us to add all of the people who are currently signed up to receive to automatically be transferred to do so, and if/when they make an account with that email address, they can claim it and configure as desired.  Anyone currently with submit privileges should continue to have them and not require peer endorsement.

jracusin:
Add email address, name, affiliation, submit True/False, receive True/False to table.  When legacy user logs in for the first time, these fields are associated with them, and old entry will be deleted from table.

dakota002:
As a legacy user (someone who is able to submit to GCN Classic), I need to be able to submit GCN circulars to the new system (Remix app) without having to go through a verification process. 

Acceptance Criteria:

* [ ] Anyone on the parsed list of legacy users who already has an account is added to the Submitters group

* [ ] A table exists with the basic info (email [required], name [?], affiliation [?]) of each remaining legacy user

* [ ] When a legacy user submits a valid email to circulars@{domain}, the function that validates the user also checks if they are on the legacy users list, if they do not currently belong to the submitters group

* [ ] When a legacy user signs up for an account, a check is run on the legacy users table. If the user exists:

	* They are added to the submitters group
	* Their info gets migrated into the corresponding fields in their cognito account
	* Circulars they have submitted update their sub field to match the new one
	* Their info is removed from the legacy table


If I missed anything please let me know

lpsinger:
The web site components are done now. @dakota002 is working on the CloudFormation changes now, tracked in a separate PR.

Issue #612, opened 2023-02-15
Distribute GCN Circulars by email


lpsinger:
We also need to work out how people opt in to receiving them.

- [x] Cognito users
- [x] Legacy users

lpsinger:
NB: we must remember that legacy users may be signed up to just receive or just send Circulars and take that into account in the email-incoming lambda. See https://github.com/nasa-gcn/gcn.nasa.gov/pull/618#discussion_r1107763657.

dakota002:
@lpsinger for legacy users who only receive, is there any information other than their email that we need? I think it might just be enough for that group to just put the receivers' emails into a dynamodb table and just get the whole list

lpsinger:
No, I don't think so.

dakota002:
For the Cognito Users: #689

dakota002:
Legacy Users: #704

PR #618, opened 2023-02-15
Onboarding Step 1

dakota002:
Adds legacy_users table, checks if user is a legacy user if they do not have valid permissions, updates sandbox seed and invoker tests

Should address the 2nd and 3rd points of https://github.com/nasa-gcn/gcn.nasa.gov/issues/291
Point 1 will likely be a one off script, once the other steps are set up.
Point 4 might need to be wrapped into the post sign up confirm lambda

### DIFF
--- a/app/email-incoming/index.ts
+++ b/app/email-incoming/index.ts
@@ -29,6 +29,13 @@ import { sendEmail } from '~/lib/email'
 import { getOrigin } from '~/lib/env'
 import { tables } from '@architect/functions'
 
+interface UserData {
+  email: string
+  sub?: string
+  name?: string
+  affiliation?: string
+}
+
 const cognito = new CognitoIdentityProviderClient({})
 const origin = getOrigin()
 
@@ -67,22 +74,11 @@ async function handleRecord(record: SNSEventRecord) {
     return
   }
 
-  // Check if submitter is in the submitters group
-  const data = await cognito.send(
-    new ListUsersInGroupCommand({
-      GroupName: group,
-      UserPoolId: process.env.COGNITO_USER_POOL_ID,
-    })
-  )
+  const userData =
+    (await getCognitoUserData(userEmail)) ??
+    (await getLegacyUserData(userEmail))
 
-  const userTypeData = data.Users?.find(
-    (user) => extractAttributeRequired(user, 'email') == userEmail
-  )
-  const legacyUserData = !userTypeData
-    ? await getLegacyUser(userEmail)
-    : undefined
-
-  if (!userTypeData && !legacyUserData) {
+  if (!userData) {
     await sendEmail(
       'GCN Circulars',
       userEmail,
@@ -92,20 +88,6 @@ async function handleRecord(record: SNSEventRecord) {
     return
   }
 
-  const userData = userTypeData
-    ? {
-        sub: extractAttributeRequired(userTypeData, 'sub'),
-        email: extractAttributeRequired(userTypeData, 'email'),
-        name: extractAttribute(userTypeData, 'name'),
-        affiliation: extractAttribute(userTypeData, 'custom:affiliation'),
-      }
-    : {
-        sub: legacyUserData.email,
-        email: legacyUserData.email,
-        name: legacyUserData.name,
-        affiliation: legacyUserData.affiliation,
-      }
-
   const circular = {
     dummy: 0,
     createdOn: Date.now(),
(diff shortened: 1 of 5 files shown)


app/email-incoming/index.ts
```diff
@@ -87,12 +92,19 @@ async function handleRecord(record: SNSEventRecord) {
...
-  }
+  const userData = userTypeData
+    ? {
+        sub: extractAttributeRequired(userTypeData, 'sub'),
+        email: extractAttributeRequired(userTypeData, 'email'),
+        name: extractAttribute(userTypeData, 'name'),
+        affiliation: extractAttribute(userTypeData, 'custom:affiliation'),
+      }
+    : {
+        sub: legacyUserData.email,
```
lpsinger, line 103: This is not an OIDC subject.

```suggestion
```

lpsinger (changes requested):
I suggest restructuring this as two functions, `getUser` and `getLegacyUser`, both of which return a promise of a type like this:

```ts
interface UserData {
  email: string
  sub?: string
  name?: string
  affiliation?: string
}
```

And then you can get the user by using the nullish coalescing operator:

```ts
const user = (await getUser()) ?? (await getLegacyUser())
```

app/email-incoming/index.ts
jracusin: Make sure you track that some legacy users are submit & receive, or only submit, or only receive.

app/email-incoming/index.ts
```diff
@@ -133,10 +118,47 @@ export async function handler(event: SNSEvent) {
...
+    (user) => extractAttributeRequired(user, 'email') == userEmail
+  )
+  return userTypeData
+    ? {
+        sub: extractAttributeRequired(userTypeData, 'sub'),
+        email: extractAttributeRequired(userTypeData, 'email'),
+        name: extractAttribute(userTypeData, 'name'),
+        affiliation: extractAttribute(userTypeData, 'custom:affiliation'),
+      }
+    : undefined
```
lpsinger, line 144: Two slightly shorter options:

```suggestion
  if (userTypeData)
    return {
        sub: extractAttributeRequired(userTypeData, 'sub'),
        email: extractAttributeRequired(userTypeData, 'email'),
        name: extractAttribute(userTypeData, 'name'),
        affiliation: extractAttribute(userTypeData, 'custom:affiliation'),
      }
```

or

```suggestion
  return userTypeData && {
      sub: extractAttributeRequired(userTypeData, 'sub'),
      email: extractAttributeRequired(userTypeData, 'email'),
      name: extractAttribute(userTypeData, 'name'),
      affiliation: extractAttribute(userTypeData, 'custom:affiliation'),
    }
```

app/email-incoming/index.ts
dakota002: Sweet, okay I updated it to use two functions, but opted for names getCognitoUserData (so as not to confuse with the existing getUser function) and getLegacyUserData. 

As for the receive/submit difference, that should all be okay, for this stuff we are specifically focusing on Submitters, so the only users that should get uploaded into this specific table should be Submitters

app/email-incoming/index.ts
```diff
@@ -133,10 +118,47 @@ export async function handler(event: SNSEvent) {
...
   const db = await tables()
-  return await db.legacy_users.get({ email: userEmail })
+  const data = await db.legacy_users.get({ email: userEmail })
+  return data
+    ? {
+        email: data.email,
+        name: data.name,
+        affiliation: data.affiliation,
+      }
+    : undefined
```
lpsinger, line 163: (same comment as above)

app/lib/email.ts
```diff
@@ -45,7 +46,11 @@ export async function sendEmail(
   try {
     await client.send(command)
   } catch (e) {
-    if (
+    if (e instanceof MessageRejected && hostname == 'localhost') {
```
lpsinger, line 49: This must still throw in production.

lpsinger (changes requested):

app/lib/email.ts
dakota002: That should be okay though since the the hostname should be different right?

dakota002: Hmm true, think I should just let it throw normally and remove this condition?

lpsinger: That's fine with me.

app/email-incoming/index.ts
lpsinger: Whether it's the same table or a different one, we just have to remember to take care of this. I've made a note of it at https://github.com/nasa-gcn/gcn.nasa.gov/issues/612#issuecomment-1433651046.

PR #628, opened 2023-02-22
Legacy migration

dakota002:


dakota002:
For point 4 in https://github.com/nasa-gcn/gcn.nasa.gov/issues/291

My thought with this is that we could trigger it directly from the PostConfirmSignupFunction, by adding a lambda-client and using the [invoke command](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-lambda/classes/invokecommand.html).

dakota002:
Latest Update: stripped down this PR to only include some minor changes, since the function will exist in the template.

The `@sandbox` thing in the app.arc wasn't actually doing anything

esbuild.mjs
```diff
@@ -5,7 +5,7 @@ const watch = args.includes('--watch')
 
 await build({
   bundle: true,
-  entryPoints: ['./app/email-incoming/index.js'],
+  entryPoints: ['./app/email-incoming/index.ts'],
```
lpsinger, line 8: What does this change do?

lpsinger (changes requested):

esbuild.mjs
dakota002: the actual file is a ts file, not js

dakota002: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/email-incoming/index.ts

lpsinger: OK. It would be better next time for this kind of change to be in a separate PR. (Keep each PR focused on one set of related changes.)

app.arc
```diff
@@ -59,8 +59,9 @@ circular_endorsements
   endorserSub *String
   name circularEndorsementsByEndorserSub
 
-@sandbox
-invoker events
```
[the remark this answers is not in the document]
dakota002: Nope, intentional. It doesn't do anything in the app.arc. events pragma are included by default with the invoker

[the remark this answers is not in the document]
lpsinger: OK, that's fine. Again, that's not related to the subject of this PR ("Legacy migration"). Better in the future to peel this kind of thing into a separate PR.

lpsinger (approved):

PR #629, opened 2023-02-23
Adds multiple email sends for creation and updates of endorsement requests

dakota002:


dakota002:
@lpsinger I think the wording on the emails could be improved, do you have any suggestions?

app/routes/user/endorsements.tsx
```diff
@@ -236,6 +238,11 @@ export function EndorsementRequestCard({
               name="requestorSub"
               value={endorsementRequest.requestorSub}
             />
+            <input
+              type="hidden"
+              name="requestorEmail"
+              value={endorsementRequest.requestorEmail}
+            />
```
lpsinger, line 245: Can we get the user's email address from the session rather than injecting it into the form?

lpsinger (changes requested):

lpsinger:
> @lpsinger I think the wording on the emails could be improved, do you have any suggestions?

It's a good start. The emails should be a bit more self-contained - imagine a GCN user doesn't know what an endorsement is and they get an endorsement request. The email should explain why they are getting the message and what actions they can take.

app/routes/user/endorsements.tsx
dakota002: The requestorEmail is not the same as the current user's in this case. When this form is submitted, the current user is the 'Endorser'

lpsinger: I see. But can't we get this from DynamoDB?

dakota002: Hm I think we could, but we already do it in the loader method, with the getEndorsements('endorser') returning the email for display. I could update the updateEndorsementRequestStatus to just do a get on the dynamo table though, if that's better for this system

lpsinger: Let's do it. We need to validate _anything_ that we get from a form submission, even if it's in a hidden field that we populated. In this case validation would be checking that the email matches what's in our database. So we should just get it from our database in the first place.

app/routes/user/endorsements.server.ts
```diff
@@ -124,6 +128,21 @@ export class EndorsementsServer {
       ConditionExpression:
         'NOT (endorserSub = :endorserSub and requestorSub = :requestorSub and #status <> :status)',
     })
+
+    await sendEmail(
+      'GCN Endorsements',
```
jracusin, line 133: GCN Peer Endorsements - I think peer is key so the user understands why they're getting this.

app/routes/user/endorsements.server.ts
```diff
@@ -124,6 +128,21 @@ export class EndorsementsServer {
...
+      endorserEmail,
+      'New GCN Endorsement Request',
+      `You have a new endorsement request for NASA's General Coordinate Network (GCN) from ${
```
jracusin, line 136:
```suggestion
      `You have a new endorsement request for NASA's General Coordinates Network (GCN) from ${
```

jracusin, line 136:
```suggestion
      `You have a new peer endorsement request for NASA's General Coordinates Network (GCN) from ${
```

app/routes/user/endorsements.server.ts
```diff
@@ -124,6 +128,21 @@ export class EndorsementsServer {
...
+        this.#currentUserEmail
+      }. Approval of an endorsement means that the requestor, ${
+        this.#currentUserEmail
+      } is in good standing with the astronomy community and will permit them to submit GCN circulars. In addition, they will also be able to receive endorsement requests from other users.
+
```
jracusin, line 141:
```suggestion

Please approve this request if you are familiar with the requester, and agree with the criteria. Thank you for your contributions to the GCN community.

```

lpsinger (approved):

PR #646, opened 2023-03-01
Email spam virus checks

### LINKED ISSUE #638: Evaluate email authentication, spam, and virus scanning results for incoming GCN Circulars
Amazon SES provides a number of authentication, spam, and virus evaluation results in message headers and in SNS notification records. Determine which of these evaluations are relevant and check their results in the email-incoming lamba.

See https://docs.aws.amazon.com/ses/latest/dg/receiving-email-concepts.html#receiving-email-auth-and-scan.

dakota002:
resolves #638

### DIFF
--- a/app/email-incoming/index.ts
+++ b/app/email-incoming/index.ts
@@ -12,6 +12,7 @@ import {
 } from '@aws-sdk/client-cognito-identity-provider'
 import type { SNSEvent, SNSEventRecord } from 'aws-lambda'
 
+import type { ParsedMail } from 'mailparser'
 import { simpleParser } from 'mailparser'
 
 import {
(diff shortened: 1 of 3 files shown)


lpsinger:
What, if anything, should we do with `spfVerdict`, `dkimVerdict`, and `dmarcVerdict`?

app/email-incoming/index.ts
```diff
@@ -160,3 +165,11 @@ async function getLegacyUserData(
...
   )
 }
+
+function validateMessageIsNotSpam(messageReceipt: any) {
+  if (
+    messageReceipt.spamVerdict.status != 'PASS' ||
+    messageReceipt.virusVerdict.status != 'PASS'
+  )
+    throw new Error('Message caught in virus/spam detection.')
+}
```
lpsinger, line 175: I would prefer to keep this inline, inside the `handleRecord` function, so that we don't need the `any` type.

lpsinger (changes requested):

app/email-incoming/index.ts
dakota002: Can do! Will have that and the other verdicts included in the next commit

PR #663, opened 2023-03-08
Adds a function to parse events from subject lines

dakota002:


lpsinger (changes requested):
- The match should only be at the beginning of the subject line.
- We should not only capture the object name, but also attempt to normalize it somewhat to a a common format. For example, both `GRB123456A` and `GRB 123456A` should map to `GRB 123456A`.
- We need some unit tests! I'll add Jest so you can do that.

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +145,65 @@ export function formatAuthor({
...
+export function parseEventFromSubject(value: string): string {
+  const patterns = [
+    'GRB[.\\s|_|-]*\\d{6}[a-zA-Z|.]\\d*',
+    'SGR[.\\s|_|-]*\\d{4}\\.?\\d*\\+\\d{4}',
+    'IceCube[.\\s|_|-]*\\d{6}[A-Z]',
+    'ZTF[.\\s|_|-]*\\d{2}[a-zA-Z]*',
+    'HAWC[.\\s|_|-]*\\d{6}A',
+    'LIGO/Virgo[.\\s|_|-]*S\\d{6}[a-zA-Z]+',
+    'ANTARES[.\\s|_|-]*\\d{6}[a-zA-Z]',
+    'Baksan\\sNeutrino\\sObservatory\\sAlert[.\\s|_|-]*\\d{6}.\\d{2}',
```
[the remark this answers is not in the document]
dakota002: Cool cool, I will opt for 1 in the next commit

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +145,63 @@ export function formatAuthor({
   else if (!affiliation) return `${name} <${email}>`
   else return `${name} at ${affiliation} <${email}>`
 }
+
+export function parseEventFromSubject(value: string): string {
```
lpsinger, line 149: TS should be able to infer the return type OK for this function:

```suggestion
export function parseEventFromSubject(value: string) {
```

lpsinger (changes requested):

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +147,27 @@ export function formatAuthor({
...
+    [/LIGO\/Virgo[.\s|_|-]*(S\d{6}[a-zA-Z]+)/, ([, id]) => `LIGO/Virgo ${id}`],
+    [
+      /ANTARES[.\s|_|-]*(\d{6}[a-zA-Z])/,
+      ([, id]) => `ANTARES ${id}`.toUpperCase(),
+    ],
+    [
+      /Baksan\sNeutrino\sObservatory\sAlert[.\s|_|-]*(\d{6}.\d{2})/,
+      ([, id]) => `Baksan Neutrino Observatory Alert ${id}`,
+    ],
+  ]
```
lpsinger, line 167: Move to module scope, so that we don't redefine all of these anonymous functions on every invocation.

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +147,27 @@ export function formatAuthor({
...
+
+  for (const [regexp, normalize] of subjectMatchers) {
+    const match = regexp.exec(value)
+    if (match) return normalize(match)
+  }
```
lpsinger, line 172: First search for a match of any of the regexes at the _beginning_ of the string. Then search for a match of any of the regexes _anywhere_ in the string.

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +147,27 @@ export function formatAuthor({
   else if (!affiliation) return `${name} <${email}>`
   else return `${name} at ${affiliation} <${email}>`
 }
+
+export function parseEventFromSubject(value: string) {
+  const subjectMatchers: SubjectMatcher[] = [
+    [/GRB[.\s|_|-]*(\d{6}[a-zA-Z|.]\d*)/, ([, id]) => `GRB ${id}`],
```
lpsinger, line 153: In the delimiter part of the pattern, `[.\s|_|-]*`, does the `.` match any character? Or does it match a literal period?

Also: how about we do case-insensitive matching with the regexp `i` flag? That will make the parser a bit more forgiving, and also will eliminate the need to have `[a-zA-Z]`.

lpsinger (changes requested):

app/routes/circulars/circulars.lib.ts
```diff
@@ -143,3 +161,12 @@ export function formatAuthor({
...
   else return `${name} at ${affiliation} <${email}>`
 }
+
+export function parseEventFromSubject(value: string) {
+  for (const [regexp, normalize] of subjectMatchers) {
+    const startsWithMatch = RegExp('^' + regexp.source).exec(value)
+    if (startsWithMatch) return normalize(startsWithMatch)
+    const match = regexp.exec(value)
+    if (match) return normalize(match)
+  }
```
lpsinger, line 171: Split into two loops: we prefer _any_ match at the beginning of the string to _any_ match later in the string.

app/routes/circulars/circulars.lib.ts
```diff
@@ -20,6 +20,22 @@ export interface Circular extends CircularMetadata {
...
+  [/SGR[.\s_-]*(\d{4}\.?\d*\+\d{4})/i, ([, id]) => `SGR ${id}`],
+  [/IceCube[.\s_-]*(\d{6}[a-z])/i, ([, id]) => `IceCube ${id}`],
+  [/ZTF[.\s_-]*(\d{2}[a-z]*)/i, ([, id]) => `ZTF ${id}`],
+  [/HAWC[.\s_-]*(\d{6}A)/i, ([, id]) => `HAWC ${id}`],
+  [/LIGO\/Virgo[.\s_-]*(S\d{6}[a-z]+)/i, ([, id]) => `LIGO/Virgo ${id}`],
+  [/ANTARES[.\s_-]*(\d{6}[a-z])/i, ([, id]) => `ANTARES ${id}`.toUpperCase()],
+  [
+    /Baksan\sNeutrino\sObservatory\sAlert[.\s_-]*(\d{6}.\d{2})/i,
+    ([, id]) => `Baksan Neutrino Observatory Alert ${id}`,
+  ],
```
lpsinger, line 36: Normalize the case of all of these matches. For example:

```ts
...
[/LIGO\/Virgo[.\s_-]*S(\d{6}[a-z]+)/i, ([, id]) => `LIGO/Virgo S${id.toLowerCase()}`],
...
```

lpsinger (changes requested):

lpsinger (approved):

lpsinger:
@jracusin, would you please review this as well?

jracusin:
GCN Circ [33451](https://gcn.gsfc.nasa.gov/gcn3/33451.gcn3) Trigger 1159327: Swift detection of SGR1806-20 is another confusing case as the event is SGR 1806-20, but the subject focuses on the trigger.

lpsinger:
> GCN Classic handled this correctly.

Did it? Or did Scott or Teresa fix it?

jracusin:
This was submitted today, so it was automated.

lpsinger:
> This was submitted today, so it was automated.

OK, so @dakota002, would you add a unit test for that please?

jracusin:
> > This was submitted today, so it was automated.
> 
> OK, so @dakota002, would you add a unit test for that please?

But if it's not the first time the source was detected, and clearly not for this one, and you parse the string "SGR1806-20", is it actually okay?  Though note the missing space.

jracusin (approved):

PR #704, opened 2023-03-23
Adds flag to check for Submit/Receive-only users on legacy table

dakota002:
I'll work on a script to populate this, similar to the legacy submitters. 

Also will need an update to the post sign up to remove users from this table as they sign up for Cognito accounts


lpsinger (commented):
What about just a boolean flag on the `legacy_users` table that says whether they receive circulars?

dakota002:
That could work too! I could use a scan with a filter on that bool to get the receivers, and I think the submitter code would need to be updated too

lpsinger:
> I could use a scan with a filter on that bool to get the receivers, and I think the submitter code would need to be updated too

Remember that we always want to avoid scans because they are expensive both in terms of time and billing. You can add a global secondary that includes the boolean field as a partition or sort key so that you can query on it.

app/events/email-incoming/index.ts
```diff
@@ -32,6 +32,7 @@ interface UserData {
   sub?: string
   name?: string
   affiliation?: string
+  receiveOnly?: boolean
```
lpsinger, line 35: How about two flags, `submit` and `receive`? Users can be submit only, receive only, or submit and receive. Is that right, @jracusin?

lpsinger (changes requested):

app/events/email-incoming/index.ts
dakota002: Sure! I can add that now

lpsinger:
Would you please rebase this?

app.arc
```diff
@@ -81,6 +81,10 @@ circulars_subscriptions
   sub *String
   name circularsSubscriptionsBySub
 
+legacy_users
+  receive *Number
```
lpsinger, line 85: DynamoDB does have a boolean type. Does it support boolean fields as partition keys?

lpsinger (changes requested):

lpsinger (approved):

PR #736, opened 2023-04-03
Makes the announcement banner reusable with message and event props

dakota002:
How it would look:

image

How to use:

image

### DIFF
--- a/app/components/AnnounceBanner.tsx
+++ b/app/components/AnnounceBanner.tsx
@@ -16,10 +16,10 @@ type AnnouncedEvents = {
 
 export default function AnnounceBanner({
   message,
-  events,
+  children,
 }: {
   message: string
-  events?: AnnouncedEvents[]
+  children?: React.ReactNode
 }) {
   const [showFullBanner, setShowFullBanner] = useState(true)
 
@@ -55,13 +55,7 @@ export default function AnnounceBanner({
             className="usa-banner__content usa-accordion__content padding-top-2"
             id="announcement-banner-blue"
           >
-            {events && (
-              <div className="grid-row">
-                {events.map((event) => (
-                  <AnnouncementEvent key={event.link} {...event} />
-                ))}
-              </div>
-            )}
+            {children}
             <p>
               If you cannot attend live, then you can get the{' '}
               <Link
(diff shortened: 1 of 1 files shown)


app/components/AnnounceBanner.tsx
```diff
@@ -46,30 +55,19 @@ export default function AnnounceBanner() {
...
+              <div className="grid-row">
+                {events.map((event) => (
+                  <div key={event.link} className="mobile-lg:grid-col-4">
+                    {event.time}
+                    <div>(best for {event.region}):</div>
+                    <Link rel="external" href={event.link}>
+                      {event.link}
+                    </Link>
+                  </div>
+                ))}
```
lpsinger, line 68: It would be simpler to extract this as another component.

lpsinger (changes requested):

app/components/AnnounceBanner.tsx
dakota002: Will do!

dakota002:
Revised usage example:
image

app/components/AnnounceBanner.tsx
```diff
@@ -19,7 +19,7 @@ export default function AnnounceBanner({
   events,
 }: {
   message: string
-  events?: announcedEvents[]
+  events?: AnnouncedEvents[]
```
lpsinger:
```suggestion
  children?: AnnouncedEvents[]
```

app/components/AnnounceBanner.tsx
```diff
@@ -84,3 +78,15 @@ export default function AnnounceBanner({
     </section>
   )
 }
+
+export function AnnouncementEvent(event: AnnouncedEvents) {
```
lpsinger, line 76:
```suggestion
export function AnnouncementEvent(props: AnnouncedEvents) {
```

Also just inline the type here in the function signature.

app/components/AnnounceBanner.tsx
```diff
@@ -86,3 +66,19 @@ export default function AnnounceBanner() {
...
   )
 }
+
+export function AnnouncementEvent(props: {
+  time: string
+  link: string
+  region: 'Atlantic' | 'Pacific' | 'Asia and Oceania'
+}) {
+  return (
+    <div className="mobile-lg:grid-col-4">
```
lpsinger, line 76:
```suggestion
    <Grid mobileLg={{col: 4}}>
```

lpsinger (changes requested):

PR #755, opened 2023-04-07
updated subject keywords and not case sensitive

jracusin:
keywords now agree with GCN Classic, and we removed the case sensitivity

### DIFF
--- a/app/routes/circulars/circulars.lib.ts
+++ b/app/routes/circulars/circulars.lib.ts
@@ -76,7 +76,9 @@ export function subjectIsValid(subject: string) {
   if (subject.length)
     return (
       !emailIsAutoReply(subject) &&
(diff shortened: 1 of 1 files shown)


app/routes/circulars/circulars.lib.ts
```diff
@@ -76,7 +76,9 @@ export function subjectIsValid(subject: string) {
   if (subject.length)
     return (
       !emailIsAutoReply(subject) &&
-      validSubjectKeywords.some((x) => subject.startsWith(x))
+      validSubjectKeywords.some((x) =>
+        subject.toLowerCase().startsWith(x.toLowerCase())
```
lpsinger, line 80: Lift `subject.toLowerCase()` out of the loop.

lpsinger (changes requested):

lpsinger (approved):

PR #757, opened 2023-04-07
Process emails forwarded from capella2

lpsinger:
- Email addresses from non-NASA addresses are rewritten from e.g. `Albert Einstein <albert.einstein@ligo.org>` to `"Albert Einstein <albert.einstein@ligo.org> via gcncirc" <mailnull@capella2.gsfc.nasa.gov>`. Parse out the original email address.
- Send confirmation addresses to the Reply-to address, if one is present. (This is standard practice for _all_ mail replies.)
- Since Reply-to may contain multiple addresses, we also need our generic sendEmail function to take an array of recipients.


app/events/email-incoming/index.ts
```diff
@@ -51,6 +53,42 @@ const cognito = new CognitoIdentityProviderClient({})
...
+ *
+ * Messages forwarded by capella2 from non-NASA addresses have From headers
+ * that are rewritten like this:
+ *
+ * From: Albert Einstein <albert.einstein@ligo.org>
+ * --rewritten to--
+ * From: "Albert Einstein <albert.einstein@ligo.org> via gcncirc" <mailnull@capella2.gsfc.nasa.gov>
+ *
+ */
+export function getFromAddress(fromAddressObject?: AddressObject) {
```
lpsinger, line 70: Oh right! I exported this so that I could write a unit test for it. I forgot to write that test. Will do shortly.

dakota002 (approved):

PR #775, opened 2023-04-10
Makes scope visible in Credential Cards

### LINKED ISSUE #760: Show credential scope to user for easier credentials management
For credentials that have already been created by a user, and shown on https://gcn.nasa.gov/user/credentials
there is no clear way to determine what scope they have!

eg, when testing I had to create a whole new credential to ensure that an issue was/wasn't caused by improper scope.

dakota002:
image

resolves #760

### DIFF
--- a/app/components/CredentialCard.tsx
+++ b/app/components/CredentialCard.tsx
@@ -25,6 +25,7 @@ export default function CredentialCard({
   name,
   client_id,
   created,
+  scope,
 }: RedactedClientCredential) {
   const ref = useRef<ModalRef>(null)
   const fetcher = useFetcher()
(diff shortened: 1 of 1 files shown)


lpsinger:
This changes the sizing of the box so that the Delete and Select buttons are no longer centered with the text.

How would it look if we decreased the size of the name ("Show me scope" here) to match the rest of the text, but kept it bold? Would that get us back to the original metrics?

dakota002:
@lpsinger Here is an example with the smaller text and centered buttons:
image

lpsinger:
That looks great. Can you post screenshots on a few different device sizes including mobile and tablet?

dakota002:
Mobile:
image

Tablet:
image

lpsinger (approved):

PR #785, opened 2023-04-11
Add documentation for new GCN Circulars

lpsinger:


dakota002 (approved):

jracusin:
On the archive page, there was previously a mention of the GCN Viewer.  It's still a useful resource and an archive of circulars and notices, even if we still need to build a new version integrated into this system.

lpsinger:
> On the archive page, there was previously a mention of the GCN Viewer. It's still a useful resource and an archive of circulars and notices, even if we still need to build a new version integrated into this system.

I don't think that the Circulars documentation is the right place to advertise the GCN Viewer. In any case, the GCN Viewer is still mentioned on the front page: https://heasarc.gsfc.nasa.gov/wsgi-scripts/tach/gcn_v2/tach.wsgi/

lpsinger:
OK, @jracusin, I added back the text about the GCN Viewer.

Vidushi-GitHub:
I have a slight suggestion about GCN Circulars archive, it's sharp blue color, similar as old one. Shall we replace it with ADS archive blue color?

lpsinger:
> I have a slight suggestion about GCN Circulars archive, it's sharp blue color, similar as old one. Shall we replace it with ADS archive blue color?

This is just the theme link color from nasawds/USWDS. If we change it, then it won't match the other links on our web site.

PR #800, opened 2023-04-14
Expanding Some Code Coverage

dakota002:
100% Announcement banner example


lpsinger (commented):
Very nice!

Since we will now have a dependency on babel, should we switch away from ts-jest?

dakota002:
Hmm I think that might warrant some studying on my part before I can say for certain. I think the two are working together, but I definitely need to dig deeper

lpsinger (approved):

PR #834, opened 2023-04-19
Makes visible the disallowed keywords

dakota002:
Updates the UI in circulars/new:
image

And in the styleguide:
image


dakota002:
resolves #830

lpsinger (approved):

PR #835, opened 2023-04-19
Filters groups provided to client_credentials server constructor

dakota002:
image

### DIFF
--- a/app/routes/user/client_credentials.server.ts
+++ b/app/routes/user/client_credentials.server.ts
@@ -48,7 +48,7 @@ export class ClientCredentialVendingMachine {
     if (!user) throw new Response('not signed in', { status: 403 })
     return new this(
       user.sub,
-      user.groups.filter((x) => x.includes('kafka'))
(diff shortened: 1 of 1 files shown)


app/routes/user/client_credentials.server.ts
```diff
@@ -46,7 +46,10 @@ export class ClientCredentialVendingMachine {
   static async create(request: Request) {
     const user = await getUser(request)
     if (!user) throw new Response('not signed in', { status: 403 })
-    return new this(user.sub, user.groups)
+    return new this(
+      user.sub,
+      user.groups.filter((x) => x.includes('kafka'))
```
lpsinger, line 51: How about startsWith?

lpsinger (changes requested):

app/routes/user/client_credentials.server.ts
dakota002: Sure, I can make that the implementation, will change it now

lpsinger (approved):

PR #840, opened 2023-04-20
Switch to Remix v2 route convention

### LINKED ISSUE #831: Migrate to Remix V2 route module paths
This will simplify implementing #372.

lpsinger:
Fixes #831.


lpsinger:
@dakota002, this is obviously a huge diff and it has merge conflicts at the moment, but when you have a chance would you check this out locally and tell me what you think of this new route module convention? I am on the fence about it.

dakota002:
Oh I briefly read over the docs that mentioned this transition a few days ago. I'm kinda indifferent about this format. My only negative on it is that we lose some of the organizational features of the tree view (have to scroll through the files and not being able to collapse them down):
image

lpsinger:
> My only negative on it is that we lose some of the organizational features of the tree view (have to scroll through the files and not being able to collapse them down):

In this last push, I added `"explorer.sortOrder": "mixed"` to .vscode/settings.json, which fixes that.

This is now up to date and ready for review. Please take it for a test drive, and see if anything is broken!

dakota002:
Is the naming convention not supported for mdx files? I see the mission pages all have folders still

dakota002:
Also i think it might be worth expanding more on the .vscode/settings.json to something like the following to allow us to fold down the file nesting:

image

Folded down to docs.tsx
image

The pattern matching is fairly limited, but it should make for a slightly better dev experience

lpsinger:
> Is the naming convention not supported for mdx files? I see the mission pages all have folders still

It is supported. I created folders for most of those to group their images with the routes.

lpsinger:
> Also i think it might be worth expanding more on the .vscode/settings.json to something like the following to allow us to fold down the file nesting:
> 
> ...
> 
> The pattern matching is fairly limited, but it should make for a slightly better dev experience.
> 
> The added lines:
> 
> ```
> "explorer.fileNesting.enabled": true,
> "explorer.fileNesting.patterns": {
>     "circulars.*":"${basename}.*",
>     "docs.*": "${basename}.*",
>     "quickstart.*": "${basename}.*",
>     "user.*":"${basename}.*" 
> }
> ```

That's really neat! Unfortunately, it doesn't group together both files and folders. For example, in the screen shot, you can see that the docs.tsx file and docs.history folder are not grouped. So I would opt not to.

dakota002:
Yea, it's really limited with how much folding it does, it would be great if the pattern matching could be done more granularly.

But overall the site seems to be working as expected, looks good to me

PR #860, opened 2023-04-24
New dev banner

### LINKED ISSUE #790: Add dev banner to emails
Email notifications from the dev and test environments should be clearly distinguished from production. Add something similar to the web site's development banner to all emails.

dakota002:
Closes #790

### DIFF
--- a/app/lib/email.server.ts
+++ b/app/lib/email.server.ts
@@ -42,6 +42,11 @@ function getBaseMessage({
   subject,
   body,
 }: BaseMessageProps): Omit<SendEmailCommandInput, 'Destination'> {
+  if (getHostname().startsWith('dev') || getHostname().startsWith('test')) {
+    body =
+      `******** This notification is from the internal development version of GCN. For the production version, go to https://gcn.nasa.gov/ ********
+    ` + body
+  }
   return {
     FromEmailAddress: `${fromName} <no-reply@${getHostname()}>`,
     ReplyToAddresses: replyTo,


app/lib/email.server.ts
```diff
@@ -42,6 +42,11 @@ function getBaseMessage({
   subject,
   body,
 }: BaseMessageProps): Omit<SendEmailCommandInput, 'Destination'> {
+  if (getHostname().startsWith('dev') || getHostname().startsWith('test')) {
```
lpsinger, line 45: Refactor so that `getHostname()` is only called once.

app/lib/email.server.ts
```diff
@@ -42,6 +42,11 @@ function getBaseMessage({
...
+    body =
+      `******** This notification is from the internal development version of GCN. For the production version, go to https://gcn.nasa.gov/ ********
```
lpsinger, line 47: We have some code here that generates different dev banner text depending on the hostname here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/components/DevBanner.tsx#L21-L33

Please refactor to use this code in both places.

lpsinger (changes requested):

lpsinger (changes requested):
Remember that `getHostname()` will _only_ work on the server side, and `useHostname()` will _only_ work in a React component.

dakota002:
That is the case, getHostname is only being used in the .ts files and useHostname is in the tsx.

lpsinger:
> That is the case, getHostname is only being used in the .ts files and useHostname is in the tsx.

No, it's not. You have this line in DevBanner.tsx:

```typescript
const { heading, description } = getEnvBannerHeaderAndDescription()
```

And `getEnvBannerHeaderAndDescription()` calls `getHostname()`, which in turn inspects `process.env`, which isn't set on the browser.

dakota002:
Ah okay I see, fixing now

lpsinger (approved):

PR #894, opened 2023-05-03
Schema browser

dakota002:


app/routes/schema-browser/index.tsx
```diff
@@ -0,0 +1,28 @@
+import { Link } from '@remix-run/react'
+
+export default function () {
+  return (
+    <>
+      <h1>Schema Browser</h1>
+      <p>
+        Browse the schema definitions for GCN Notices as distributed by GCN
+        Kafka. Navigate through the directory tree and select a schema for more
```
jracusin, line 9: Not sure how to explain it, but the IGWN JSON schema don't follow the unified schema, so we just need to make it clear that this is for other types, and link to the IGWN schema docs:  https://emfollow.docs.ligo.org/userguide/content.html#

dakota002:
# General Overview of updates:

As a note, this is not the latest version of the schema. When the tag goes out, we can update the package and the content will follow suit

## Notices is a dropdown:
image

Notice Reference leads to the /notices page

Schema Browser leads to /schema-browser
image


## Side nav is expandable for browsing the schema. Generated buttons from @nasa-gcn/gcn-schema installed w/ npm
image

## Page content generates based on the Title, Description, allOf, anyOf, oneOf, defs, etc:
image

image

## Rendered Examples:
image

lpsinger (changes requested):
Initial comments:

- [ ] Replace `<p>` elements with `<p className="usa-paragraph">`.
- [ ] Note that by including the schema as an npm package, we will need to update the lock file and redeploy the web site for schema changes to be live. Is that what we want?

dakota002:
> Initial comments:
> 
> * [x]  Replace `<p>` elements with `<p className="usa-paragraph">`.
> * [ ]  Note that by including the schema as an npm package, we will need to update the lock file and redeploy the web site for schema changes to be live. Is that what we want?

I tried implementing a solution that pulled the data from github previously, but there were more headaches in that method then I think were worth for the time being.

.gitignore
```diff
@@ -19,3 +19,4 @@ sam.yaml
 
 .env
 .eslintcache
+.schemaTree.json
```
lpsinger, line 22: Where does this file come from?

app/routes/schema-browser.tsx
```diff
@@ -0,0 +1,125 @@
...
+type SchemaTreeItem = {
+  name: string
+  path: string
+  children?: SchemaTreeItem[]
+}
+
+export async function loader() {
+  if (!feature('SCHEMA')) throw new Response(null, { status: 404 })
+  const localDataTree = dirTree(
+    '../../node_modules/@nasa-gcn/schema/gcn/notices'
```
lpsinger, line 26: Are you sure that this is included in the server bundle? Check the settings in https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/remix.config.mjs.

app/routes/schema-browser/$.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+import { loadJson, loadSchemaExamples } from '../../lib/schema-data'
+import { Highlight } from '~/components/Highlight'
+import type { Schema, SchemaProperty } from '~/components/SchemaBrowserElements'
+import { SchemaPropertiesTableBody } from '~/components/SchemaBrowserElements'
+import { ReferencedElementTable } from '~/components/SchemaBrowserElements'
+import { Tab, Tabs } from '~/components/Tabs'
+
+export async function loader({ params: { '*': path } }: DataFunctionArgs) {
+  if (!path) throw new Response('path must be defined', { status: 400 })
```
lpsinger, line 20:
```suggestion
  if (!path) throw new Response(null, { status: 404 })
```

app/components/SchemaBrowserElements.tsx
```diff
@@ -0,0 +1,141 @@
...
+              <thead>
+                <tr>
+                  <th>Name</th>
+                  <th>Type</th>
+                  <th>Description</th>
+                </tr>
+              </thead>
+              <SchemaPropertiesTableBody schema={item.schema} />
+            </Table>
+          </td>
```
lpsinger, line 71: Rather than a nested table, can we have a single table with inherited fields indented?

lpsinger (changes requested):

.gitignore
dakota002: This was from local testing, I will delete that

app/components/Header.tsx
```diff
@@ -90,9 +91,43 @@ export function Header() {
...
+                      >
+                        Notice References
+                      </Link>,
+                      <Link
+                        key="schema"
+                        to="/schema-browser"
+                        onClick={() => setNoticeMenuIsOpen(!noticeMenuIsOpen)}
+                      >
+                        Schema Browser
+                      </Link>,
```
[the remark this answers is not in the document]
dakota002: That works, I'll reorganize

app/components/SchemaBrowserElements.tsx
dakota002: Probably, do you have any reference or something I could use as an example for what you are looking for?

lpsinger: Sure, something like the macOS file browser. Notice how subfolders have their names indented, but the other columns are not.

Screenshot 2023-05-23 at 14 22 21

lpsinger:
> > Note that by including the schema as an npm package, we will need to update the lock file and redeploy the web site for schema changes to be live. Is that what we want?
> 
> I tried implementing a solution that pulled the data from github previously, but there were more headaches in that method then I think were worth for the time being.

If we `npm install` it, then how will we display multiple versions? I don't know a way to depend upon multiple versions of the same project.

dakota002:
> > > Note that by including the schema as an npm package, we will need to update the lock file and redeploy the web site for schema changes to be live. Is that what we want?
> > 
> > 
> > I tried implementing a solution that pulled the data from github previously, but there were more headaches in that method then I think were worth for the time being.
> 
> If we `npm install` it, then how will we display multiple versions? I don't know a way to depend upon multiple versions of the same project.

We have a couple options:
we could use the tags to install the package with an alias when we update https://stackoverflow.com/questions/56134857/how-to-install-npm-package-under-alias-or-different-name

or we could do something that pulls from github. The main problem with pulling from github was running into the request limit for their  api. I had implemented a caching method that mostly resolved this in a previous commit and could be worth revisiting, but I think that would be best considered as a future update when there are actually more than one tagged versions.

dakota002:
> > or we could do something that pulls from github. The main problem with pulling from github was running into the request limit for their api.
> 
> Were you making the REST API calls anonymously, or were you using an access token? According to https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting, personal access tokens have a rate limit of 5,000 requests per hour.
> 
> We should declare public, static cache headers for tagged versions --- limiting origin requests to those resources at approximately 1 per year per page per region.

Anonymously at first, but I got an access token and used it with the caching option once I realized how quickly it was burning through requests.  I'm all for this route, but I still think it is a further enhancement

app/routes/schema-browser.tsx
dakota002: Would that just be an update to the serverDependenciesToBundle var to this `[/^(?!@?aws-sdk\/)/, /^(@nasa-gcn\/schema\/)/]`

app/components/SchemaBrowserElements.tsx
dakota002: What do you think of this?
image

Mobile:
image

lpsinger: I like that. But add disclosure arrows to expand/collapse the sub tables.

dakota002: How is this?
image

dakota002:
> > > > or we could do something that pulls from github. The main problem with pulling from github was running into the request limit for their api.
> > > 
> > > 
> > > Were you making the REST API calls anonymously, or were you using an access token? According to https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting, personal access tokens have a rate limit of 5,000 requests per hour.
> > > We should declare public, static cache headers for tagged versions --- limiting origin requests to those resources at approximately 1 per year per page per region.
> > 
> > 
> > Anonymously at first, but I got an access token and used it with the caching option once I realized how quickly it was burning through requests. I'm all for this route, but I still think it is a further enhancement
> 
> OK, that's fine. I'd like to see how that will work, though. Which of the following courses would you like to take?
> 
> 1. Implement browsing of multiple schema versions in this PR, or
> 2. Review and merge this PR on its merits and keep it hidden behind a feature flag until the multiple versions are implemented?

I am on the side of 2, but I would also argue that the multiple versions feature should be a later feature and not be a reason to keep it hidden.

lpsinger:
> > > > > or we could do something that pulls from github. The main problem with pulling from github was running into the request limit for their api.
> > > > 
> > > > 
> > > > Were you making the REST API calls anonymously, or were you using an access token? According to https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting, personal access tokens have a rate limit of 5,000 requests per hour.
> > > > We should declare public, static cache headers for tagged versions --- limiting origin requests to those resources at approximately 1 per year per page per region.
> > > 
> > > 
> > > Anonymously at first, but I got an access token and used it with the caching option once I realized how quickly it was burning through requests. I'm all for this route, but I still think it is a further enhancement
> > 
> > 
> > OK, that's fine. I'd like to see how that will work, though. Which of the following courses would you like to take?
> > 
> > 1. Implement browsing of multiple schema versions in this PR, or
> > 2. Review and merge this PR on its merits and keep it hidden behind a feature flag until the multiple versions are implemented?
> 
> I am on the side of 2, but I would also argue that the multiple versions feature should be a later feature and not be a reason to keep it hidden.

OK. I'll do a review by Monday. (Sorry, busy with travel this week.)

app/components/Header.tsx
```diff
@@ -100,9 +101,43 @@ export function Header() {
...
+                    ]}
+                  />
+                </>
+              ) : (
+                <NavLink className="usa-nav__link" to="/docs" key="/docs">
+                  Documentation
+                </NavLink>
+              ),
```
lpsinger, line 140: I'd rather that the Documentation nav button not be a dropdown right now.

The user nav button is a dropdown, but the dropdown menu items exactly match the sidenav items on the /user route. And it's a small number of menu items.

On the other hand, the Documentation section has a large (and likely to grow larger still) number of menu items.

lpsinger (changes requested):
Move /routes/schema-browser to /routes/docs/schema, and add to /docs sidenav.

app/components/Header.tsx
lpsinger: Well... on the other hand, there will be a schema for circulars, which are not notices.

OK, I guess I'm alright with putting it under /notices.

dakota002: Ah thats a good point, I forgot about the circulars ones. That does make it seem like it would be better under docs. That makes a lot more sense now, I will update it to be under docs and add update the other nav stuff

dakota002: Small update, the docs header nav is no longer a dropdown, and i added a side nav option to get to the schema-browser and a back button. I did not move the files though. In trying to move /routes/schema-browser to /routes/docs/schema it broke, the side nav. It only shows the doc level side nav

dakota002:
> I keep flip-flopping about where on the site this should go --- under Documentation or under Notices. Either way, it should be reachable from the dropdown menu and (if we are being consistent with other existing pages) the sidenav. Where does that leave us?

I can make it navigable via the dropdown, and I do think it makes more sense for it to be under the Documentation dropdown if we are anticipating the circulars to have schema as well. As for side nav, I can make it navigable from there, but the route will still need to be /schema-browser so that the folder nav doesnt get messed up

Schema-browser under /docs :
image

Schema-browser as its own directory:
image


While it would add some extra work, if we need this to exist under the /docs/ path, we could implement some changes to either collapse or hide the parent side nav, or possibly reduce the margins to make more use of the extra space on the screen

lpsinger:
Oh, I like the "Back" arrow. Now it makes sense. Nice and simple! I like the screen shots, and I am on board with it being under Documentation. If that's the case, then I'd like that to be reflected in the route path too.

This patch introduces several new components that are only used in the schema browser. In the interest of greater modularity and in the spirit of the coming Remix v2 route module convention, would you please move those new special-purpose components alongside the route module file, with the extension `.lib.ts` or `lib.tsx` to exclude them from consideration as route modules?

dakota002:
Sure, working on some refactors now!

dakota002:
@lpsinger what are your thoughts on this, I changed the back button to hide/display the side nav for the docs pages when viewing the schema-browser
image

image

lpsinger:
> @lpsinger what are your thoughts on this, I changed the back button to hide/display the side nav for the docs pages when viewing the schema-browser

Oof, no, I don't like the two side navs side by side. I liked the back button. Is the problem that you don't know how to break out of the /docs.tsx layout route? There is a simple fix for that if we switch to the v2 route module convention because it supports [nested URLs without layout nesting](https://remix.run/docs/en/main/file-conventions/route-files-v2#nested-urls-without-layout-nesting).

lpsinger:
> I'll give that a read, my thought was that it would be nice to still have access to the docs nav, but I don't really like the squeezed nav either.
> 
> I can change it back to just be the back button for now though, or i can make it a togglable option to switch between the docs nav and the schema nav, thoughts on either of those routes?

I like the back button.

To implement that cleanly and have the schema browser live at the URI /docs/schema-browser, do we need to adopt the v2 route module convention?

dakota002:
> > I'll give that a read, my thought was that it would be nice to still have access to the docs nav, but I don't really like the squeezed nav either.
> > I can change it back to just be the back button for now though, or i can make it a togglable option to switch between the docs nav and the schema nav, thoughts on either of those routes?
> 
> I like the back button.
> 
> To implement that cleanly and have the schema browser live at the URI /docs/schema-browser, do we need to adopt the v2 route module convention?

Sounds good, the new convention would definitely make this task easier, but I can also make it so without that change. I would opt for doing it without that change first so that we can keep the overall PR more focused and then later when we make that switch I can just clean up the hacky way

app/routes/docs/schema-browser.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return { localDataTree }
+}
+
+export default function Schema() {
+  const { localDataTree } = useLoaderData<typeof loader>()
+
+  const items: React.ReactNode[] = ([localDataTree] as SchemaTreeItem[])
+    .filter((x) => !x.name.includes('.example.json'))
+    .map(RenderSchemaTreeItem)
```
lpsinger, line 37: You can do this filtering in the loader to reduce the amount of data that is sent to the browser.

app/components/Header.tsx
```diff
@@ -100,9 +101,43 @@ export function Header() {
...
+                    id="docs"
+                    isOpen={docsMenuIsOpen}
+                    items={[
+                      <Link
+                        key="docs"
+                        to="/docs"
+                        onClick={() => setDocsMenuIsOpen(!docsMenuIsOpen)}
+                      >
+                        Documentation
+                      </Link>,
```
lpsinger, line 125: So there the nav drop down item Documentation has a menu item called Documentation? That doesn't make sense.

lpsinger (changes requested):

app/components/Header.tsx
dakota002: Do you have a better name in mind? I just copied the original button into there. /docs brings the user to the About page, so  maybe About?

lpsinger: No... maybe just don't make it a dropdown for now.

app/routes/docs/schema-browser/SchemaBrowserElements.lib.tsx
```diff
@@ -0,0 +1,139 @@
...
+}: {
+  schema: Schema
+  indent?: boolean
+}) {
+  return (
+    <>
+      {schema.properties &&
+        Object.keys(schema.properties).map((itemKey) => (
+          <tr key={itemKey}>
+            <th scope="row" className={indent ? 'indent' : ''}>
```
lpsinger, line 102: USWDS already provides a utility class for this:

```suggestion
            <th scope="row" className={indent ? 'text-indent-2' : ''}>
```

lpsinger (changes requested):

app/routes/docs/schema-browser/SchemaBrowserElements.lib.tsx
dakota002: Thats handy, I'll commit this and check it out locally to make sure it looks right

lpsinger (approved):

lpsinger (approved):

PR #898, opened 2023-05-05
network outage news item

jracusin:


app/root.tsx
```diff
@@ -219,6 +219,7 @@ function Document({ children }: { children?: React.ReactNode }) {
         <DevBanner />
         <Header />
         <NewsBanner message="GCN Circulars are now part of the new GCN!" />
+        <NewsBanner message="GCN Classic outage due to local network maintenance - May 6 12:00-19:00 UTC." />
```
lpsinger, line 222: Remove old banner.

```suggestion
        <NewsBanner message="GCN Classic outage due to local network maintenance - May 6 12:00-19:00 UTC." />
```

lpsinger (changes requested):

lpsinger (approved):

PR #937, opened 2023-05-18
Send Circulars using To addresses, not Bcc

lpsinger:
There is one institution that is bouncing our messages because they don't have any To: or Cc: recipients set. I don't know if that is a common spam signal or not, but by setting the To: address we should at least unblock the messages for that institution.

I added an SES template that simply has a subject and a body field.


lpsinger:
@dakota002, please go ahead and review this, but don't merge it yet. I'm going to manually deploy it to dev to try it out.

Issue #938, opened 2023-05-18
New producers documentation does not have correct topic name pattern


lpsinger:
The topics should start with `gcn.notices.mission`.

PR #942, opened 2023-05-19
corrected topic format

jracusin:
addresses #938


lpsinger (changes requested):
Also update the code sample on this page.

lpsinger (approved):

Issue #923, opened 2023-05-10
"Back" after a search takes you to full list rather than your search output


jracusin:

lpsinger:
This would be a great use for react-router's [useNavigate](https://reactrouter.com/en/main/hooks/use-navigate) hook.

lpsinger:
Fixed by #967.

PR #967, opened 2023-05-22
[nasa-gcn/gcn.nasa.gov#923] back now returns to search output

tylerbarna:
search value is included in url string when navigating from a search, back button now points back to that search. When returning to main circulars page from an object without searching, full object list is shown


app/routes/circulars/index.tsx
```diff
@@ -214,7 +214,7 @@ export default function () {
           <ol>
             {allItems.map(({ circularId, subject }) => (
               <li key={circularId} value={circularId}>
-                <Link to={`/circulars/${circularId}`}>{subject}</Link>
+                <Link to={`/circulars/${circularId}?query=${query ?? ''}`}>{subject}</Link>
```
lpsinger, line 217: - [ ] We need to save both the query string and page number.
- [ ] I suggest that we lift the generation of the query string out of the loop, to avoid repeatedly evaluating it on every iteration. Like this:

```ts
  const [searchParams] = useSearchParams()
  let searchParamsString = searchParams.toString()
  if (searchParamsString) searchParamsString = `?${searchParamsString}`
```

Then, inside the loop:

```jsx
<Link to={`/circulars/${circularId}${searchParamsString}`}>
```

tylerbarna: ah, I neglected to consider the page number, thanks. I'll change the query generation as you suggest as well.

lpsinger:
Thank you, and congratulations on your first PR!

Issue #943, opened 2023-05-19
Include circular number in search criteria


jracusin:
Currently searches the subject and body, but not the number itself.  So when you search for a circular by number, it results in a list of the circulars that reference that circular number, but not the circular of that number itself.

PR #973, opened 2023-05-22
[nasa-gcn/gcn.nasa.gov#943] redirect to circular page when searching specific ID

tylerbarna:
resolves #943 

in the circular search, users can prepend their search with 'circular:' followed by the specific circular number, which will redirect to the specific item webpage

PR #976, opened 2023-05-23
[nasa-gcn/gcn.nasa.gov#943] redirect to circular page when searching specific ID

tylerbarna:
resolves https://github.com/nasa-gcn/gcn.nasa.gov/issues/943

in circulars search, users can prepend their search with "circular:" followed by a specific circular number to redirect to the specific circular webpage

Also addressed changes regarding topic branching and pre-commit hook issues (specifically newline format and prettier) requested in https://github.com/nasa-gcn/gcn.nasa.gov/pull/973

### DIFF
--- a/app/routes/circulars/index.tsx
+++ b/app/routes/circulars/index.tsx
@@ -213,10 +213,10 @@ export default function () {
         </Link>
       </ButtonGroup>
       <Hint id="searchHint">
-        Search for Circulars by submitter, subject, or body text (e.g. 'Fermi GRB').
-        <br />
-        To navigate to a specific circular, enter the associated Circular
-        ID (e.g. 'gcn123', 'Circular 123', or '123').
+        Search for Circulars by submitter, subject, or body text (e.g. 'Fermi
+        GRB'). <br />
+        To navigate to a specific circular, enter the associated Circular ID
+        (e.g. 'gcn123', 'Circular 123', or '123').
       </Hint>
       {clean && (
         <>


lpsinger (changes requested):
Let's recognize the following styles of search strings as references to GCN Circulars:

```
12345
GCN 12345
GCN12345
GCN Circular 12345
GCNCircular12345
Circular 12345
ciRCular12345
```

i.e., the words `GCN` and `Circular` may be present, with or without whitespace between them, with any capitalization, plus a number. Consider using a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions).

Add a [Hint](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/components/Hint.tsx) component below the search box to give examples of valid search strings.

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,13 @@ export async function put(subject: string, body: string, request: Request) {
     submitter: formatAuthor(user),
   })
 }
+
+export async function circularRedirect(query: string | undefined) {
+  if (!query) return
+  const validCircularSearchStyles = /(GCN\s*|Circular\s*|(^[0-9]+$))/i
```
lpsinger, line 179: * [ ] This should match the entire string from beginning to end, so the regex should start with `^` and end with `$`.
* [ ] We don't care about leading or trailing whitespace, so add `\s*` between each token.
* [ ] Use an optional, [non-capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Groups_and_backreferences#using_groups) to make it so that the strings "gcn" and "circular" are fully optional. A non-capturing group has the syntax `(?:subpattern)`, and to make it optional, you can suffix it with a `?`, as in `(?:subpattern)?`.
* [ ] Do have a capturing group for the Circular number itself. Note that for silly historical reasons, there are a handful of Circulars with negative or half-integer IDs. `(-?\d+(?:\.\d)?)`

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,13 @@ export async function put(subject: string, body: string, request: Request) {
...
+  if (validCircularSearchStyles.test(query)) {
+    const circularId = query.replace(/\D/g, '')
```
lpsinger, line 181: In your regex, having surrounded the actual circular number in a capturing group, you can simplify the test for a matching circular ID using optional chaining like this:

```ts
const circularId = validCircularSearchStyles.exec(query)?.[0]
if (circularId) {
   // do redirect
}
```

app/routes/circulars/index.tsx
```diff
@@ -206,6 +208,12 @@ export default function () {
           </Button>
         </Link>
       </ButtonGroup>
+      <p className="usa-hint">
+        <Hint>
+          To navigate to a specific circular, search for the Circular ID
+          ('gcn123', 'Circular 123', '123', etc.)
```
lpsinger, line 214: Also give instructions and examples for ordinary full-text search.

lpsinger (changes requested):
Currently, if the user enters a circular ID that does not exist, then they will be redirected anyway and get a 404 error. Upon searching for a circular ID that does not exist, a preferable behavior would be to execute an ordinary search.

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,19 @@ export async function put(subject: string, body: string, request: Request) {
...
+  const validCircularSearchStyles =
+    /^(?:\s*GCN\s*)?(?:\s*CIRCULAR\s*)?((\s*-?\d+(?:\.\d)?)\s*)$/i
+  const circularID = validCircularSearchStyles.exec(query)?.[1]
+  if (circularID) {
+    const db = await tables()
+    const result = await db.circulars.get({
+      circularId: Number(circularID),
+    })
+    if (!result) return
+    const circularURL = `/circulars/${circularID}`
```
lpsinger, line 188: Very slightly terser:

```suggestion
  const circularId = parseFloat(validCircularSearchStyles.exec(query)?.[1])
  if (!isNaN(circularID)) {
    const db = await tables()
    const result = await db.circulars.get({ circularId })
    const circularURL = `/circulars/${circularId}`
```

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
tylerbarna: line 187 of the commit is handling the case where the circular Id doesn't exist, I believe the omission of that will throw an error if you try to search a circular Id that doesn't exist. I was able to remove the first line of the function (the conditional where query is undefined) by tweaking your line 181 to be 
`const circularID = parseFloat(validCircularSearchStyles.exec(query)?.[1]) ?? NaN`
and using your suggested conditional on line 182

lpsinger: `parseFloat` returns NaN if its argument is null, undefined, or an invalid string --- so there is no need for the `?? NaN`.

lpsinger: Ah, but you're right, you do need the `if (!result) return` line.

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,19 @@ export async function put(subject: string, body: string, request: Request) {
...
+
+export async function circularRedirect(query: string | undefined) {
+  if (!query) return
+  const validCircularSearchStyles =
+    /^(?:\s*GCN\s*)?(?:\s*CIRCULAR\s*)?((\s*-?\d+(?:\.\d)?)\s*)$/i
+  const circularID = validCircularSearchStyles.exec(query)?.[1]
+  if (circularID) {
+    const db = await tables()
+    const result = await db.circulars.get({
+      circularId: Number(circularID),
```
lpsinger, line 185:
```suggestion
  const circularId = parseFloat(validCircularSearchStyles.exec(query)?.[1])
  if (!isNaN(circularID)) {
    const db = await tables()
    const result = await db.circulars.get({
      circularId,
```

app/routes/circulars/circulars.server.ts
tylerbarna: ah, okay. I am having an issue where parseFloat on that line is causing the pre-commit to fail because 

> Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
>   Type 'undefined' is not assignable to type 'string'.

though this issue is persisting after altering index.tsx to only call circularRedirect if the query is not undefined and changing the query argument to only be accepted as a string. Looking online, this might be an issue with my intellisense on VSCode

app/routes/circulars/circulars.server.ts
tylerbarna: yeah, I'm still getting the error after restarting the typesense server, are you encountering something similar?

lpsinger: I see. Although JavaScript permits this, apparently TypeScript makes a principled (and perhaps controversial) decision not to support it. See, for instance, https://github.com/microsoft/TypeScript/issues/4002 and https://github.com/microsoft/TypeScript/issues/15198.

I think that it works if you replace `parseFloat` with `Number`.

tylerbarna: I was able to avoid the error by changing it to 
``const circularId = parseFloat(validCircularSearchStyles.exec(query)?.[1] || '')``

app/routes/circulars/circulars.server.ts
```diff
@@ -174,18 +174,17 @@ export async function put(subject: string, body: string, request: Request) {
...
-  if (circularID) {
+  const circularId = parseFloat(
+    validCircularSearchStyles.exec(query)?.[1] || ''
+  )
+  if (!isNaN(circularId)) {
     const db = await tables()
-    const result = await db.circulars.get({
-      circularId: Number(circularID),
-    })
+    const result = await db.circulars.get({ circularId: circularId })
```
lpsinger, line 185:
```suggestion
    const result = await db.circulars.get({ circularId })
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
```diff
@@ -206,6 +211,16 @@ export default function () {
...
+      <p className="usa-hint">
+        <Hint>
+          Search for Circulars by submitter, subject, or body text. For example,
+          to search for GRB circulars from Fermi, search for 'Fermi GRB'.
+        </Hint>
+        <Hint>
+          To navigate to a specific circular, search for the associated Circular
+          ID ('gcn123', 'Circular 123', '123', etc.)
+        </Hint>
+      </p>
```
lpsinger, line 223: - React doesn't like `<p>` tags to contain `<div>`s. (The `<Hint>` component has a `<div>` inside it, so a `<p>` cannot contain a `<Hint>`.) See screen shot below showing the client-side error from React.
- Add an `id` attribute so that we can link the input field to the hint using Aria attributes.

Screenshot 2023-05-25 at 14 43 47

```suggestion
        <Hint id="searchHint">
          Search for Circulars by submitter, subject, or body text. For example,
          to search for GRB circulars from Fermi, search for 'Fermi GRB'.<br/>
          To navigate to a specific circular, search for the associated Circular
          ID ('gcn123', 'Circular 123', '123', etc.)
        </Hint>
```

app/routes/circulars/index.tsx
```diff
@@ -184,6 +188,7 @@ export default function () {
             name="query"
             type="search"
             defaultValue={query}
+            placeholder="Search"
```
lpsinger, line 191:
```suggestion
            placeholder="Search"
            aria-describedby="searchHint"
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
tylerbarna: ah, my apologies, I missed that

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,18 @@ export async function put(subject: string, body: string, request: Request) {
     submitter: formatAuthor(user),
   })
 }
+
+export async function circularRedirect(query: string) {
+  const validCircularSearchStyles =
+    /^(?:\s*GCN\s*)?(?:\s*CIRCULAR\s*)?((\s*-?\d+(?:\.\d)?)\s*)$/i
```
lpsinger, line 179: - Remove unnecessarily repeated `\s*`
- Remove unnecessary doubly-nested capturing group

```suggestion
    /^\s*(?:GCN)?\s*(?:CIRCULAR)?\s*(-?\d+(?:\.\d)?)\s*$/i
```

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
```diff
@@ -172,3 +173,18 @@ export async function put(subject: string, body: string, request: Request) {
...
+    /^\s*(?:GCN)?\s*(?:CIRCULAR)?\s*(-?\d+(?:\.\d)?)\s*$/i
+  const circularId = parseFloat(
+    validCircularSearchStyles.exec(query)?.[1] || ''
+  )
+  if (!isNaN(circularId)) {
+    const db = await tables()
+    const result = await db.circulars.get({ circularId })
+    if (!result) return
+    const circularURL = `/circulars/${circularId}`
+    throw redirect(circularURL, { status: 302 })
```
lpsinger, line 188: According to https://remix.run/docs/en/main/utils/redirect, 302 is the default redirect status code, so we can omit it.

```suggestion
    throw redirect(circularURL)
```

app/routes/circulars/index.tsx
```diff
@@ -35,6 +36,9 @@ const limit = 100
 export async function loader({ request: { url } }: DataFunctionArgs) {
   const { searchParams } = new URL(url)
   const query = searchParams.get('query') || undefined
+  if (query) {
+    await circularRedirect(query.toString())
```
lpsinger, line 40: It's already guaranteed to be a string by this point.

```suggestion
    await circularRedirect(query)
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
tylerbarna: oh yeah that was a holdover from when I was trying to figure out the issue with parseFloat

lpsinger (changes requested):

lpsinger (approved):

PR #980, opened 2023-05-23
Use parseFloat to process circular IDs for DDB lookup

lpsinger:
This fixes a bug that would have made it impossible to view circulars with half-integer IDs.


dakota002:
> This fixes a bug that would have made it impossible to view circulars with half-integer IDs.

Do we have those?

lpsinger:
> > This fixes a bug that would have made it impossible to view circulars with half-integer IDs.
> 
> Do we have those?

Sadly, yes. 😮‍💨 

Screenshot 2023-05-23 at 18 04 46

dakota002 (approved):

PR #1010, opened 2023-06-06
Schema versioning url

dakota002:


dakota002:
@lpsinger When you get a chance can we go over how the caching stuff works with our CDN? The array of tags will definitely be useful to have cached, especially if I am going to implement some UI for browsing various versions

lpsinger:
When you are ready for a review, please remember to remove the draft status.

lpsinger (changes requested):
I'm not crazy about using the query string here. How about just redirecting `https://gcn.nasa.gov/${path}` to `https://raw.githubusercontent.com/nasa-gcn/gcn-schema/${path}`? That should handle all branches and tags implicitly.

dakota002:
> I'm not crazy about using the query string here. How about just redirecting `https://gcn.nasa.gov/${path}` to `https://raw.githubusercontent.com/nasa-gcn/gcn-schema/${path}`? That should handle all branches and tags implicitly.

Yea, that should work. I'll try that out

lpsinger (changes requested):
I'm not sure that we should automatically add "main" to the path. We would like to do redirects for arbitrary branches which may have any spelling.

If anything, we should have a "stable" route that redirects to the most recent tag.

dakota002:
> I'm not sure that we should automatically add "main" to the path. We would like to do redirects for arbitrary branches which may have any spelling.
> 
> If anything, we should have a "stable" route that redirects to the most recent tag.

Okay, that makes sense. Are there any other specific keywords that would be worth remapping or just stable?

lpsinger:
> > I'm not sure that we should automatically add "main" to the path. We would like to do redirects for arbitrary branches which may have any spelling.
> > If anything, we should have a "stable" route that redirects to the most recent tag.
> 
> Okay, that makes sense. Are there any other specific keywords that would be worth remapping or just stable?

Just `stable`, I think.

Note that you should be able to achieve this by creating two routes `routes/schema/$.ts` and `routes/schema/latest.ts`.

app/routes/schema/stable/$.ts
```diff
@@ -0,0 +1,32 @@
...
+export async function loader({ params: { '*': path } }: DataFunctionArgs) {
+  const GITHUB_API_TOKEN = getEnvOrDieInProduction('GITHUB_API_ACCESS')
+  const tags: TagObject[] = await (
+    await fetch('https://api.github.com/repos/nasa-gcn/gcn-schema/tags', {
+      headers: {
+        Accept: 'application/vnd.github+json',
+        Authorization: `Bearer ${GITHUB_API_TOKEN}`,
+      },
+    })
+  ).json()
```
lpsinger, line 25: Are you sure that this request gives you the _most recent_ tag?

There's an API endpoint for the latest release: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release

Also we already have a dependency on octokit, so you might as well use that rather than hitting the API endpoint directly.

app/routes/schema/stable/$.ts
```diff
@@ -0,0 +1,32 @@
...
+import { type DataFunctionArgs, redirect } from '@remix-run/node'
+
+import { getEnvOrDieInProduction } from '~/lib/env.server'
+
+type TagObject = {
+  name: string
+}
+
+export async function loader({ params: { '*': path } }: DataFunctionArgs) {
+  const GITHUB_API_TOKEN = getEnvOrDieInProduction('GITHUB_API_ACCESS')
```
lpsinger, line 17:
```suggestion
  const GITHUB_API_TOKEN = getEnvOrDieInProduction('GITHUB_API_TOKEN')
```

Also, please document this environment variable in `contributing.mdx`.

app/routes/schema/stable/$.ts
```diff
@@ -0,0 +1,32 @@
...
+  const tags: TagObject[] = await (
+    await fetch('https://api.github.com/repos/nasa-gcn/gcn-schema/tags', {
+      headers: {
+        Accept: 'application/vnd.github+json',
+        Authorization: `Bearer ${GITHUB_API_TOKEN}`,
+      },
+    })
+  ).json()
+
+  path = path?.replace('stable', tags[0]?.name ?? 'main')
```
lpsinger, line 27: I think that if there aren't any tags, then this should just return a 404 status (or even just fail outright).

app/routes/schema/stable/$.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  return redirect(
+    `https://raw.githubusercontent.com/nasa-gcn/gcn-schema/main/${path ?? ''}`
```
lpsinger, line 30: `main` here is wrong; it should be the tag name.

lpsinger (changes requested):

app/routes/schema/stable/$.ts
dakota002: I was unaware of the octokit stuff, but that seems like its going to solve a lot of the issues I was having previously. I'll make use of it in the next commit. The tag api returns the values in descending order, so the first one in the list should be the most recent.

Are we going to be making releases off of tags?  If so I can use that api instead, and to be sure we are getting the most recent I can sort by the created date.

lpsinger: Yes, we'll use GitHub releases.

PR #1017, opened 2023-06-09
Deliver emails to S3 bucket rather than SNS

lpsinger:
SNS has a message size limit of 150 kB, whereas S3 has a message size limit of 40 MB.

### DIFF
--- a/app.arc
+++ b/app.arc
@@ -95,9 +95,6 @@ architecture arm64
 memory 256
 timeout 10
 
-@storage-private
-email-incoming
-
 @search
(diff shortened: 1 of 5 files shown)


lpsinger:
We can't merge this yet because we don't have permission to create S3 bucket policies.

dakota002 (approved):

lpsinger:
Oops!

> We can't merge this yet because we don't have permission to create S3 bucket policies.

I'll have to revert this and reopen the MR.

Issue #878, opened 2023-04-25
Limit search by date range


lpsinger:

lpsinger:
> Couple of clarifying questions:
> 
> * [ ]  I imagine we want this to be a modifier to a normal search so one can search for something like fermi circulars in a given date range?

Yes.

> * [ ]  For the date range, the date in question be the submission date, right?

Yes.

> * [ ]  Is there a specific syntax for a date range query that would work best, maybe something similar to ADS? Similarly, should it be able to handle European style dates as well (ie MM-DD-YYYY and DD-MM-YYYY)?

Definitely ISO 8601, possibly human date ranges too (like "10 days ago"). Also we definitely will want some kind of date and time picker widget.

> * [ ]  Do we want "open" queries (ie should there be an option to search for all circulars prior to or after a given date)?

Yes.

> * [ ]  Also, should there be a shorthand for getting circulars in a given day/month/year (so searching something like year:2022 would return all circulars from 2022 and searching month:05-2023 would return all circulars from May 2023)?

I'm not sure.

Note that `year:2022` looks a lot like [Lucene query syntax](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html) which is under the hood and supported by OpenSearch.

lpsinger:
> For (1), I was looking into the [opensearch docs](https://opensearch.org/docs/2.0/opensearch/supported-field-types/range/) regarding the range field, which has a type specific to dates. This seemed like a potential solution, but I was having some trouble incorporating it with the multi_search in the search function, and I'm not sure if it can handle the native createdOn format easily.

We may not have defined the fields as the correct type in the [index mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/explicit-mapping.html). You can experiment with that in your local ElasticSearch instance that starts when you run `npm run dev`.

tylerbarna:
@lpsinger do you think a date picker might be easier to implement? https://www.npmjs.com/package/react-datepicker

I was thinking it might be better to have some type of expandable component that contains "advanced" search inputs. For this issue, there can be a date picker, and issue #816 could also be fixed by adding some type of selection for object type; there could also be one to filter by mission 

I'd just need to figure out how to implement an expandable element

lpsinger:
On second thought, times are not a requirement.

tylerbarna:
> I encourage you to find and reuse an existing picker, but I think we want something that can do dates _and times_.

[...]

PR #1026, opened 2023-06-12
Circulars search backend for date range

tylerbarna:
Backend for Issue #878 

Date range filter has been implemented inside circulars.server.ts with the addition of an advancedQuery optional argument, which contains optional start and end date arguments that will default to negative and positive infinity, respectively. When the frontend is added in a separate PR, react-uswds has a date range picker element that can be used.

Added functions to circulars.lib.ts for converting a unix time to mm-dd-yy format and another function to do the opposite. The URLSearchParams for the start and end dates are in unix time (the circular items store the createdOn time in unix time format). In the search function, the iso8061 format dates are converted to unix times to simplify the filter. However, as Leo said in the 06/12/23 GCN meeting, it would be better if the date range values were stored in the iso8061 format in the url so they're human readable. This would require changes to index.tsx, but the formatDateEpoch implementation in the search function shouldn't require any changes since it should be able to convert to unix time if provided an iso8061 string.

Currently, the changes made in index.tsx to incorporate the date range parameters cause an unintended issue with the way URLSearchParams are retreived by the loader function after initiating a search; in the loader function, it seems that the only parameter retreived is the query parameter. The lack of the page number parameter causes the query results to not display correctly, though logging the results variable does show that they're being loaded, so I don' think it's due to the search function. Inspecting the url parameters in other functions, which print in the client-side console, seem to have all of the expected parameters. This is explained in slightly more detail in [this comment](https://github.com/nasa-gcn/gcn.nasa.gov/issues/878#issuecomment-1583244451) in the related issue.

### DIFF
--- a/app/routes/circulars/circulars.lib.ts
+++ b/app/routes/circulars/circulars.lib.ts
@@ -76,6 +76,11 @@ export function formatDateISO(date: number) {
   return new Date(date).toISOString().replace(/\.\d+Z$/, 'Z')
 }
 
+/** convert a date in format mm-dd-yyyy (or YYYY-MM_DD) to ms since 01/01/1970 */
+export function formatDateEpoch(date: string) {
+  return new Date(date).getTime()
+}
+
 /** Return true if the subject is valid, false if it is invalid, or undefined if it is an empty string */
 export function subjectIsValid(subject: string) {
   if (subject.length) {
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -15,6 +15,7 @@ import memoizee from 'memoizee'
 import { getUser } from '../__auth/user.server'
 import { bodyIsValid, formatAuthor, subjectIsValid } from './circulars.lib'
 import type { Circular, CircularMetadata } from './circulars.lib'
+import { formatDateEpoch } from './circulars.lib'
 import { search as getSearch } from '~/lib/search.server'
 
 export const group = 'gcn.nasa.gov/circular-submitter'
@@ -48,16 +49,23 @@ export async function search({
   query,
   page,
   limit,
+  startDate,
+  endDate,
 }: {
   query?: string
   page?: number
   limit?: number
+  startDate?: string
+  endDate?: string
 }): Promise<{
   items: CircularMetadata[]
   totalPages: number
   totalItems: number
 }> {
   const client = await getSearch()
+
+  const startTime = formatDateEpoch(startDate!) || -Infinity
+  const endTime = formatDateEpoch(endDate!) + 86400000 || Infinity
 
   const {
     body: {
@@ -69,11 +77,29 @@ export async function search({
   } = await client.search({
     index: 'circulars',
     body: {
-      query: query && {
-        multi_match: { query, fields: ['submitter', 'subject', 'body'] },
+      query: {
+        bool: {
+          must: query
+            ? {
+                multi_match: {
+                  query,
+                  fields: ['submitter', 'subject', 'body'],
+                },
+              }
+            : undefined,
+          filter: {
+            range: {
+              createdOn: {
+                gte: startTime,
+                lte: endTime,
+              },
+            },
+          },
+        },
       },
-      fields: ['subject'],
-      _source: false,
+      _source: {
+        includes: ['subject', 'createdOn'],
+      },
       sort: {
         circularId: {
           order: 'desc',
@@ -88,15 +114,17 @@ export async function search({
   const items = hits.map(
     ({
       _id: circularId,
-      fields: {
-        subject: [subject],
-      },
+      _source: { subject, createdOn },
     }: {
       _id: string
-      fields: { subject: string[] }
+      _source: {
+        subject?: string
+        createdOn?: string
+      }
     }) => ({
       circularId,
       subject,
+      createdOn,
     })
   )
 
--- a/app/routes/circulars/index.tsx
+++ b/app/routes/circulars/index.tsx
@@ -39,8 +39,17 @@ export async function loader({ request: { url } }: DataFunctionArgs) {
   if (query) {
     await circularRedirect(query)
   }
+  const startDate = searchParams.get('startDate') || undefined
+  const endDate = searchParams.get('endDate') || undefined
   const page = parseInt(searchParams.get('page') || '1')
-  const results = await search({ query, page: page - 1, limit })
+  const results = await search({
+    query,
+    page: page - 1,
+    limit,
+    startDate,
+    endDate,
+  })
+
   return { page, ...results }
 }
 
@@ -53,9 +62,17 @@ export async function action({ request }: DataFunctionArgs) {
   return await put(subject, body, request)
 }
 
-function getPageLink(page: number, query?: string) {
-  const searchParams = new URLSearchParams({ page: page.toString() })
+function getPageLink(
+  page: number,
+  query?: string,
+  startDate?: string,
+  endDate?: string
+) {
+  const searchParams = new URLSearchParams({ page: page.toString() || '1' })
   if (query) searchParams.set('query', query)
+  searchParams.set('page', page.toString() || '1')
+  if (startDate) searchParams.set('startDate', startDate)
+  if (endDate) searchParams.set('endDate', endDate)
   return `?${searchParams.toString()}`
 }
 
@@ -63,10 +80,14 @@ function Pagination({
   page,
   totalPages,
   query,
+  startDate,
+  endDate,
 }: {
   page: number
   totalPages: number
   query?: string
+  startDate?: string
+  endDate?: string
 }) {
   const pages = usePagination({ currentPage: page, totalPages })
 
@@ -82,7 +103,7 @@ function Pagination({
                   key={i}
                 >
                   <Link
-                    to={getPageLink(number!, query)}
+                    to={getPageLink(number!, query, startDate, endDate)}
                     className="usa-pagination__link usa-pagination__previous-page"
                     aria-label="Previous page"
                   >
@@ -108,7 +129,7 @@ function Pagination({
                   key={i}
                 >
                   <Link
-                    to={getPageLink(number!, query)}
+                    to={getPageLink(number!, query, startDate, endDate)}
                     className="usa-pagination__link usa-pagination__next-page"
                     aria-label="Next page"
                   >
@@ -124,7 +145,7 @@ function Pagination({
                   key={i}
                 >
                   <Link
-                    to={getPageLink(number!, query)}
+                    to={getPageLink(number!, query, startDate, endDate)}
                     className={classNames('usa-pagination__button', {
                       'usa-current': isCurrent,
                     })}
@@ -152,11 +173,12 @@ export default function () {
 
   const [searchParams] = useSearchParams()
   const query = searchParams.get('query') ?? undefined
+
   let searchParamsString = searchParams.toString()
   if (searchParamsString) searchParamsString = `?${searchParamsString}`
 
-  const [input, setInput] = useState(query)
-  const clean = input === query
+  const [inputQuery, setInputQuery] = useState(query)
+  const clean = inputQuery === query
 
   const submit = useSubmit()
 
@@ -187,11 +209,11 @@ export default function () {
             id="query"
             name="query"
             type="search"
-            defaultValue={query}
+            defaultValue={inputQuery}
             placeholder="Search"
             aria-describedby="searchHint"
             onChange={({ target: { form, value } }) => {
-              setInput(value)
+              setInputQuery(value)
               if (!value) submit(form)
             }}
           />


tylerbarna:
> Backend for Issue #878
> 
> Date range filter has been implemented inside circulars.server.ts with the addition of an advancedQuery optional argument, which contains optional start and end date arguments that will default to negative and positive infinity, respectively. When the frontend is added in a separate PR, react-uswds has a date range picker element that can be used.
> 

My thoughts on the frontend was to add an advanced query button (something like [this](https://github.com/tylerbarna/gcn.nasa.gov/blob/68391db7acf37ace4e5e4479379e90e54673da4a/app/routes/circulars/index.tsx#LL206C3-L206C3)) to the search bar that expands an accordion menu ([react-uswds documentation](https://trussworks.github.io/react-uswds/?path=/docs/components-accordion--borderless)) with different advanced query options, the current addition being a date range picker ([react-uswds documentation](https://trussworks.github.io/react-uswds/?path=/docs/components-date-range-picker--complete-date-range-picker))

app/routes/circulars/circulars.lib.ts
```diff
@@ -76,6 +76,25 @@ export function formatDateISO(date: number) {
...
+  return dateStr.replace(/\//g, '-') // replace / with -
+}
+
+/** convert a date in format mm-dd-yyyy to ms since 01/01/1970 */
+export function formatDateEpoch(date: string | number) {
+  if (typeof date === 'string') {
+    return new Date(date).getTime()
+  }
+  return date
+}
```
lpsinger, line 96: Use the format `YYYY-MM-DD`, not `MM-DD-YYYY`, to more closely match the international ISO 8601 format.

app/routes/circulars/circulars.lib.ts
```diff
@@ -76,6 +76,25 @@ export function formatDateISO(date: number) {
...
+  })
+  return dateStr.replace(/\//g, '-') // replace / with -
+}
+
+/** convert a date in format mm-dd-yyyy to ms since 01/01/1970 */
+export function formatDateEpoch(date: string | number) {
+  if (typeof date === 'string') {
+    return new Date(date).getTime()
+  }
+  return date
```
lpsinger, line 95: I think it's a bit too fancy for it to take either a UNIX timestamp or a string. Let's just have this take a string, so that this function is simply the inverse of `formatDateMMDDYYYY `.

```suggestion
export function formatDateEpoch(date: string) {
  return new Date(date).getTime()
```

app/routes/circulars/circulars.server.ts
```diff
@@ -48,17 +49,29 @@ export async function search({
...
   limit,
+  advancedQuery,
 }: {
   query?: string
   page?: number
   limit?: number
+  advancedQuery?: {
+    startDate?: string | number
+    endDate?: string | number
+  }
```
lpsinger, line 60: Move these out of the `advancedQuery` object:

```suggestion
  startDate?: string | number
  endDate?: string | number
```

app/routes/circulars/circulars.server.ts
```diff
@@ -69,11 +82,27 @@ export async function search({
...
+                gte: startTime,
+                lte: endTime,
+              },
+            },
+          },
+        },
       },
-      fields: ['subject'],
-      _source: false,
+      _source: ['subject', 'createdOn'],
```
lpsinger, line 105: This was probably a typo. I think that the `_source` argument is supposed to be a boolean.

app/routes/circulars/circulars.server.ts
```diff
@@ -107,13 +138,12 @@ export async function search({
...
 /** Get a circular by ID. */
 export async function get(circularId: number): Promise<Circular> {
-  if (isNaN(circularId)) throw new Response(null, { status: 404 })
   const db = await tables()
   const result = await db.circulars.get({
     circularId,
   })
   if (!result)
-    throw new Response(null, {
+    throw new Response('The requested circular does not exist', {
```
lpsinger, line 146: Whoops, looks like this patch undid some unrelated changes. Possibly a merge artifact?

lpsinger (changes requested):
I'm not sure of the cause of the bug that you are seeing, but I would suggest not storing complicated composite objects with `useState` because that is an easy source of bugs.

app/routes/circulars/circulars.lib.ts
tylerbarna: the rationale I had behind this was so the search function was more flexible when defining the startTime and endTime constants so it could intake either the unix format as a number or the human-readable/iso8601 format as a string (see [here](https://github.com/tylerbarna/gcn.nasa.gov/blob/6d1838715fdd7b4da9b28ef040fb4e1ddfc856de/app/routes/circulars/circulars.server.ts#L70)), though I see why it's better to only keep it as one type. Do you think it would be better to keep it as the string up to the point of defining it in the search function?

app/routes/circulars/circulars.server.ts
tylerbarna: Looking through some docs ([1](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html)) ([2](https://opster.com/guides/elasticsearch/glossary/elasticsearch-source/)), I think I implemented the syntax slightly incorrectly, but setting it to false seems to break the circulars indexing on initial load. I probably need to look this over a bit more to understand what's happening, but altering it to the following seems to work (doesn't fix the query submission issue that I'm encountering though):

```
      _source: {
        includes: ['subject', 'createdOn'],
      }
```

app/routes/circulars/circulars.server.ts
tylerbarna: yeah, not sure how this happened, I think my fork may have been out of date when I created this branch or something along those lines. Resolved in [adf6556](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1026/commits/adf6556f2ea022c03eea7237e9efd56f2d737254)

app/routes/circulars/circulars.server.ts
tylerbarna: Made this change here and in relevant functions in [67335f3](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1026/commits/67335f39365fe353b6c9a2d5b513efe572d1a73e). I think I fixed search being passed the default values (Infinity and -Infinity), but need to check if providing Infinity as a string to formatDateEpoch will return Infinity as a number

app/routes/circulars/circulars.lib.ts
tylerbarna: I added a function for YYYY-MM-DD in 0e2c267. Given the goal to have the input and url string be human readable in  iso8601, this function might end up being unneeded, as the start and end dates should just be passed as strings at all points and only converted to unix time inside the search function. That being said, looking around a bit more, the default infinity values don't seem to interact well when using the Date objects. One solution to that could be to change it so the default start and end dates are undefined and at the point of conversion inside the search, they are set to infinity and negative infinity if they're undefined (similar to how query is defined when there is no query), though this would require tweaking the declaration of their types in the various functions.

lpsinger: Right. We don't need the function to convert from epoch to ISO. We only need the function to convert from ISO to epoch.

I agree: use undefined, not infinity.

app/routes/circulars/circulars.lib.ts
```diff
@@ -76,6 +76,28 @@ export function formatDateISO(date: number) {
...
+    day: '2-digit',
+  })
+  return dateStr.replace(/\//g, '-') // replace / with -
+}
+/** convert a ms value to YYYY-MM-DD format */
+export function formatDateYYYYMMDD(date: number) {
+  const dateObj = new Date(date)
+  var dateStr = dateObj.toISOString().split('T')[0]
+  return dateStr
+}
```
lpsinger, line 94: Remove, unused:

```suggestion
```

app/routes/circulars/circulars.lib.ts
```diff
@@ -76,6 +76,28 @@ export function formatDateISO(date: number) {
...
+
+/** convert a date in format mm-dd-yyyy (or YYYY-MM_DD) to ms since 01/01/1970 */
+export function formatDateEpoch(date: string) {
+  return new Date(date).getTime()
+}
```
lpsinger, line 99: Take care: when you search for records with dates between `yyyy-mm-dd` (lower limit) and `YYYY-MM-DD` (upper limit) that you are searching between the times `yyyy-mm-ddT00:00:00Z` and `YYYY-MM-DDT23:59:59Z` (so that hits include the whole day that comprises the upper end of the interval).

app/routes/circulars/index.tsx
```diff
@@ -151,12 +184,20 @@ export default function () {
...
+
   let searchParamsString = searchParams.toString()
   if (searchParamsString) searchParamsString = `?${searchParamsString}`
+  const combinedQuery = { query, page: page, startDate, endDate }
+  console.log('default function combinedQuery', combinedQuery)
 
-  const [input, setInput] = useState(query)
-  const clean = input === query
+  const [input, setInput] = useState(combinedQuery)
+  const clean = input === combinedQuery
```
lpsinger, line 198: My advice: don't create states that are composite objects. Have a separate state variable for each input field, like this:

```ts
const [inputQuery, setInputQuery] = useState(query)
const [inputDateGte, setInputDateGte] = useState(gte)
const [inputDateLte, setInputDateLte] = useState(lte)
const clean = inputQuery === query && inputDateGte === gte && inputDateLte === lte
...
```

The reason is that there are [many subtleties around properly updating React states that are objects](https://react.dev/learn/updating-objects-in-state).

lpsinger (changes requested):

lpsinger:
@dakota002, while I am out next week, would you please mentor @tylerbarna and help him to work out the remaining kinks in this PR?

app/routes/circulars/index.tsx
tylerbarna: this actually fixed the issue with the loading not working I think

app/routes/circulars/circulars.lib.ts
tylerbarna: removed in fbce467

app/routes/circulars/circulars.lib.ts
tylerbarna: > Right. We don't need the function to convert from epoch to ISO. We only need the function to convert from ISO to epoch.
> 
> I agree: use undefined, not infinity.

made the change so they're undefined in 64a34b2.

app/routes/circulars/circulars.server.ts
tylerbarna: with most recent commit (64a34b2), double checked that the formatDateEpoch function works inside of the search function to limit the range of circulars by date. Manually changed the definition of startTime and endTime constants so instead of referencing startDate and endDate, they were hard-coded strings in the YYYY-MM-DD format. With this, the search function successfully limited the circulars to the provided date range (prior to the commit, I reverted the hard coded date range back to the constants). The backend for the search function seems to be functional as of now.

tylerbarna:
with most recent commit (https://github.com/nasa-gcn/gcn.nasa.gov/commit/64a34b23330bd96b2584383d677af3df3f8e81d2), double checked that the formatDateEpoch function works inside of the search function to limit the range of circulars by date. Manually changed the definition of startTime and endTime constants so instead of referencing startDate and endDate, they were hard-coded strings in the YYYY-MM-DD format. With this, the search function successfully limited the circulars to the provided date range (prior to the commit, I reverted the hard coded date range back to the constants). The backend for the search function seems to be functional as of now.

tylerbarna:
note that, while the backend does not have a frontend, the [definition of the start and end date input constants](https://github.com/tylerbarna/gcn.nasa.gov/blob/8999d1529269a10eb3d313d82168d322f188690b/app/routes/circulars/index.tsx#L197) are commented out so commits can pass the husky pre-commit hook (when uncommented, they are defined but not referenced). With the frontend, they will be referenced inside of the date range picker elements similar to the way [inputQuery](https://github.com/tylerbarna/gcn.nasa.gov/blob/8999d1529269a10eb3d313d82168d322f188690b/app/routes/circulars/index.tsx#L196) is referenced [in the search box element](https://github.com/tylerbarna/gcn.nasa.gov/blob/8999d1529269a10eb3d313d82168d322f188690b/app/routes/circulars/index.tsx#L230)

tylerbarna:
I noticed while looking over the work for this PR, but the changes made in #967 are such that there won't need to be any additional changes for startDate, endDate, and any future URLSearchParameters to be preserved when using the back button after navigating to a specific circular.

app/routes/circulars/circulars.lib.ts
tylerbarna: I made a small change in 55e52d9 so the endDate, when defined, should include circulars recorded on the date of the endDate. I did this by just adding the number of milliseconds in a day to the endTime constant, which is in unix format,  when it's defined inside of the search function. Wasn't sure if that's sufficient or if it would it be better to look for a more elegant implementation.

tylerbarna:
@dakota002 After applying some of Leo's requested changes, the date range backend seems to be working without causing any issues with the standard query search. Still need to remove any console.log calls in index.tsx that I was using for debugging.

tylerbarna:
> > @dakota002 After applying some of Leo's requested changes, the date range backend seems to be working without causing any issues with the standard query search. Still need to remove any console.log calls in index.tsx that I was using for debugging.
> 
> Awesome, let me know when you want me to take a look at the PR

yeah, I think it's probably ready for review

app/routes/circulars/index.tsx
```diff
@@ -151,12 +183,18 @@ export default function () {
...
+
   let searchParamsString = searchParams.toString()
   if (searchParamsString) searchParamsString = `?${searchParamsString}`
 
-  const [input, setInput] = useState(query)
-  const clean = input === query
+  const [inputQuery, setInputQuery] = useState(query)
+  // const [inputDateGte, setInputDateGte] = useState(startDate)
+  // const [inputDateLte, setInputDateLte] = useState(endDate)
+  const clean = inputQuery === query //&& inputDateGte === startDate && inputDateLte === endDate
```
dakota002, line 197: These commented out lines can be deleted

app/routes/circulars/index.tsx
```diff
@@ -34,13 +34,25 @@ import searchImg from 'app/theme/img/usa-icons-bg/search--white.svg'
 const limit = 100
 
 export async function loader({ request: { url } }: DataFunctionArgs) {
+  console.log('loader url', url)
```
dakota002, line 37:
```suggestion
```

app/routes/circulars/index.tsx
```diff
@@ -34,13 +34,25 @@ import searchImg from 'app/theme/img/usa-icons-bg/search--white.svg'
...
   const { searchParams } = new URL(url)
+  console.log('loader search params', searchParams)
```
dakota002, line 39:
```suggestion
```

app/routes/circulars/index.tsx
```diff
@@ -34,13 +34,25 @@ import searchImg from 'app/theme/img/usa-icons-bg/search--white.svg'
...
   const query = searchParams.get('query') || undefined
   if (query) {
     await circularRedirect(query)
   }
+  const startDate = searchParams.get('startDate') || undefined
+  const endDate = searchParams.get('endDate') || undefined
   const page = parseInt(searchParams.get('page') || '1')
-  const results = await search({ query, page: page - 1, limit })
+  console.log('loader page', page)
```
dakota002, line 47:
```suggestion
```

app/routes/circulars/index.tsx
```diff
@@ -53,22 +65,42 @@ export async function action({ request }: DataFunctionArgs) {
...
+  query?: string,
+  startDate?: string,
+  endDate?: string
+) {
+  const searchParams = new URLSearchParams({ page: page.toString() || '1' })
   if (query) searchParams.set('query', query)
+  searchParams.set('page', page.toString() || '1')
+  if (startDate) searchParams.set('startDate', startDate)
+  if (endDate) searchParams.set('endDate', endDate)
+  console.log('getPageLink search params', searchParams.toString())
```
dakota002, line 79:
```suggestion
```

app/routes/circulars/index.tsx
```diff
@@ -53,22 +65,42 @@ export async function action({ request }: DataFunctionArgs) {
...
+  endDate?: string
 }) {
   const pages = usePagination({ currentPage: page, totalPages })
+  console.log(
+    'Pagination search params',
+    query?.toString(),
+    startDate?.toString(),
+    endDate?.toString()
+  )
+  console.log('Pagination page', page)
```
dakota002, line 103:
```suggestion
```

dakota002 (changes requested):
Looks good, but these are just some cleanup requests.  Please remove commented code and console logs

app/routes/circulars/index.tsx
tylerbarna: these are for the frontend, but yeah, probably best to leave them out until the actual frontend PR

tylerbarna:
> Looks good, but these are just some cleanup requests. Please remove commented code and console logs

I've implemented all the requested changes

dakota002 (approved):
Looks good to me!

tylerbarna:
@dakota002 I believe I squashed the commits down

tylerbarna:
> @dakota002 I believe I squashed the commits down

in the future, is it better to squash the commits in the same PR/branch, or should I transfer everything to a new branch/PR with a single commit?

dakota002:
> > @dakota002 I believe I squashed the commits down
> 
> in the future, is it better to squash the commits in the same PR/branch, or should I transfer everything to a new branch/PR with a single commit?

You can do it all on the same PR/branch. I've had a few cases where I've needed to do it multiple times over on the same branch to avoid large conflicts

PR #1052, opened 2023-06-22
Browser v2

dakota002:
Removing local dependency on schema repo. Still need to add a means of selecting versions

### DIFF
--- a/app/lib/schema-data.ts
+++ b/app/lib/schema-data.ts
@@ -5,42 +5,50 @@
  *
  * SPDX-License-Identifier: NASA-1.3
  */
-import { readFile, readdir } from 'fs/promises'
-import { dirname, extname, join } from 'path'
+import { Octokit } from '@octokit/rest'
+import { extname, join } from 'path'
 
+import { getEnvOrDieInProduction } from './env.server'
 import type {
   ReferencedSchema,
   Schema,
 } from '~/routes/docs/schema-browser/SchemaBrowserElements.lib'
 
+const GITHUB_API_TOKEN = getEnvOrDieInProduction('GITHUB_API_TOKEN')
+const octokit = new Octokit({ auth: GITHUB_API_TOKEN })
+const repoData = {
+  repo: 'gcn-schema',
+  owner: 'nasa-gcn',
+}
+
 function isErrnoException(e: unknown): e is NodeJS.ErrnoException {
   return e instanceof Error && 'code' in e && 'errno' in e
 }
 
-export async function loadJson(filePath: string): Promise<Schema> {
+export async function getVersionRefs() {
+  const releases = (await octokit.rest.repos.listReleases(repoData)).data.map(
+    (x) => ({ name: x.name, ref: x.tag_name })
+  )
+  return [...releases, { name: 'main', ref: 'main' }]
+}
+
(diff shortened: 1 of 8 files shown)


dakota002:
This will require a github api token for the site before merging

app/routes/docs/contributing/index.mdx
```diff
@@ -97,6 +97,11 @@ Here are instructions for getting the GCN site set up and running on your own co
     # RECAPTCHA_SITE_KEY=fill-me-in
     # RECAPTCHA_SECRET_KEY=fill-me-in
 
+    # (Optional) Github Integrations
+    # For parts of the site that pull data using Github's content api (i.e. schema-browser)
+    # See https://github.com/settings/tokens to generate an api token
```
lpsinger, line 102:
```suggestion
    # For parts of the site that pull data using Github's content API (i.e. schema-browser)
    # See https://github.com/settings/tokens to generate an API token
```

app/lib/schema-data.ts
```diff
@@ -52,19 +60,34 @@ export async function loadJson(filePath: string): Promise<Schema> {
...
+  const ghData = (
+    await octokit.repos.getContent({
+      ...repoData,
+      path: path.replaceAll('\\', '/'),
+      ref: ref ?? 'main',
+    })
+  ).data as GitContentDataResponse
+  if (!ghData.content) throw new Response(null, { status: 404 })
+
+  return JSON.parse(Buffer.from(ghData.content, 'base64').toString()) as Schema
```
lpsinger, line 73: I'm surprised that it returns the content base64-encoded. Can you avoid the decoding step by setting the [media type](https://github.com/octokit/octokit.js#media-type-formats)?

app/routes/docs/schema-browser/$.tsx
```diff
@@ -22,17 +24,24 @@ import {
 } from '~/routes/docs/schema-browser/SchemaBrowserElements.lib'
 
 export async function loader({ params: { '*': path } }: DataFunctionArgs) {
-  if (!path) throw new Response(null, { status: 404 })
+  if (!path || !path.includes('.schema.json'))
```
lpsinger, line 27:
```suggestion
  if (!path?.includes('.schema.json'))
```

app/lib/schema-data.ts
```diff
@@ -73,26 +96,59 @@ export type ExampleFiles = {
...
+      if (item.type == 'dir' || item.type == 'submodule') {
+        item.children = await getContentTree(item.path, ref)
+      }
+    }
+
+    return data
+  } catch (error) {
+    return []
+  }
+}
```
lpsinger, line 154: I'm troubled by the performance implications of recursively walking the entire directory tree. It occurs to me that if the schema browser side nav supports expanding _only a single directory at a time_, then we only need to return the listing of a single directory in any one loader invocation.

lpsinger (changes requested):

lpsinger (changes requested):
Please address the outstanding comments from the last review. Sorry, this is a big PR, it's going to take me a few iterations to spot everything.

dakota002:
Still working on the changes from the feedback. The reworking of the getContentTree to not be recursive is just taking a bit longer than I'd like, but should have it done soon

dakota002:
Closed by #1130

PR #1064, opened 2023-06-26
Use `fields`, not `_source`, to control data returned by OpenSearch

lpsinger:
The Elasticsearch documentation hints that this should be more efficient. See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#field-retrieval-methods.

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -101,9 +101,8 @@ export async function search({
           },
         },
       },
-      _source: {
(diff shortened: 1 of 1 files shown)


tylerbarna (approved):
changes look good and I'm not encountering any errors when running; I recall we discussed a change along these lines in the original date range backend PR, but I wasn't able to get it working for that.

lpsinger:
It does seem to be working.

PR #1065, opened 2023-06-26
Propagate date range limits to query strings in pagination links

lpsinger:


tylerbarna (approved):
corrects intended behavior for pagination links

Issue #88, opened 2022-06-03
Change password form


lpsinger:
Add a change password form under user preferences. It should only be displayed for users who have signed on with username and password, and not for users that have signed in with federated identities from LaunchPad, Facebook, or Google.

It should exist as a new page under the user menu and side nav.

We can use the Cognito [ChangePassword](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ChangePassword.html) API call.

Courey:
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1070

PR #1070, opened 2023-06-29
adding password reset

Courey:
addresses [this issue](https://github.com/nasa-gcn/gcn.nasa.gov/issues/88)

This is the implementation for the password update feature.


app/routes/user/password.server.ts
```diff
@@ -0,0 +1,65 @@
...
+    newPassword: string
+  ) {
+    const user = await getUser(request)
+    if (!user) throw new Response('not signed in', { status: 403 })
+    const client = await getOpenIDClient()
+    const session = await storage.getSession(request.headers.get('Cookie'))
+    const refreshToken = session.get('refreshToken')
+    const tokenSet = await client.refresh(refreshToken)
+    const parsedTokenSet = parseTokenSet(tokenSet)
+
```
Courey, line 41: We are currently not doing any validation for password length or strength etc. what are the acceptable parameters for the password that you would like to use if any?

app/components/PasswordResetForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+  if (newPassword != confirmPassword) {
+    throw new Response('passwords must match', { status: 400 })
+  } else {
+    const passwordManager = await PasswordManager.create(
+      request,
+      oldPassword,
+      newPassword
+    )
+    await passwordManager.updatePassword()
+    return redirect('/user')
```
lpsinger, line 38: Rather than redirecting the user, what if we display  "Saved" checkmark next to the submit button like we do on some other forms?

app/components/PasswordResetForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+          setIsNewPasswordTouched(true)
+        }}
+        value={confirmPassword}
+      />
+
+      <div className="margin-1">
+        {passwordsMatch ? (
+          <br />
+        ) : (
+          <span className="text-error"> Passwords must match </span>
```
lpsinger, line 92:
```suggestion
          <span className="text-error">Passwords must match</span>
```

lpsinger (changes requested):
Shall we use react-uswds' styling to indicate [invalid fields](https://trussworks.github.io/react-uswds/?path=/story/components-text-input--error)?

app/routes/user.tsx
```diff
@@ -25,6 +38,11 @@ export default function () {
               <NavLink key="endorsements" to="endorsements">
                 Peer Endorsements
               </NavLink>,
+              !idp ? (
```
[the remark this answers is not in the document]
lpsinger: Also, observe that the empty list item causes a doubled separator:

Screenshot 2023-06-29 at 14 59 47

To fix that, try:

```tsX
</NavLink>,
...(idp ? [<Component/>] : []),
<NavLink>
```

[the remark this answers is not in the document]
Courey: got this

app/components/PasswordResetForm.tsx
```diff
@@ -0,0 +1,100 @@
+/*!
```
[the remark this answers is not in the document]
Courey: got this

app/components/PasswordResetForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+      <TextInput
+        data-focus
+        name="oldPassword"
+        id="oldPassword"
+        type="password"
+        placeholder="Old Password"
+      />
+      <Label htmlFor="newPassword">New Password</Label>
+      <TextInput
+        data-focus
```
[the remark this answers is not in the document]
Courey: got this

app/components/PasswordResetForm.tsx
Courey: got this.

app/components/PasswordResetForm.tsx
Courey: I removed this as we added in highlighting. let me know if you want me to put it back in.

app/routes/user.tsx
```diff
@@ -25,6 +38,11 @@ export default function () {
...
+                <NavLink key="password" to="password">
+                  Update Password
```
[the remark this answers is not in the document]
Courey: I made it reset instead.

app/components/Header.tsx
```diff
@@ -132,6 +133,15 @@ export function Header() {
                       >
                         Peer Endorsements
                       </Link>,
+                      !idp ? (
```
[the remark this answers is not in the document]
Courey: got this as long as it is "COGNITO".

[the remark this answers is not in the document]
lpsinger: Exactly.

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,59 @@
...
+    const passwordData = {
+      AccessToken: parsedTokenSet.accessToken,
+      PreviousPassword: oldPassword,
+      ProposedPassword: newPassword,
+    }
+
+    let response
+    try {
+      const command = new ChangePasswordCommand(passwordData)
```
Courey, line 41: This has not been tested!!! This must be tested in a development environment with working cognito idP.

app/components/Header.tsx
```diff
@@ -132,6 +133,15 @@ export function Header() {
                       >
                         Peer Endorsements
                       </Link>,
+                      idp === 'COGNITO' ? (
```
lpsinger, line 136: React generally supports `null`, `false`, and `undefined` as no-ops, so it is safe to use short-circuit evaluation here. Replace this with

```jsx
idp === 'COGNITO' && (
  <Link>
     ...
  </Link>
),
```

app/routes/user.tsx
```diff
@@ -5,14 +5,27 @@
...
+  const user = await getUser(request)
+  if (!user) throw new Response(null, { status: 403 })
+
+  const { email, idp } = user
+  return { email, idp }
+}
+
 export default function () {
+  const { idp } = useLoaderData<typeof loader>()
+
```
lpsinger, line 28: We can avoid an unnecessary data load for this route by retrieving the IdP from the root loader using `useUserIdp()`:

```suggestion
export default function () {
  const idp = useUserIdp()

```

app/routes/user/password.server.ts
lpsinger: That's fine. I'll test it.

app/routes/user/password.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const data = await request.formData()
+  const oldPassword = getFormDataString(data, 'oldPassword')
+  const newPassword = getFormDataString(data, 'newPassword')
+  const confirmPassword = getFormDataString(data, 'confirmPassword')
+
+  if (!oldPassword || !newPassword || !confirmPassword) {
+    throw new Response('all password fields must be present', { status: 400 })
+  }
+
+  if (newPassword != confirmPassword) {
```
lpsinger, line 32:
```suggestion
  if (newPassword !== confirmPassword) {
```

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+import { ChangePasswordCommand } from '@aws-sdk/client-cognito-identity-provider'
+
+import { getOpenIDClient, storage } from '../__auth/auth.server'
+import { getUser, parseTokenSet } from '../__auth/user.server'
+import { maybeThrow } from './cognito.server'
+import { cognito } from '~/lib/cognito.server'
+
+export module PasswordManager {
+  async function updatePasswordFunc(
```
lpsinger, line 16: Get rid of the unnecessary module and inner closure. Let's just have one function `updatePassword`, at _file_ scope.

app/routes/user/password.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const startsOrEndsWithWhitespace = (str: string) => {
+    return /^\s|\s$/.test(str)
+  }
+  const [newPassword, setNewPassword] = useState('')
+  const [confirmPassword, setConfirmPassword] = useState('')
+  const [isNewPasswordTouched, setIsNewPasswordTouched] = useState(false)
+  const [isConfirmPasswordTouched, setIsConfirmPasswordTouched] =
+    useState(false)
+  const passwordsMatch = newPassword === confirmPassword
+  const passwordsAreEmpty = newPassword === '' && confirmPassword === ''
```
lpsinger, line 56: Empty strings are falsey:

```suggestion
  const passwordsAreEmpty = newPassword && confirmPassword
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,140 @@
...
+      <h3 className="site-preview-heading margin-0">
+        New password must contain:
+      </h3>
+      <ul className="usa-list">
+        <li>A lower case letter</li>
+        <li>An upper case letter</li>
+        <li>A number</li>
+        <li>At least 8 characters</li>
+        <li>At least 1 special character or space</li>
+        <li>No leading or trailing spaces</li>
```
lpsinger, line 136: How do we know that this matches the password requirements of our Cognito user pool?

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,59 @@
...
+    oldPassword: string,
+    newPassword: string
+  ) {
+    const user = await getUser(request)
+    if (!user) throw new Response('not signed in', { status: 403 })
+    const client = await getOpenIDClient()
+    const session = await storage.getSession(request.headers.get('Cookie'))
+    const refreshToken = session.get('refreshToken')
+    const tokenSet = await client.refresh(refreshToken)
+    const parsedTokenSet = parseTokenSet(tokenSet)
```
lpsinger, line 27: We have the access token stored in the session. It's not returned by the `getUser` function at the moment. I'll think about where we should expose it and get back to you on that.

lpsinger (changes requested):
@Courey, you might not need AWS access keys to test this... your client ID and client secret may be enough. The Cognito API call that you are using employs the access token for authentication and it might not need any further permissions.

Courey:
> @Courey, you might not need AWS access keys to test this... your client ID and client secret may be enough. The Cognito API call that you are using employs the access token for authentication and it might not need any further permissions.

invalid_scope

I am currently getting invalid scope. I am seeing what I can do without cognito access to get it to work, but that's where I am currently.

app/routes/user/password.tsx
Courey: Screen Shot 2023-07-03 at 2 42 32 PM
I matched what is specified in the login.

lpsinger: Ah, clever.

lpsinger:
> I am currently getting invalid scope. I am seeing what I can do without cognito access to get it to work, but that's where I am currently.

That's actually _really_ useful information.

You see, most Cognito API calls (most AWS API calls, for that matter) use AWS IAM for authorization. We grant ourselves permissions for the corresponding IAM actions here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/src/plugins/lambdaCognitoPermissions.mjs#L26-L35

However, a select few Cognito API calls are authorized with the user's OpenID Connect access token --- but we have to make sure that the client ID and client secret for the web site (or for your local test version) have the appropriate scopes.

You and I should check the [Cognito Developer Guide](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html) to find out what scope is missing.

app/routes/user/password.tsx
Courey: Screen Shot 2023-07-03 at 3 29 14 PM

app/routes/user/password.server.ts
Courey: Sounds like a plan. Let me know what you come up with and I’ll add it in!

Courey:
> > I am currently getting invalid scope. I am seeing what I can do without cognito access to get it to work, but that's where I am currently.
> 
> That's actually _really_ useful information.
> 
> You see, most Cognito API calls (most AWS API calls, for that matter) use AWS IAM for authorization. We grant ourselves permissions for the corresponding IAM actions here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/src/plugins/lambdaCognitoPermissions.mjs#L26-L35
> 
> However, a select few Cognito API calls are authorized with the user's OpenID Connect access token --- but we have to make sure that the client ID and client secret for the web site (or for your local test version) have the appropriate scopes.
> 
> You and I should check the [Cognito Developer Guide](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html) to find out what scope is missing.

Very good to know. I’ll dig through and see what I can find.

app/routes/user/password.tsx
Courey: Screen Shot 2023-07-05 at 8 14 26 AM
it does not work if I do not specify the triple equals empty string.

lpsinger:
I was wrong: if idp is null, that means that the username signed in with username and password. With that change, the API call is working with just the access token.

app/routes/user.tsx
```diff
@@ -25,6 +28,13 @@ export default function () {
               <NavLink key="endorsements" to="endorsements">
                 Peer Endorsements
               </NavLink>,
+              ...(idp === 'COGNITO'
```
lpsinger, line 31:
```suggestion
              ...(!idp
```

app/components/Header.tsx
```diff
@@ -132,6 +133,15 @@ export function Header() {
                       >
                         Peer Endorsements
                       </Link>,
+                      idp === 'COGNITO' && (
```
lpsinger, line 136:
```suggestion
                      idp || (
```

lpsinger (changes requested):

app/components/Header.tsx
lpsinger: We're taking advantage of short circuit evaluation here.

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!user) throw new Response('not signed in', { status: 403 })
+  const client = await getOpenIDClient()
+  const session = await storage.getSession(request.headers.get('Cookie'))
+  const refreshToken = session.get('refreshToken')
+  const tokenSet = await client.refresh(refreshToken)
+  const parsedTokenSet = parseTokenSet(tokenSet)
+
+  if (!oldPassword || !newPassword || !parsedTokenSet.accessToken) {
+    throw new Response('all password fields must be present', { status: 400 })
+  }
```
lpsinger, line 30:
```suggestion
  const session = await storage.getSession(request.headers.get('Cookie'))
  const accessToken = session.get('accessToken')

  if (!oldPassword || !newPassword || !accessToken) {
    throw new Response('all password fields must be present', { status: 400 })
  }
```

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  let response
+  try {
+    const command = new ChangePasswordCommand(passwordData)
+    const response = await cognito.send(command)
+    return response
+  } catch (e) {
+    maybeThrow(e, 'creating fake user password')
+    return response
+  }
```
lpsinger, line 46: - No need to return anything. We don't use the response.
- No need for `maybeThrow` since this should work even in local development.

```suggestion
  await cognito.send(new ChangePasswordCommand({
    AccessToken: parsedTokenSet.accessToken,
    PreviousPassword: oldPassword,
    ProposedPassword: newPassword,
  }))
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,140 @@
...
+      </>
+      <h3 className="site-preview-heading margin-0">
+        New password must contain:
+      </h3>
+      <ul className="usa-list">
+        <li>A lower case letter</li>
+        <li>An upper case letter</li>
+        <li>A number</li>
+        <li>At least 8 characters</li>
+        <li>At least 1 special character or space</li>
```
lpsinger, line 135: What is a "special character" here?

lpsinger (changes requested):
- [ ] The New Password field should go green once it validates.
- [ ] Each item in the password requirement checklist should provide visual feedback on whether it has been met.
- [ ] Provide an appropriate error message if the old password does not match.

app/routes/user/password.tsx
Courey: typically anything that is not A-z or 0-9. This IS a configurable value though, so it depends on what we have configured or if it the default. I believe the default is ^ $ * . [ ] { } ( ) ? " ! @ # % & / \ , > < ' : ; | _ ~ ` = + - 
as [defined in the developers guide](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-policies.html)
I was unable to pull the password policy with my credentials, the connection was refused. So I do not know if we have altered the special character definition.

lpsinger: It's set to the default:

```
            "PasswordPolicy": {
                "MinimumLength": 8,
                "RequireUppercase": true,
                "RequireLowercase": true,
                "RequireNumbers": true,
                "RequireSymbols": true,
                "TemporaryPasswordValidityDays": 7
            }
```

app/components/Header.tsx
Courey: I tried it and that does not work. It fails to show in the upper right user nav. changing it to && does work.

Courey:
> * [x]  The New Password field should go green once it validates.
> 
>     * [x]  Each item in the password requirement checklist should provide visual feedback on whether it has been met.
> 
>     * [ ]  Provide an appropriate error message if changing the password failed because Cognito rejected the old password.

Still working on the error boundary.

app/components/Header.tsx
lpsinger: Are you sure? I don't see how that can be right. In pseudocode, this should be:

```
if idp is null:
    display reset password
else:
    display nothing
```

If written as `idp && <Component/>`, then if idp is null (which is falsey), then due to short circuit evaluation the right hand side operand (`<Component/>`) will not be evaluated (which is the wrong behavior).

lpsinger (changes requested):
The `/user/password` route (i.e., its loader) should issue a 400 or 404 error if the idp is non-null.

Courey:
Screen Shot 2023-07-06 at 1 16 23 PM
what do you want the error state to look like? Currently it just has the message and a link to retry that just reloads the route.

Courey:
> The `/user/password` route (i.e., its loader) should issue a 400 or 404 error if the idp is non-null.

we can't use the useUserIdp() function in a loader.
Screen Shot 2023-07-06 at 2 07 04 PM
I can get it the long way around with the loader with all the logic we already removed, or I can throw it as it loads the component. 

 I can throw it at the top of the component. that means it would be within the component's error boundary and would currently (I don't know how you want this styled) look like this if idp was not null
Screen Shot 2023-07-06 at 2 05 20 PM

app/components/Header.tsx
Courey: or_isnt_working

Courey: double_and_works

Courey:
Screen Shot 2023-07-06 at 2 25 11 PM
For visual verification of the password character requirements, I have it so that it displays a green check as the requirement is met. Is this acceptable?

app/components/Header.tsx
lpsinger: Ah! some miscommunication here: I had suggested

```jsx
idp || <Component/>
```

but you went with

```jsx
!idp && <Component/>
```

which has the same effect.

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,42 @@
...
+      ProposedPassword: newPassword,
+    }
+
+    const command = new ChangePasswordCommand(passwordData)
+    await cognito.send(command)
+  } catch (error) {
+    if (error instanceof Error) {
+      throw new Error(error.message)
+    }
+  }
```
lpsinger, line 41: Why catch the error at all if we are just re-throwing it?

app/routes/user/password.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { Button, Icon, Label, TextInput } from '@trussworks/react-uswds'
+import { useState } from 'react'
+
+import { updatePassword } from './password.server'
+import Spinner from '~/components/Spinner'
+import { getFormDataString } from '~/lib/utils'
+import { useUserIdp } from '~/root'
+
+export const handle = {
+  breadcrumb: 'Password',
```
lpsinger, line 19: This should match the link text in the nav bar:

```suggestion
  breadcrumb: 'Reset Password',
```

lpsinger (changes requested):
- [x] I don't think that the error boundary is providing the right UX for a mismatch of the old password. Instead, can we highlight the old password field in red and display a small error message beneath that field?
- [x] What is your opinion: should we have a confirm password box, or not? See [Why the Confirm Password Field Must Die](https://uxmovement.medium.com/why-the-confirm-password-field-must-die-aacb9f9c9bce#:~:text=Many%20think%20the%20confirm%20password,to%20type%20their%20password%20twice.)

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,43 @@
...
+    const session = await storage.getSession(request.headers.get('Cookie'))
+    const accessToken = session.get('accessToken')
+    const passwordData = {
+      AccessToken: accessToken,
+      PreviousPassword: oldPassword,
+      ProposedPassword: newPassword,
+    }
+
+    const command = new ChangePasswordCommand(passwordData)
+    await cognito.send(command)
```
lpsinger, line 31: The `try { }` block should enclose _just_ this line (`await cognito.send(command)`).

app/routes/user/password.server.ts
```diff
@@ -0,0 +1,43 @@
...
+  } catch (error) {
+    if (error instanceof Error) {
+      if (error.name === 'NotAuthorizedException') {
+        return 'NotAuthorizedException'
+      }
+      if (error.name === 'LimitExceededException') {
+        return 'LimitExceededException'
+      }
+    }
```
lpsinger, line 40:
```suggestion
    if (error instanceof Error and ['NotAuthorizedException', 'LimitExceededException'].includes(error.name)) {
      return error.name
    } else {
      throw e
    }
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const data = await request.formData()
+  const oldPassword = getFormDataString(data, 'oldPassword')
+  const newPassword = getFormDataString(data, 'newPassword')
+
+  let response = null
+  if (oldPassword && newPassword) {
+    response = await updatePassword(request, oldPassword, newPassword)
+  }
+
+  return response
```
lpsinger, line 47: It's safe here to simply break if oldPassword or newPassword is null, undefined, or an empty string.

```suggestion
  invariant(oldPassword && newPassword)
  return await updatePassword(request, oldPassword, newPassword)
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+}
+
+export default function () {
+  const idp = useUserIdp()
+  if (idp)
+    throw new Error(
+      'you must be logged in with a user name and password to reest password'
+    )
```
lpsinger, line 55: This will result in an "Unexpected error" page. Let's display user-friendly alternative content if the user is logged in but not with username and password, like this:

```ts
function ResetPassword() {
  ...
}

function WrongIdp() {
  ...
}

export default function () {
  return useUserIdp() ? WrongIdp() : ResetPassword()
}
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const containsUpper = /[A-Z]/.test(newPassword)
+  const containsNumber = /[0-9]/.test(newPassword)
+  const validLength = newPassword.length >= 8
+  const containsSpecialChar = /[~`!@#$%^.&*+=\-_ [\]\\';,/{}()|\\":<>?]/g.test(
+    newPassword
+  )
+  const leadingOrTrailingSpace =
+    newPassword.length !== newPassword.trim().length
+  const oldPasswordError = errorMessage && !isOldPasswordTouched
+  const valid = !checkWhitespace(newPassword) && checkPassword(newPassword)
```
lpsinger, line 82: Please construct the value of `valid` as a boolean combination of all of the conditions that you display in the checklist so that we can remove the redundant regex in the `checkPassword` function.

That way, if we ever change the password requirements, we only have to change the code in _one_ place.

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+            {!visible && (
+              <div className="usa-input-suffix" aria-hidden="true">
+                <Icon.Visibility
+                  className="margin-y-2"
+                  onClick={() => {
+                    setVisible(!visible)
+                  }}
+                />
+              </div>
+            )}
```
lpsinger, line 160: - [ ] Refactor to reduce code duplication.
- [ ] Add accessibility attributes to inform screen readers what this control does.

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+            fetcher.data !== 'NotAuthorizedException' &&
+            fetcher.data !== 'LimitExceededException' && (
+              <>
+                <Icon.Check color="green" /> Saved
+              </>
+            )}
+        </fetcher.Form>
+      </>
+
+      <div className="usa-alert usa-alert--info usa-alert--validation">
```
lpsinger, line 187: I believe that there is a react-uswds component for the USWDS alert box.

app/routes/user/password.tsx
```diff
@@ -0,0 +1,276 @@
...
+              {containsLower && (
+                <>
+                  <Icon.Check color="green" />
+                </>
+              )}
+              {!containsLower && (
+                <>
+                  <Icon.Close color="red" />
+                </>
+              )}
```
lpsinger, line 203: - [ ] Remove unnecessary React fragments (`<></>` containing a single React component).
- [ ] Transform `{x && a}{!x && b}` to the equivalent `{x ? a : b}`.

```suggestion
              {containsLower ? <Icon.Check color="green" /> : <Icon.Close color="red" />}
```

Repeat these simplifications for the other checklist items below.

lpsinger (changes requested):

app/routes/user/password.tsx
Courey: The down side to the react-uswds component Alert is that it renders a p tag for the alert text and ul can't be a child of p. 

Screen Shot 2023-07-17 at 2 25 13 PM

Screen Shot 2023-07-17 at 2 22 22 PM

under the hood it does the same thing that I had there, but instead of a p tag, I can use a div and there are no errors.

lpsinger: No, your current solution is fine. Thanks for explaining!

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const checkWhitespace = (str: string) => {
+    return /^\s|\s$/.test(str)
+  }
+
+  const [newPassword, setNewPassword] = useState('')
+  const [isOldPasswordTouched, setIsOldPasswordTouched] = useState(false)
+  const [isNewPasswordTouched, setIsNewPasswordTouched] = useState(false)
+  const [visible, setVisible] = useState(false)
+
+  const containsLower = /[a-z]/.test(newPassword)
```
lpsinger, line 68:
```suggestion
  const whitespaceValid = !(/^\s|\s$/.test(newPassword))
  const containsLower = /[a-z]/.test(newPassword)
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+    </Alert>
+  )
+}
+
+export function ResetPassword() {
+  const fetcher = useFetcher<typeof action>()
+  const errorMessage = fetcher.data
+  const checkWhitespace = (str: string) => {
+    return /^\s|\s$/.test(str)
+  }
```
lpsinger, line 61: No need for this to be a closure...

```suggestion
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const containsNumber = /[0-9]/.test(newPassword)
+  const validLength = newPassword.length >= 8
+  const containsSpecialChar = /[~`!@#$%^.&*+=\-_ [\]\\';,/{}()|\\":<>?]/g.test(
+    newPassword
+  )
+  const leadingOrTrailingSpace =
+    newPassword.length !== newPassword.trim().length
+  const oldPasswordError = errorMessage && !isOldPasswordTouched
+  const valid =
+    !checkWhitespace(newPassword) &&
```
lpsinger, line 79:
```suggestion
    whitespaceValid &&
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+            <li>
+              {containsSpecialChar ? (
+                <Icon.Check color="green" aria-label="green check" />
+              ) : (
+                <Icon.Close color="red" aria-label="red x" />
+              )}
+              At least 1 special character or space
+            </li>
+            <li>
+              {!leadingOrTrailingSpace ? (
```
lpsinger, line 234:
```suggestion
              {whitespaceValid ? (
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+              </>
+            )}
+        </fetcher.Form>
+      </>
+      <div className="usa-alert usa-alert--info usa-alert--validation">
+        <div className="usa-alert__body">
+          <h3 className="site-preview-heading margin-0 usa-alert__heading">
+            New password must contain:
+          </h3>
+          <ul className="usa-checklist" id="validate-code">
```
lpsinger, line 192: The `id` attribute does not seem to be necessary here:

```suggestion
          <ul className="usa-checklist">
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,250 @@
...
+            <li>
+              {containsLower ? (
+                <Icon.Check color="green" aria-label="green check" />
+              ) : (
+                <Icon.Close color="red" aria-label="red x" />
+              )}
+              A lower case letter
+            </li>
```
lpsinger, line 200: This code is repeated enough times, it's worth factoring into a component. For instance:

```suggestion
            <PasswordValidationStep valid={containsLower}>A lower case letter</PasswordValidationStep>
```

lpsinger (changes requested):

app/routes/user/password.tsx
```diff
@@ -0,0 +1,221 @@
...
+            </span>
+          )}
+          <Label htmlFor="newPassword">New Password</Label>
+          <InputGroup error={isError}>
+            <TextInput
+              name="newPassword"
+              id="newPassword"
+              className={
+                !isError && isNewPasswordTouched ? 'usa-input--success' : ''
+              }
```
lpsinger, line 132: The following, I think, is a little bit more comprehensible:

```suggestion
          <InputGroup error={isNewPasswordTouched && !valid}>
            <TextInput
              name="newPassword"
              id="newPassword"
              className={
                valid ? 'usa-input--success' : ''
              }
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,221 @@
...
+    containsSpecialChar
+  const shouldDisableSubmit =
+    !newPassword || !isNewPasswordTouched || !valid || !isOldPasswordTouched
+  let isError
+  if (isNewPasswordTouched && valid) {
+    isError = false
+  }
+  if (isNewPasswordTouched && !valid) {
+    isError = true
+  }
```
lpsinger, line 91:
```suggestion
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const oldPasswordError = errorMessage && !isOldPasswordTouched
+  const valid =
+    whitespaceValid &&
+    containsLower &&
+    containsUpper &&
+    containsNumber &&
+    validLength &&
+    containsSpecialChar
+  const shouldDisableSubmit =
+    !newPassword || !isNewPasswordTouched || !valid || !isOldPasswordTouched
```
lpsinger, line 84: If the new password is valid, then it is non-empty and it has been touched.

```suggestion
  const shouldDisableSubmit = !valid || !isOldPasswordTouched
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,221 @@
...
+          </ButtonGroup>
+          {fetcher.state !== 'idle' && (
+            <>
+              <Spinner /> Saving...
+            </>
+          )}
+          {fetcher.state === 'idle' &&
+            fetcher.data !== undefined &&
+            fetcher.data !== 'NotAuthorizedException' &&
+            fetcher.data !== 'LimitExceededException' && (
```
lpsinger, line 178:
```suggestion
            fetcher.data === null
```

lpsinger (changes requested):

lpsinger (changes requested):
The "Saving..." spinner appears below the Save button, but it should appear to the side.

# This page

Screenshot 2023-07-17 at 18 03 10

# Desired appearance (other existing pages)

Screenshot 2023-07-17 at 18 03 43

Screenshot 2023-07-17 at 18 03 28

lpsinger (changes requested):
Please highlight the "old password" field in red when it errors, and clear it when the user starts typing in it.

app/routes/user/password.tsx
Courey: I'll meet in the middle here. I took out the isNewPasswordTouched check but I left the !newPassword check in because when I don't, on firefox I get the following error. As long as the new password check is included it is happy. 
Screen Shot 2023-07-18 at 12 19 33 PM

app/routes/user/password.tsx
```diff
@@ -0,0 +1,211 @@
...
+              Reset Password
+            </Button>
+            {fetcher.state !== 'idle' && (
+              <>
+                <Spinner className="margin-top-3" /> Saving...
+              </>
+            )}
+            {fetcher.state === 'idle' && fetcher.data === null && (
+              <>
+                <Icon.Check className="margin-top-3" color="green" /> Saved
```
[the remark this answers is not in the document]
Courey: If you look up on line 158 you'll see that I put a margin on the button. I did this to give it a little space between the last text input and the button. I think that since they are children of the same parent, when the spinner/check render, they render without the same margin as their sibling element. 
Screen Shot 2023-07-18 at 3 41 27 PM
Screen Shot 2023-07-18 at 3 41 42 PM

app/routes/user/password.tsx
```diff
@@ -0,0 +1,210 @@
...
+                <>
+                  <Spinner /> Saving...
+                </>
+              )}
+              {fetcher.state === 'idle' && fetcher.data === null && (
+                <>
+                  <Icon.Check color="green" /> Saved
+                </>
+              )}
+            </ButtonGroup>
```
lpsinger, line 170: To fix this vertical space issue:

Screenshot 2023-07-19 at 10 44 06

```suggestion
              <Button disabled={shouldDisableSubmit} type="submit">
                Reset Password
              </Button>
              {fetcher.state !== 'idle' && (
                <span className="text-middle">
                  <Spinner /> Saving...
                </span>
              )}
              {fetcher.state === 'idle' && fetcher.data === null && (
                <span className="text-middle">
                  <Icon.Check color="green" /> Saved
                </span>
              )}
```

lpsinger (changes requested):
Remember to clear the "Saved" message if the user starts filling in the form again. (Look at how the "dirty" flag works on some other similar forms in this project.)

app/routes/user/password.tsx
```diff
@@ -0,0 +1,223 @@
...
+          {errorMessage === 'NotAuthorizedException' && !interacted && (
+            <div className="text-red margin-bottom-105 margin-top-2px">
+              Invalid Password
+            </div>
+          )}
+          {errorMessage === 'LimitExceededException' && !interacted && (
+            <div className="text-red margin-bottom-105 margin-top-2px">
+              Attempts Exceeded. Please try again later.
+            </div>
+          )}
```
lpsinger, line 131: - [ ] Remember to reserve space for the message, so that the display doesn't bounce around: if neither message is displayed, then you can show a `<div>` containing nothing but `&nbsp;`. Also, you can lift the style `<div>` outside to eliminate duplicate code.
- [ ] Clear the error _when the user starts changing the old password field_, not when the user interacts with _any_ field in the form.

lpsinger (changes requested):

app/routes/user/password.tsx
lpsinger: @Courey, do you have sufficient access in GitHub to tick off the checklist items in the above comment?

app/routes/user/password.tsx
```diff
@@ -105,31 +119,21 @@ export function ResetPassword() {
...
-              Invalid Password
-            </div>
-          )}
-          {errorMessage === 'LimitExceededException' && !interacted && (
-            <div className="text-red margin-bottom-105 margin-top-2px">
-              Attempts Exceeded. Please try again later.
-            </div>
-          )}
-          <Label htmlFor="newPassword" className="margin-top-0">
+          {ErrorMessage(errorMessage, interacted)}
```
lpsinger, line 135: Use JSX syntax here:

```suggestion
          <ErrorMessage errorMessage={errorMessage} interacted={interacted} />
```

app/routes/user/password.tsx
```diff
@@ -54,6 +54,18 @@ export function WrongIdp() {
   )
 }
 
+export function ErrorMessage(errorMessage: any, interacted: boolean) {
```
lpsinger, line 57: No need to export this function:

```suggestion
function ErrorMessage(errorMessage: any, interacted: boolean) {
```

lpsinger (changes requested):
Again: please make sure that _only editing the old password field clears the old password error message_. Editing the _new password field_ should _not_ clear the old password error message.

app/components/PasswordValidationStep.tsx
```diff
@@ -0,0 +1,20 @@
+import { Icon } from '@trussworks/react-uswds'
+
+export default function PasswordValidationStep({
```
lpsinger, line 3: I don't see us using this outside of the context of the change password form. Please move this component inside password.tsx.

app/components/Header.tsx
```diff
@@ -132,6 +133,15 @@ export function Header() {
                       >
                         Peer Endorsements
                       </Link>,
+                      !idp && (
+                        <Link
+                          key="password"
+                          to="/user/password"
+                          onClick={() => setUserMenuIsOpen(!userMenuIsOpen)}
```
lpsinger, line 140: Rebase from the main branch. This should now be:

```suggestion
                          onClick={onClickUserMenuItem}
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,230 @@
...
+export async function action({ request }: DataFunctionArgs) {
+  const data = await request.formData()
+  const oldPassword = getFormDataString(data, 'oldPassword')
+  const newPassword = getFormDataString(data, 'newPassword')
+
+  invariant(oldPassword && newPassword)
+  return await updatePassword(request, oldPassword, newPassword)
+}
+
+export function WrongIdp() {
```
lpsinger, line 49: Don't export this:

```suggestion
function WrongIdp() {
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,230 @@
...
+  ) {
+    return (
+      <div className="text-red">Attempts Exceeded. Please try again later.</div>
+    )
+  } else {
+    return <div>&nbsp;</div>
+  }
+}
+
+export function ResetPassword() {
```
lpsinger, line 72: Don't export this:

```suggestion
function ResetPassword() {
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,230 @@
...
+            id="oldPassword"
+            type="password"
+            placeholder="Old Password"
+            onChange={(e) => {
+              setOldPassword(e.target.value)
+              setInteracted(true)
+              setIsOldPasswordTouched(true)
+            }}
+          />
+          {ErrorMessage(errorMessage, isOldPasswordTouched)}
```
lpsinger, line 138: Use JSX syntax:

```suggestion
          <ErrorMessage errorMessage={errorMessage} isOldPasswordTouched={isOldPasswordTouched} />
```

app/routes/user/password.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+export function WrongIdp() {
+  return (
+    <Alert type="error" heading="Error:" headingLevel="h4">
+      You must be signed in with a username and password to reset your password.
+    </Alert>
+  )
+}
+
+function ErrorMessage(errorMessage: any, isOldPasswordTouched: boolean) {
```
lpsinger, line 57: Transform this to a React functional component so that we can invoke this using JSX syntax:

```suggestion
function ErrorMessage({errorMessage, isOldPasswordTouched}: {errorMessage: any, isOldPasswordTouched: boolean}) {
```

lpsinger (changes requested):

lpsinger (changes requested):
There's another bug here: when the form clears, the green outline around the New Password field remains. It should clear away.

Maybe to keep this simple, we just never display the green border? That would make the New Password and Old Password fields behave more consistently with each other.

https://github.com/nasa-gcn/gcn.nasa.gov/assets/728407/6c8b2a22-ae9b-4dc6-a84e-a8e90a956dcd

lpsinger (approved):

Courey:
> Maybe to keep this simple, we just never display the green border? That would make the New Password and Old Password fields behave more consistently with each other.
I think this makes sense. There is enough visual feedback that it's in an acceptable state that I think it's redundant. I removed it.

lpsinger:
Phew! What a tricky PR. I'll squash and merge once the workflow finishes.

PR #1088, opened 2023-07-05
using light theme to avoid low contrast menu items

Courey:
This is a quick fix option to avoid the mobile contrast issue in the user navigation menu.

### DIFF
--- a/build/theme/custom.scss
+++ b/build/theme/custom.scss
@@ -548,3 +548,23 @@ a {
   margin-top: 0;
   margin-bottom: 0;
 }
+
+/*
+ * Override user nav text colors to add contrast when not desktop
+ */
+.usa-nav__primary a:not(.usa-button):not(.usa-current):not(.active) {
+  @media (max-width: 63.99em) {
+    color: white;
+  }
+}
+.usa-nav__primary a:not(.usa-button):not(.usa-current):hover {
+  @media (max-width: 63.99em) {
+    color: #0050d8;
+  }
+}
+
+ul#user {
+  @media (max-width: 63.99em) {
+    background-color: #1b1b1b;
+  }
+}
--- a/build/theme/custom.scss
+++ b/build/theme/custom.scss
@@ -551,6 +551,7 @@ a {
 
 /*
  * Override user nav text colors to add contrast when not desktop
+ * FIXME: remove once https://github.com/bruffridge/nasawds/issues/10 is fixed
  */
 .usa-nav__primary a:not(.usa-button):not(.usa-current):not(.active) {
   @media (max-width: 63.99em) {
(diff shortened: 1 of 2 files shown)


lpsinger:
I like the idea, but this adversely affects the appearance of the nav when the windows is desktop size:

Screenshot 2023-07-05 at 10 18 54

Is there a straightforward way to have the mobile nav use the light theme and the desktop nav use the dark theme?

Courey:
yeah, I'm working on seeing if there is a quick fix for that currently. I just wasn't sure you were okay with the mobile menu looking like that.

Courey:
Screen Shot 2023-07-05 at 10 36 37 AM
This preserves dark on desktop and light on mobile. The only thing that is not my favorite is if you take the window and shrink it down to mobile size, when it hits 768 pixels wide, you can see the nav bar change colors. I don't know that it's avoidable if we use different styles based on size. 

I'm still seeing if there is an easy quick fix that is a more consistent experience.

lpsinger:
> This preserves dark on desktop and light on mobile. The only thing that is not my favorite is if you take the window and shrink it down to mobile size, when it hits 768 pixels wide, you can see the nav bar change colors. I don't know that it's avoidable if we use different styles based on size.
> I'm still seeing if there is an easy quick fix that is a more consistent experience.

This has the unwanted side effect of changing the background of the header itself:

Screenshot 2023-07-05 at 15 26 37

This is all a stopgap measure, right? In the long run, we would like to fix USWDS so that the links in the dark themed mobile nav have high contrast colors, like the white color that is used for nested menus like the "leo.p.singer@nasa.gov +" link in the screen shot below.

Screenshot 2023-07-05 at 15 28 29

Courey:
yes, just a stopgap. This is just to get contrast quickly while I find a better solution. Alternative is waiting till I find that solution to do anything at all. I don't know how much of an issue it has been for users.

Courey:
Screen Shot 2023-07-07 at 11 00 05 AM
Screen Shot 2023-07-07 at 11 00 23 AM
Here's a better version. it should only impact when not desktop width and preserves highlighting

build/theme/custom.scss
```diff
@@ -548,3 +548,23 @@ a {
   margin-top: 0;
   margin-bottom: 0;
 }
+
+/*
+ * Override user nav text colors to add contrast when not desktop
```
lpsinger, line 553:
```suggestion
 * Override user nav text colors to add contrast when not desktop
 * FIXME: remove once https://github.com/bruffridge/nasawds/issues/10 is fixed
```

lpsinger (changes requested):

lpsinger (approved):

PR #1092, opened 2023-07-06
Adds an optional note field for endorsements

### LINKED ISSUE #847: Require user to set profile before sending peer endorsement
I got a peer endorsement request from a users with an ambiguous email address.  If they had their full name and affiliation set in the system and the request came with that it would be much easier to evaluate.  An optional ability to write a sentence explaining who they are would also help.

dakota002:
Closes #847, addressing the comment of adding a note field 

Updated UI for Requestor:
image

Collapsible display for endorser:

image

image


lpsinger:
Can you have the expanded text span the width of the card?

dakota002:
Sure! I'll get a commit out shortly for that

dakota002:
Here's the updated version:
image

lpsinger:
What if you move "show less" up so that the alignment of the username does not change when you expand the message?

dakota002:
> What if you move "show less" up so that the alignment of the username does not change when you expand the message?

I think it looks really good that way on the desktop view:
image

But it looks a little strange on the mobile
image

I'm seeing if I can balance it out a bit more though

dakota002:
Fixed the mobile:

image

lpsinger:
That looks better.

app/routes/user/endorsements.server.ts
```diff
@@ -87,7 +88,7 @@ export class EndorsementsServer {
    *   - User `endorserSub` does not have an email attribute
    *   - There is an existing endorsement with the same requester and submitter sub
    */
-  async createEndorsementRequest(endorserSub: string) {
+  async createEndorsementRequest(endorserSub: string, note?: string) {
```
lpsinger, line 91: The DynamoDB API calls will probably fail if `note` is `undefined`, which is permitted by the type signature. DynamoDB generally seems to support `null` but not `undefined`.

lpsinger (changes requested):
- [ ] Add markup to indicate to screen readers the disclosure button nature of the new field: i.e.  `<details>` elements or aria attributes.
- [x] Let's make it clear in the emails and in the approver's view that the note text came _from the other user_ and is not part of the content of our site. It might be sufficient to just label it as "Comment from the user:"?
- [x] I see that we are making a decision on whether or not to display the disclosure button based on the character length of the comment. Can we instead use CSS overflow-related attributes so that it is based on the layout itself?
- [x] Please use a disclosure button icon. Spelunk around the app to find some examples. The new circulars form has one.

lpsinger (commented):
What is the rationale for not using the ButtonGroup component?

lpsinger:
> Which style do you prefer?

Yuck. Are we overriding any styles that would cause the ButtonGroup to look like that? I had thought that ButtonGroup one mobile fills the page width. Is that only for segmented buttons?

dakota002:
Ahh I see what happened, I needed the button group wrapped in a div to get it to apply the correct spacing:
image

app/routes/user/endorsements.tsx
```diff
@@ -215,49 +222,101 @@ export function EndorsementRequestCard({
...
   const fetcher = useFetcher()
   const disabled = fetcher.state !== 'idle'
+  const [showMore, setShowMore] = useState(false)
+  const [isOverflowing, setIsOverflowing] = useState(true)
+  const ref = useRef<HTMLDivElement>(null)
+
+  function overflowing() {
+    if (ref.current)
+      setIsOverflowing(ref.current?.offsetWidth < ref.current?.scrollWidth)
+  }
```
lpsinger, line 232: There's really no simple way to do this with CSS?

app/routes/user/endorsements.tsx
```diff
@@ -215,49 +222,101 @@ export function EndorsementRequestCard({
...
+
+  useEffect(() => overflowing(), [])
```
lpsinger, line 234: This dependency list is probably incorrect. It at least depends on `ref.current`.

app/routes/user/endorsements.tsx
```diff
@@ -215,49 +222,101 @@ export function EndorsementRequestCard({
...
 
   return (
     <fetcher.Form method="POST">
       {role === 'endorser' ? (
         <Grid row style={disabled ? { opacity: '50%' } : undefined}>
-          <div className="tablet:grid-col flex-fill">
+          <div className="tablet:grid-col flex-fill order-first">
             <div className="margin-y-0">
-              <strong>{endorsementRequest.requestorEmail}</strong>
+              <strong className="">{endorsementRequest.requestorEmail}</strong>
```
lpsinger, line 242: Why the empty `className` property?

app/routes/user/endorsements.tsx
```diff
@@ -215,49 +222,101 @@ export function EndorsementRequestCard({
...
+                      <>
+                        more <Icon.ArrowDropDown />
+                      </>
+                    )}
+                  </Button>
+                )}
+              </div>
+            )}
+          </div>
+          <div className="tablet:grid-col flex-auto order-last">
```
lpsinger, line 270: We lost the `flex-center` prop here. Why?

app/routes/user/endorsements.tsx
```diff
@@ -215,49 +222,101 @@ export function EndorsementRequestCard({
...
+              >
+                Approve
+              </Button>
+              <Button
+                type="submit"
+                outline
+                disabled={disabled}
+                name="intent"
+                value="rejected"
+                className="margin-y-1 tablet:margin-y-0"
```
lpsinger, line 291: What do the manual margin settings do here?

app/routes/user/endorsements.tsx
```diff
@@ -466,17 +525,27 @@ export function EndorsementRequestForm() {
...
+      <Grid row>
+        <div className="tablet:grid-col-fill">
+          <TextInput
+            id="note"
+            name="note"
+            type="text"
+            placeholder="Note for endorser"
+          />
+        </div>
+        <div className="tablet:grid-col-auto">
```
lpsinger, line 538: There's a way to write this with the `<Grid>` component. Can we do that, please?

app/routes/user/endorsements.tsx
dakota002: Oops, removing this

app/routes/user/endorsements.tsx
dakota002: It squeezes the buttons down, 

With:
image

Without:
image

app/routes/user/endorsements.tsx
dakota002: Nothing anymore! removing these too

app/routes/user/endorsements.tsx
dakota002: Not really, but it does work as expected

app/routes/user/endorsements.tsx
lpsinger: Mightn't you need to revalidate if the component changes dimensions, due to a change in the window size or device orientation. Would something like [useMesaure](https://usehooks.com/usemeasure) help here?

app/routes/user/endorsements.tsx
```diff
@@ -466,13 +522,23 @@ export function EndorsementRequestForm() {
...
         </Grid>
-        <Grid col="auto">
+      </Grid>
+      <Grid row>
+        <Grid tablet={{ col: 'fill' }}>
+          <TextInput
+            id="note"
+            name="note"
+            type="text"
+            placeholder="Note for endorser"
```
lpsinger, line 532: How about:

```suggestion
            placeholder="Optional message to the endorser"
```

lpsinger (changes requested):

app/routes/user/endorsements.tsx
dakota002: It looks useful, but is it supported by typescript? I tried adding the package but had errors. I found https://usehooks-ts.com/ but that does not have a useMeasure hook. Also the page for [useMeasure](https://usehooks.com/usemeasure) is showing incorrect info, there are two issues about it on their github

lpsinger: https://usehooks-ts.com/react-hook/use-element-size

dakota002: thats handy, implemented it in the latest commit

lpsinger (approved):

lpsinger (changes requested):
Looks good! The only issue that I see is the merge conflict. Please fix.

dakota002:
Should be resolved now!

lpsinger (approved):
Once it is deployed to dev, please test it live thoroughly.

PR #1101, opened 2023-07-10
Circulars date range search frontend

### LINKED ISSUE #878: Limit search by date range

tylerbarna:
Resolves #878 

In-progress work for adding a frontend for the date range searching.

- [x] Implement a functional date range search object that plays well with search
- [x] Use tooltip or similar to create custom css object
- [ ] Implement custom css object to pull up a date range search object 
- [ ] associate that date range search object with the calendar button in the search bar
- [ ] (minor) figure out icon color for calendar icon (should be white to match other icons)


tylerbarna:
Rough idea of what the final product should look like:

image

lpsinger:
> Rough idea of what the final product should look like:

That looks great, with one difference: I would like to keep the search term field and the submit button with the looking glass icon together, so the date range control should be to the right of it.

tylerbarna:
Also need to resolve the anchoring issue with the dropdown content, as it current renders the dropdown at the start of the button group rather than at the calendar button

tylerbarna:
excessively stale PR caused some merge conflicts I was having trouble solving, so I'm closing this PR and opening a fresh one on a more updated branch (#1265)

PR #1104, opened 2023-07-11
Choose number of results to display on circulars search page

### LINKED ISSUE #1103: Add option for number of results in circulars
For the circulars search, the functionality already exists inside of the search and pagination functions to alter the limit on the number of results displayed on a given page. This really just requires tweaking the default and loader functions and either adding the limit as a url parameter or as another type of state if we don't want to clutter the url parameters too much, though I do believe it will require a reload upon changing the limit so the pagination function can be rerun with the new limit

tylerbarna:
Resolves #1103 

This alters the circulars page layout slightly so the new circulars button is in the upper right hand corner, inline with the page title, to create more space on the search line so a dropdown menu can be added for choosing the number of results to display. The new circular button has also had its text changed to "submit new circular."

The loader and default functions have been changed to now interpret the limit argument as a parameter similar to the query or page number, and there is a dropdown menu with 3 selections for the result limit per page (current default is 100). 

This is being submitted as a draft PR because of the following issues:
- The submit new circular button needs to be right justified or placed in an agreed upon position. The same is true of the result limit dropdown, though it may be better to have this shown on a seperate line (though this might be impacted by needing to remain in the form element)
- Currently, the dropdown menu selector doesn't actually work. I think this is something to do with how the dropdown element handles onChange, but that needs to be checked.


tylerbarna:
choose number of results is now functional, this is the new UI:
image

tylerbarna:
found an edge case issue where changing the number of results without submitting a query fails to load results; this is probably related to the empty query parameter being added to the url when the page limit is changed

dakota002:
@tylerbarna is this ready for review or are you still working on it?

lpsinger:
I'll take it for review.

tylerbarna:
> @tylerbarna is this ready for review or are you still working on it?

I noticed the issue relating to changing the number of results with a blank query not working, but I haven't been able to resolve that issue yet.

the url parameters will update to something like "query=&limit=500," and, upon reloading the page, the results will display as expected, but changing the limit again requires a manual reload as well.

dakota002:
> > @tylerbarna is this ready for review or are you still working on it?
> 
> I noticed the issue relating to changing the number of results with a blank query not working, but I haven't been able to resolve that issue yet.
> 
> the url parameters will update to something like "query=&limit=500," and, upon reloading the page, the results will display as expected, but changing the limit again requires a manual reload as well.

Is this possibly related to the issue with the "clean" flag that we started looking at yesterday?

lpsinger (changes requested):
For the search form at the top of this page, I prefer the interface to be as slim as possible --- avoiding labels that are separate from the input fields and using inline placeholder text instead.

Given that, would you please try putting the number of results selector down by the pagination control, similar to what ADS does?

Screenshot 2023-07-13 at 11 12 25

app/routes/circulars/circulars.server.ts
```diff
@@ -128,7 +128,7 @@ export async function search({
       subject,
     })
   )
-
+
```
tylerbarna, line 131:
```suggestion

```

tylerbarna:
> found an edge case issue where changing the number of results without submitting a query fails to load results; this is probably related to the empty query parameter being added to the url when the page limit is changed

Issue #1110 needs to be fixed to address this

lpsinger:
Please resolve conflicts.

lpsinger (changes requested):
Please fix merge conflicts.

app/routes/circulars/index.tsx
```diff
@@ -204,14 +223,10 @@ export default function () {
...
       </p>
-      <ButtonGroup className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1">
-        <Form
-          className="display-inline-block usa-search usa-search--small"
-          role="search"
-        >
-          <Label srOnly={true} htmlFor="query">
-            Search
-          </Label>
+      <div className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1">
```
lpsinger, line 226: Why did you change this from a ButtonGroup to a div?

app/routes/circulars/index.tsx
```diff
@@ -204,14 +223,10 @@ export default function () {
...
+        <Form className="usa-search usa-search--small width-full">
+          <Label htmlFor="query">Search</Label>
```
lpsinger, line 228: This label used to be for screen readers only (`srOnly={true}`). Why did you make it visible?

app/routes/circulars/index.tsx
```diff
@@ -55,17 +55,20 @@ export async function loader({ request: { url } }: DataFunctionArgs) {
 
 function getPageLink({
   page,
+  limit,
   query,
   startDate,
   endDate,
 }: {
   page: number
+  limit?: string
```
lpsinger, line 64: Please follow the example of the `page` argument here and accept the `limit` argument as a number.

```suggestion
  limit?: number
```

app/routes/circulars/index.tsx
```diff
@@ -191,7 +196,21 @@ export default function () {
...
+          <Link to={`/circulars/new${searchString}`}>
+            <Button
+              type="button"
+              className="height-4 padding-top-0 padding-bottom-0"
+            >
+              <Icon.Edit /> Submit New Circular
+            </Button>
+          </Link>
+        </div>
+      </Grid>
```
lpsinger, line 213: These UX changes appear to be unrelated to the number of results.

lpsinger (changes requested):
Please clean up unrelated changes.

app/routes/circulars/index.tsx
tylerbarna: converted to number type in [2487eb6](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1104/commits/2487eb68d08cd1ecbf7fb90d62dd4a41b3d7142a)

app/routes/circulars/index.tsx
tylerbarna: this was removed by mistake and overlooked. Fixed in [45cf283](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1104/commits/45cf2833f6c53d26fc129ec4e1942986c630024a)

app/routes/circulars/index.tsx
tylerbarna: I believe this was originally changed after the repositioning of the end of the Form element, as the ButtonGroup didn't work, but I see now that changing it to a div removes the sticky property. I was able to restore it to a ButtonGroup by changing the order of the Form start and where the ButtonGroup starts. changes made in [c206c92](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1104/commits/c206c92cc44cfd49a8657c1ca3df6fa70ebc1f66)

app/routes/circulars/index.tsx
tylerbarna: > These UX changes appear to be unrelated to the number of results.

image

This was a change made when I had originally placed the result selector dropdown inline with the search box, but it's not entirely necessary now that the result selector is at the bottom of the page. My rationale was that repositioning the new circular button would reduce the business of the search bar/make it so controls inline with the search bar were only those related to searching. Moving it to the top of the page also made it a bit quicker to find for those only interested in submitting a new circular

lpsinger: It's a reasonable idea, but it's important to keep each PR focused on _one_ change, bug fix, or feature. If you want to pursue this design change, then it needs to be in a separate PR.

tylerbarna: ah, yeah, that's a fair point. In order to revert it to have the same appearance as before and also pull the Form element out of the button group, I had to do a nested button group type structure in order for the search and new button to have the same appearance as before the form and button group element order had to be changed.

tylerbarna:
> There are still an awful lot of lines of code hit here that don't seem to have anything to do with the topic of the PR. Please look at the diff and try to undo any changes that are _not_ related to the topic of the PR.
> 
> I am perplexed by the nested button group. The style and positioning of the search box worked before. I don't see why we need to change it in order to support a dropdown that is somewhere else entirely on the page. The nested button group workaround seems contrived anyway.
> 
> I do understand that some of the structural changes here are due to moving the `<Form>` component up in the tree. However, form elements do not have any inherent styling associated with them; they are purely logical elements. So you should be able to replace this:
> 
> ```
> ...misc markup...
> <form class="foo">
> ...more misc markup...
> </form>
> ...yet more misc markup...
> ```
> 
> with this:
> 
> ```
> <form>
> ...misc markup...
> <div class="foo">
> ...more misc markup...
> </div>
> ...yet more misc markup...
> </form>
> ```
> 
> without any visual difference. Notice how the `<form>` element moved up, but in its place is now a `<div>` with the same CSS class as the original `<form>` element?
> 
> If the above does result in visual elements, well, then that means that the USWDS style sheet is attaching some style to forms. In that event, please leave the form where it was and as it was, and instead use the [`form` attribute of the `<input>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form) to associate the detached form inputs. Example:
> 
> ```
> <form id="foo">
> ...
> </form>
> ...
> 
> <input form="id">
> ```

I was unaware of being able to associate a detached element with a specific form by assigning the form an id; that negates the need to relocate the form and most of the styling challenges I was encountering. From what I can tell, the form element is styled in some way by react-uswds. To my knowledge, all changes not necessary for adding the result selection element and associated functionality have been removed with 4dbbe59

app/routes/circulars/index.tsx
```diff
@@ -263,15 +269,41 @@ export default function () {
...
+                <Label className="margin-top-auto" htmlFor="value">
+                  Results Per Page
+                </Label>
+                <Dropdown
+                  id="limit"
+                  className="usa-select height-4 padding-y-0"
+                  name="limit"
+                  defaultValue={limit}
+                  form="searchForm"
+                  onChange={({ target: { form, value } }) => {
```
lpsinger, line 284: Unused variable:

```suggestion
                  onChange={({ target: { form } }) => {
```

app/routes/circulars/index.tsx
```diff
@@ -263,15 +269,41 @@ export default function () {
...
+                    submit(form)
+                  }}
+                >
+                  <option value="100">100</option>
+                  <option value="250">250</option>
+                  <option value="500">500</option>
```
lpsinger, line 290: Let's make 100 the _maximum_ number of results per page.

```suggestion
                  <option value="10">10</option>
                  <option value="20">20</option>
                  <option value="50">50</option>
                  <option value="100">100</option>
```

Also, we need to validate _on the server side_ that the requested number of results is less than an or equal to the maximum.

lpsinger (changes requested):

app/routes/circulars/index.tsx
tylerbarna: > Let's make 100 the _maximum_ number of results per page.
> 
> Also, we need to validate _on the server side_ that the requested number of results is less than an or equal to the maximum.

would the default number of results now be 10 rather than 100?

as for the server side validation, are we essentially preventing more than 100 results being displayed?

lpsinger: > would the default number of results now be 10 rather than 100?

I don't think so. 100 is still a good default.

> as for the server side validation, are we essentially preventing more than 100 results being displayed?

Yes.

app/routes/circulars/index.tsx
tylerbarna: moved this to a separate pr (#1142)

app/routes/circulars/index.tsx
```diff
@@ -42,6 +41,11 @@ export async function loader({ request: { url } }: DataFunctionArgs) {
   const startDate = searchParams.get('startDate') || undefined
   const endDate = searchParams.get('endDate') || undefined
   const page = parseInt(searchParams.get('page') || '1')
+  const limitRequest = parseInt(searchParams.get('limit') || '100')
+  var limit = 100
+  if (limitRequest < 100) {
+    limit = limitRequest
+  }
```
lpsinger, line 48: Try [`clamp` from lodash](https://lodash.com/docs/4.17.15#clamp) here.

app/routes/circulars/index.tsx
```diff
@@ -263,15 +273,42 @@ export default function () {
...
+            <div className="flex-align-start tablet:flex-fill">
+              <div>
+                <Label className="margin-top-auto" htmlFor="value">
+                  Results Per Page
+                </Label>
+                <Dropdown
+                  id="limit"
+                  className="usa-select height-4 padding-y-0"
+                  name="limit"
+                  defaultValue={'100'}
```
lpsinger, line 286:
```suggestion
                  defaultValue="100"
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
tylerbarna: oh, that's much cleaner, thanks for the tip

app/routes/circulars/index.tsx
```diff
@@ -23,6 +23,7 @@ import {
   TextInput,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
+import { clamp } from 'lodash'
```
lpsinger, line 26: Load just the function that we need, to avoid bundling all of lodash.

```suggestion
import clamp from 'lodash/clamp'
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
```diff
@@ -177,6 +182,7 @@ export default function () {
   const allItems = [...(newItem ? [newItem] : []), ...(items || [])]
 
   const [searchParams] = useSearchParams()
+  const limit = searchParams.get('limit') ?? '100'
   const query = searchParams.get('query') ?? undefined
   const startDate = searchParams.get('startDate') ?? undefined
   const endDate = searchParams.get('endDate') ?? undefined
```
lpsinger, line 188: This fixes the bug where changing the limit causes the results to disappear:

```suggestion
  const limit = searchParams.get('limit') || '100'
  const query = searchParams.get('query') || undefined
  const startDate = searchParams.get('startDate') || undefined
  const endDate = searchParams.get('endDate') || undefined
```

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,42 @@ export default function () {
...
-              startDate={startDate}
-              endDate={endDate}
-            />
-          )}
+          <div className="display-flex flex-row flex-wrap">
+            <div className="flex-align-start tablet:flex-fill">
+              <div>
+                <Label className="margin-top-auto" htmlFor="value">
+                  Results Per Page
+                </Label>
```
lpsinger, line 278:
```suggestion
```

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,42 @@ export default function () {
...
+                  defaultValue="100"
+                  form="searchForm"
+                  onChange={({ target: { form } }) => {
+                    submit(form)
+                  }}
+                >
+                  <option value="10">10</option>
+                  <option value="20">20</option>
+                  <option value="50">50</option>
+                  <option value="100">100</option>
```
lpsinger, line 292:
```suggestion
                  <option value="10">10 / page</option>
                  <option value="20">20 / page</option>
                  <option value="50">50 / page</option>
                  <option value="100">100 / page</option>
```

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,42 @@ export default function () {
...
-              endDate={endDate}
-            />
-          )}
+          <div className="display-flex flex-row flex-wrap">
+            <div className="flex-align-start tablet:flex-fill">
+              <div>
+                <Label className="margin-top-auto" htmlFor="value">
+                  Results Per Page
+                </Label>
+                <Dropdown
```
lpsinger, line 279:
```suggestion
                <Select
```

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,42 @@ export default function () {
...
+                  form="searchForm"
+                  onChange={({ target: { form } }) => {
+                    submit(form)
+                  }}
+                >
+                  <option value="10">10</option>
+                  <option value="20">20</option>
+                  <option value="50">50</option>
+                  <option value="100">100</option>
+                </Dropdown>
```
lpsinger, line 293:
```suggestion
                </Select>
```

app/routes/circulars/index.tsx
```diff
@@ -17,11 +17,13 @@ import {
 import {
   Button,
   ButtonGroup,
+  Dropdown,
   Icon,
   Label,
```
lpsinger, line 22: `Dropdown` is deprecated and replaced by `Select`.

```suggestion
  Icon,
  Label,
  Select,
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,42 @@ export default function () {
...
-          )}
+          <div className="display-flex flex-row flex-wrap">
+            <div className="flex-align-start tablet:flex-fill">
+              <div>
+                <Label className="margin-top-auto" htmlFor="value">
+                  Results Per Page
+                </Label>
+                <Dropdown
+                  id="limit"
+                  className="usa-select height-4 padding-y-0"
```
lpsinger, line 281:
```suggestion
                  className="width-auto"
```

lpsinger (changes requested):

app/routes/circulars/index.tsx
tylerbarna: Updated dropdown:
image

lpsinger (changes requested):
This is very close now!

At small screen sizes, the two controls overlap a little bit, and it becomes apparent that they are not vertically aligned (my last suggested patch might have broken the latter).

Screenshot 2023-07-21 at 18 07 59

tylerbarna:
9837c1c changes:

Desktop:
image

Mobile:
image

Edge case (640p screen):
image
Edge case is fixed by adding a small margin to the right side of the select menu

lpsinger (changes requested):

app/routes/circulars/index.tsx
```diff
@@ -271,11 +271,11 @@ export default function () {
             ))}
           </ol>
           <div className="display-flex flex-row flex-wrap">
-            <div className="flex-align-start tablet:flex-fill">
+            <div className="display-flex flex-align-self-center margin-right-2 width-auto">
               <div>
                 <Select
                   id="limit"
-                  className="width-auto"
+                  className="usa-select height-5 padding-y-0 margin-y-0"
```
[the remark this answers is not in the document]
tylerbarna: removing usa-select from the className has some odd effects on the styling, specifically for mobile 

with usa-select:
image

Without usa-select:
image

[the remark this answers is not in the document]
lpsinger: Why is that? https://github.com/trussworks/react-uswds/blob/main/src/components/forms/Select/Select.tsx#L30-L37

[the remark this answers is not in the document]
tylerbarna: I'm not entirely sure, though from the react-uswds docs, they seem to include usa-select in the className https://designsystem.digital.gov/components/select/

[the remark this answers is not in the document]
tylerbarna: > Ah. It's not the `usa-select` class, it's `width-auto`. Remove `usa-select` but add back `width-auto`.

oh okay. Could you by any chance make that a suggested change? For some reason, [the commit I just made](https://github.com/tylerbarna/gcn.nasa.gov/commit/97898aac10c7f1e9be284b4e45f6f4da8e2faffa) is showing in my fork but isn't reflected in the pr (also not sure why it's saying I made the commit 16 hours ago)

lpsinger:
This is nice!

How would it look if we left-aligned the pagination component? Most of the rest of the page is left-aligned.

Screenshot 2023-07-25 at 13 40 20

tylerbarna:
> How would it look if we left-aligned the pagination component? Most of the rest of the page is left-aligned.

First Page:
Desktop:
image
Mobile:
image

Middle Pages:
Desktop:
image
image

Mobile:
image
image


Last Page:
Desktop:
image

Mobile:
image

app/routes/circulars/index.tsx
```diff
@@ -263,15 +270,39 @@ export default function () {
...
+                  }}
+                >
+                  <option value="10">10 / page</option>
+                  <option value="20">20 / page</option>
+                  <option value="50">50 / page</option>
+                  <option value="100">100 / page</option>
+                </Select>
+              </div>
+            </div>
+            <div className="display-flex flex-justify-center flex-fill flex-align-self-center">
```
tylerbarna, line 293: In the event that we want to left justify the pagination

```suggestion
            <div className="display-flex flex-fill">
```

lpsinger:
> > How would it look if we left-aligned the pagination component? Most of the rest of the page is left-aligned.

Yes, I do like that better.

lpsinger (approved):

PR #1108, opened 2023-07-12
Preserve circulars search params when returning from new circular form

### LINKED ISSUE #1105: Going back from circulars/new does not preserve url search parameters
When navigating back to the circulars search page from an instance of circulars/new (either through the on-page back button or by declining the sign in prompt), the url search parameters are not preserved. This is essentially the same issue as #923

lpsinger:
Fixes #1105.

### DIFF
--- a/app/lib/utils.ts
+++ b/app/lib/utils.ts
@@ -5,6 +5,7 @@
  *
  * SPDX-License-Identifier: NASA-1.3
  */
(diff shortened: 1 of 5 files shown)


tylerbarna (approved):
useSearchString is an improvement over the previous searmParamsString method and is more generally usable for instances where we want a UI element that navigates to the previous page to preserve said previous page's url parameters.

PR #1130, opened 2023-07-18
Schema Browser Updates: Routes as Schema

dakota002:


dakota002:
@lpsinger I added back some UI for selecting the version:

Closed:
image

Open:
image

What are your thoughts?

lpsinger:
> What are your thoughts?

I like it, except that this too feels like an application for the rich dropdown control that @tylerbarna needs.

dakota002:
Do they have a component for that, or need one? I could pop this out into a reusable component. It's mostly just a [details](https://www.w3schools.com/tags/tag_details.asp) tag with some css

tylerbarna:
> Do they have a component for that, or need one? I could pop this out into a reusable component. It's mostly just a [details](https://www.w3schools.com/tags/tag_details.asp) tag with some css

yeah, this would pretty much address the needs for the popup menu, provided it's possible to set its z-index to display over other page elements rather than displacing them and also that it can accept child components

dakota002:
I'll break it out as a separate PR

dakota002:
The component PR: https://github.com/nasa-gcn/gcn.nasa.gov/pull/1147

lpsinger:
Would you please give this PR a more descriptive name?

app/routes/docs/schema-browser/latest.$.ts
```diff
@@ -0,0 +1,19 @@
...
+  owner: 'nasa-gcn',
+  repo: 'gcn-schema',
+}
+const octokit = new Octokit()
+
+export async function loader({ params: { '*': path } }: DataFunctionArgs) {
+  let latestRelease
+  try {
+    latestRelease = (await octokit.rest.repos.getLatestRelease(githubData)).data
+  } catch (error) {
```
lpsinger, line 15: What exception does it raise if there have been no releases? Can we make this `catch` block more specific?

app/routes/docs/schema-browser/latest.$.ts
```diff
@@ -0,0 +1,19 @@
+import { Octokit } from '@octokit/rest'
+import type { DataFunctionArgs } from '@remix-run/node'
+import { redirect } from '@remix-run/node'
+
+const githubData = {
+  owner: 'nasa-gcn',
+  repo: 'gcn-schema',
+}
+const octokit = new Octokit()
```
lpsinger, line 9: Reuse the existing octokit client instance in schema-data.ts.

That may mean defining a function in schema-data.ts that gets the latest release.

app/routes/docs.tsx
```diff
@@ -103,7 +103,10 @@ export default function () {
                 </>,
                 ...(enableSchemaBrowser
                   ? [
-                      <NavLink key="schema-browser" to="schema-browser">
+                      <NavLink
+                        key="schema-browser"
+                        to="schema-browser/main/gcn"
```
lpsinger, line 108: This should be a link to the latest release, not the main branch.

app/routes/docs/schema-browser/$version/$.tsx
```diff
@@ -1,34 +1,164 @@
...
+} from '~/lib/schema-data'
 
-  return { path, result, examples }
+export async function loader({
+  params: { version, '*': path },
+}: DataFunctionArgs) {
+  if (!version) throw new Response(null, { status: 404 })
+  if (path?.endsWith('/')) {
+    return redirect(`${path.slice(0, -1)}`)
+  }
```
lpsinger, line 41: Why is this necessary?

lpsinger (changes requested):
Move the call to `getVersionRefs()` and the version dropdown component one route module higher so that we don't need to query GitHub for the version refs on every single request.

app/routes/docs/schema-browser/$version/$.tsx
dakota002: Ah maybe not totally needed, there were errors when the path ended with '/' because of how it tried to get the content data from the api. We could just let it throw the error

lpsinger: I guess it's fine.

app/routes/docs/schema-browser/latest.$.ts
lpsinger: I suggest catching 404 errors but re-raising anything else.

dakota002: Is there a correct way to implement that with like ` if (error instanceof ???)`  

Or some other way to catch the 404?

lpsinger: Something like that, yes.

dakota002: Any suggestions? Or examples? There is no class exported by octokit that allows for the instanceof method

lpsinger: https://github.com/octokit/octokit.js#request-error-handling

dakota002: Perfect, thanks!

dakota002:
I missed the thing about moving the version selector, working on that now

dakota002:
Its ready for review

app/routes/docs/schema-browser.tsx
```diff
@@ -5,127 +5,75 @@
...
+import { getVersionRefs } from '~/lib/schema-data'
 
-// Schema treeItem
-type SchemaTreeItem = {
-  name: string
-  path: string
-  children?: SchemaTreeItem[]
-}
+export async function loader({
+  params: { version, '*': path },
```
lpsinger, line 20: How will this route ever have params? The route module name does not contain any splats.

lpsinger (changes requested):

app/routes/docs/schema-browser.tsx
lpsinger: Interesting.

lpsinger (approved):

PR #1131, opened 2023-07-18
Update to Remix 1.19.0

lpsinger:
Remix 1.19.0 is in pre-releases right now, but the DX is so much nicer that folks might want to try it early!


lpsinger:
Remix 1.19.0 is out! This is now ready for review.

dakota002 (approved):

Courey (approved):

PR #1132, opened 2023-07-18
Fix several mobile nav visibility bugs

lpsinger:
- Collapse the mobile nav on clicking any link in it.
- Collapse the mobile nav on clicking the overlay.
- The overlay still does not yet automatically hide when the browser size changes, but at least you can now click it away.

### DIFF
--- a/app/components/Header.tsx
+++ b/app/components/Header.tsx
@@ -68,11 +68,25 @@ export function Header() {
   const email = useEmail()
   const [expanded, setExpanded] = useState(false)
   const [userMenuIsOpen, setUserMenuIsOpen] = useState(false)
-  const onClick = () => setExpanded((prvExpanded) => !prvExpanded)
+
+  function toggleMobileNav() {
+    setExpanded((expanded) => !expanded)
+  }
+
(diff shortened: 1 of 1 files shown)


app/components/Header.tsx
```diff
@@ -68,11 +68,25 @@ export function Header() {
   const email = useEmail()
   const [expanded, setExpanded] = useState(false)
   const [userMenuIsOpen, setUserMenuIsOpen] = useState(false)
-  const onClick = () => setExpanded((prvExpanded) => !prvExpanded)
+
+  function toggleMobileNav() {
+    setExpanded((expanded) => !expanded)
```
Courey, line 73: just so I understand, why does does this have to be done this way instead of just !expanded?

Courey (approved):

app/components/Header.tsx
lpsinger: It probably doesn't make a difference here, but it makes a difference if you [call setState multiple times in the same evaluation of the functional component](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state).

PR #1141, opened 2023-07-19
Fix glob for esbuild entrypoints

lpsinger:
I failed to update this when rebasing #1131.


lpsinger:
Just going ahead and merging this, since it's broken on main.

PR #1142, opened 2023-07-19
Relocate new circular button

tylerbarna:
Moves the new circular button to the upper right, inline with the title of the circulars page, to be more immediately visible.

Current Position:
image
image

Relocated:
image
image

The label for the circular is also slightly more clear in its labelling.

This will also simplify adding new search parameters to the search, as only search-related elements will be inline with the search bar


lpsinger (commented):
This is a nice alternative. I'm hesitant to use it until we really know that we don't have the space in the existing combined search and button bar, but we have it if we need it.

PR #1147, opened 2023-07-21
Component version of a details and summary with a dropdown-like display

dakota002:
Here is the example:
image

The summary and children props are both ReactNodes so you should be able to put pretty much whatever you want inside them


lpsinger:
Neat!

What do we think about this style vs. the tooltip style?

dakota002:
My vote is for this way, its more flexible than the existing Tooltip component in that it allows for more than just a string as the label. I think this design would also be a better UX, since tooltips tend to disappear quickly once the mouse is moved away.

@tylerbarna will this meet your needs for what you are working on?

tylerbarna:
> I think this design would also be a better UX, since tooltips tend to disappear quickly once the mouse is moved away.

Yeah, I think having something that toggles its presence on clicking rather than on hovering is more along the lines of what we'd want. If you click on another part of the page, does the dropdown element disappear?

lpsinger:
> My vote is for this way, its more flexible than the existing Tooltip component in that it allows for more than just a string as the label. I think this design would also be a better UX, since tooltips tend to disappear quickly once the mouse is moved away.

That's not what I meant. I meant using the visual appearance of the USWDS tooltip component but with on-click rather than on-hover behavior. And of course, supporting an arbitrary component to trigger the display.

lpsinger:
> This is a rough idea of how I imagined the date range search frontend (pr #1101) would look after speaking with Leo (apologies for the penmanship)
> 
> image
> where the element that's expanded after selecting the custom date option is the [react-uswds date range picker](https://designsystem.digital.gov/components/date-range-picker/) and a button for submitting

That's great, but would you please post that on #1101 instead?

dakota002:
> > My vote is for this way, its more flexible than the existing Tooltip component in that it allows for more than just a string as the label. I think this design would also be a better UX, since tooltips tend to disappear quickly once the mouse is moved away.
> 
> That's not what I meant. I meant using the visual appearance of the USWDS tooltip component but with on-click rather than on-hover behavior. And of course, supporting an arbitrary component to trigger the display.

Ah I see, I can play around with the styling to make it look like that. Does that entail anything more than the pointy bit and the black background?
image

lpsinger:
> Ah I see, I can play around with the styling to make it look like that. Does that entail anything more than the pointy bit and the black background?

No, that was my idea. Can we reuse the existing `usa-tooltip` CSS class?

dakota002:
Looks like the usa-tooltip class only applies the following:
```
.usa-tooltip {
  display: inline-block;
  position: relative;
}
```
but i can apply some custom css to get the intended effects

lpsinger:
> Looks like the usa-tooltip class only applies the following:
> 
> ```
> .usa-tooltip {
>   display: inline-block;
>   position: relative;
> }
> ```
> 
> but i can apply some custom css to get the intended effects

Perhaps you are looking for `usa-tooltip__body`. See https://github.com/uswds/uswds/blob/develop/packages/usa-tooltip/src/styles/_usa-tooltip.scss.

Try to do this _without_ custom CSS.

dakota002:
The tooltip body class blocks mouse events within the content, which makes links unclickable. I just created a custom css class, and copied the styling from the body then removed bits until it started functioning as expected. 

So far I have this as my test case, but i think looks a bit intense with the black background.
image

What are your thoughts on it?

lpsinger:
I agree. The card based style is nicer.

lpsinger (changes requested):
- [x] Unfortunately, I don't think that `<details>`/`<summary>` is a good fit here because there is a lot of variation between browsers on whether or not a built-in disclosure icon is shown, and [there isn't widespread support for customizing that appearance](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget). See below for examples of how this looks on Safari versus Firefox.
- [x] A lot of the CSS classes here only make sense in the specific context of the tree view, and the width doesn't make sense if it is used on any ordinary page. Please pare back the CSS classes so that this control is reusable.
- [x] Consider splitting this into two controls: one for the menu button and another for the contents.

Screenshot 2023-07-21 at 18 46 08
Screenshot 2023-07-21 at 18 46 15

dakota002:
I trimmed it back to be fairly generic, but I think it should just be one component.  There are really only two components that make up this one, so splitting them out seems like it defeats the purpose of having the component at all

lpsinger:
> There are really only two components that make up this one, so splitting them out seems like it defeats the purpose of having the component at all

There are only two components inside of this one, but combining them into one component, you lose the ability to apply styles to each of those components.

Once you add all of the accessibility attributes that you need, neither control will be all that simple.

app/components/DetailsDropdown.tsx
```diff
@@ -0,0 +1,33 @@
...
+  summary: ReactNode
+  children: ReactNode
+  className?: string
+}) {
+  const [showContent, setShowContent] = useState(false)
+
+  return (
+    <div className={className}>
+      <Button
+        className="display-flex flex-align-center usa-button usa-button--outline"
```
lpsinger, line 19: The flex properties and the outline style may be specific to the version dropdown application.

```suggestion
        className="usa-button"
```

lpsinger (changes requested):
Add Aria roles and accessibility attributes related to expanding and collapsing content.

app/components/DetailsDropdownControl.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Button
+      className={`usa-button ${className}`}
+      type="button"
+      onClick={onClick}
+      aria-controls="details-content"
+      role="button"
+      {...props}
+    >
+      <span className="margin-right-auto">{children}</span>
```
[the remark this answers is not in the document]
dakota002: Ah nothing now. I had previously been using it along with some other utilities to do spacing between the wording and the icon. I'll remove it in the next commit

app/components/DetailsDropdownButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+}: {
+  children: ReactNode
+  className?: string
+} & Omit<ButtonProps & JSX.IntrinsicElements['button'], 'type'>) {
+  return (
+    <Button
+      className={`usa-button ${className ?? ''}`}
+      type="button"
+      aria-controls="details-content"
+      role="button"
```
lpsinger, line 18: Isn't this role implied by `type="button"`?

app/components/DetailsDropdownButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children,
+  className,
+  ...props
+}: {
+  children: ReactNode
+  className?: string
+} & Omit<ButtonProps & JSX.IntrinsicElements['button'], 'type'>) {
+  return (
+    <Button
+      className={`usa-button ${className ?? ''}`}
```
lpsinger, line 15: Isn't `usa-button` already applied by the `<Button>` component?

lpsinger (changes requested):

app/components/DetailsDropdownButton.tsx
dakota002: Apparently so. I have not used the ARIA stuff that much, didn't realize it did that

app/components/DetailsDropdownButton.tsx
dakota002: Yup, that was another accidental holdover from the details/summary version. Fixing now

app/components/DetailsDropdownButton.tsx
```diff
@@ -0,0 +1,24 @@
+import { Button, Icon } from '@trussworks/react-uswds'
```
lpsinger, line 1: Add license header

app/components/DetailsDropdownContent.tsx
```diff
@@ -0,0 +1,19 @@
+import { Card, CardGroup } from '@trussworks/react-uswds'
```
lpsinger, line 1: Add license header

app/components/DetailsDropdownButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  ...props
+}: {
+  children: ReactNode
+  className?: string
+} & Omit<ButtonProps & JSX.IntrinsicElements['button'], 'type'>) {
+  return (
+    <Button
+      className={`${className ?? ''}`}
+      type="button"
+      aria-controls="details-content"
```
lpsinger, line 17: id property still hardcoded!

lpsinger (changes requested):

app/components/DetailsDropdownContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+import type { ReactNode } from 'react'
+
+export default function DetailsDropdownContent({
+  children,
+  className,
+}: {
+  children: ReactNode
+  className?: string
+}) {
```
lpsinger, line 17: Add `...props` and pass through to the `<CardGroup>`.

lpsinger (changes requested):

app/components/DetailsDropdownContent.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type { ReactNode } from 'react'
+
+export default function DetailsDropdownContent({
+  children,
+  className,
+  ...props
+}: {
+  children: ReactNode
+  className?: string
+}) {
```
lpsinger, line 18:
```suggestion
  children,
  className,
  ...props
}: : Omit<JSX.IntrinsicElements['ul'], 'role'>) {
```

lpsinger (changes requested):

lpsinger (approved):

PR #1171, opened 2023-07-31
Circulars Search UI Fix 

dakota002:
Moves search form styling off of form to prevent unintended propagation of styles to all form elements

This will be relevant for the DetailsDropdown updates to follow


lpsinger (commented):
Alright, I see that all this is doing is changing this:

```jsx
<Form className="foobar">
  ...form contents...
</Form>
```

to this:

```jsx
<Form>
  <div className="foobar">
    ...form contents...
  </div>
</Form>
```

What does that get us?

dakota002:
Its subtle, but it fixes a couple of the UI issues that occur when we start adding more buttons inside the form

PR #1172, opened 2023-07-31
Fuzzy date range backend

tylerbarna:
added backend for fuzzy date range. This is implemented by adding a new url parameter, called 'last' to relevant functions. if last is provided without start or end date parameters, there is a call to the new fuzzyTimeRange function. There are several strings that can be supplied to the function:

- hour
- today (starting at 00:00 GMT on current day)
- day
- week
- month
- year
- mtd (starting at the first day of the current month)
- ytd (starting at first day of current year)

If start or end date is specified, the 'last' url parameter is overridden. In the frontend, switching from a manual date range to a fuzzy date range can be implemented by having the onClick of the fuzzy dates clear the start and end dates


lpsinger (changes requested):
The "last" attribute seems unrelated to the fuzzy date range. Please remove that to a separate PR.

tylerbarna:
> The "last" attribute seems unrelated to the fuzzy date range. Please remove that to a separate PR.

it's how the fuzzy date range is implemented; the 'last' parameter is the string that's used to determine which fuzzy date range to use (see the fuzzyTimeRange function in circulars.server.ts)

lpsinger:
> > The "last" attribute seems unrelated to the fuzzy date range. Please remove that to a separate PR.
> 
> it's how the fuzzy date range is implemented; the 'last' parameter is the string that's used to determine which fuzzy date range to use (see the fuzzyTimeRange function in circulars.server.ts)

Oh, I see.

What I would like to have both the `startDate` and `endDate` parameters take values like "2022-08-27", "now", "1 week ago", "2 weeks ago", etc.

app/routes/circulars/circulars.server.ts
```diff
@@ -129,7 +158,7 @@ export async function search({
     }) => ({
       circularId,
       subject,
-    })
+    }),
```
tylerbarna, line 161:
```suggestion
    })
```

app/routes/circulars/circulars.server.ts
```diff
@@ -217,7 +246,7 @@ export async function circularRedirect(query: string) {
   const validCircularSearchStyles =
     /^\s*(?:GCN)?\s*(?:CIRCULAR)?\s*(-?\d+(?:\.\d)?)\s*$/i
   const circularId = parseFloat(
-    validCircularSearchStyles.exec(query)?.[1] || ''
+    validCircularSearchStyles.exec(query)?.[1] || '',
```
tylerbarna, line 249:
```suggestion
    validCircularSearchStyles.exec(query)?.[1] || ''
```

tylerbarna:
> > > The "last" attribute seems unrelated to the fuzzy date range. Please remove that to a separate PR.
> > 
> > 
> > it's how the fuzzy date range is implemented; the 'last' parameter is the string that's used to determine which fuzzy date range to use (see the fuzzyTimeRange function in circulars.server.ts)
> 
> Oh, I see.
> 
> What I would like to have both the `startDate` and `endDate` parameters take values like "2022-08-27", "now", "1 week ago", "2 weeks ago", etc.

Are there fuzzy dates where we would want to define the endDate as something other than the present? My idea for the fuzzy dates was that it would stretch from the current date to some amount of time in the past, whereas the manual date range would be used to specify a unique end date.

Because of the way the date range selector works in react-uswds, it provides a format slightly different from "2022-08-27", and I've found that trying to change that format messes up the behaviour of the date filtering when returning to the page after navigating away due to the way it's encoded in the url parameter. This may also be true of it trying to accept strings like "2 weeks ago".

lpsinger:
> Are there fuzzy dates where we would want to define the endDate as something other than the present?

What do you think, @Vidushi-GitHub, @Courey?

lpsinger:
> What I can do is add a regex expression to look for the 'MM/DD/YYYY' format and, if it does not adhere to that, then submit the string to the fuzzy time range

I would like to use the format YYYY-MM-DD rather than MM/DD/YYYY.

tylerbarna:
> > What I can do is add a regex expression to look for the 'MM/DD/YYYY' format and, if it does not adhere to that, then submit the string to the fuzzy time range
> 
> I would like to use the format YYYY-MM-DD rather than MM/DD/YYYY.

I had tried that before; while this is fine if you manually input the date in YYYY-MM-DD, the issue is that using the calendar selector in the date range selector automatically uses MM/DD/YYYY. The date range picker also won't show a date in the YYYY-MM-DD format in the input field if one links to a search that includes that format, though the backend handles the YYYY-MM-DD format fine.

tylerbarna:
weirdly enough, the backend of the date range picker uses YYYY-MM-DD for defining minimum and maximum dates, though I haven't been able to find any means of changing its user-facing format 

https://designsystem.digital.gov/components/date-picker/

lpsinger:
> weirdly enough, the backend of the date range picker uses YYYY-MM-DD for defining minimum and maximum dates, though I haven't been able to find any means of changing its user-facing format

I challenge you to make a PR for https://github.com/trussworks/react-uswds to set the output format!

Vidushi-GitHub:
> > What I can do is add a regex expression to look for the 'MM/DD/YYYY' format and, if it does not adhere to that, then submit the string to the fuzzy time range
> 
> I would like to use the format YYYY-MM-DD rather than MM/DD/YYYY.

(a) Are there fuzzy dates where we would want to define the endDate as something other than the present?
- I think it's good to set default endDate as present, for Fuzzy dates "now" or "a week ago"

As our users are astronomers, I would like to see UTC pattern for date. Searching with start and end date as YYYY-MM-DD is more general method and consistent with Circulars publish date.

tylerbarna:
> > > What I can do is add a regex expression to look for the 'MM/DD/YYYY' format and, if it does not adhere to that, then submit the string to the fuzzy time range
> > 
> > 
> > I would like to use the format YYYY-MM-DD rather than MM/DD/YYYY.
> 
> (a) Are there fuzzy dates where we would want to define the endDate as something other than the present?
> 
> * I will also go with setting the default endDate to the "present" for fuzzy dates like "now" or "a week ago."
> 
> (b) Since our users are astronomers, I would like to see the UTC pattern for the date. Searching with start and end date as YYYY-MM-DD is a more general method and consistent with Circulars publish date. Even I was willing to ask if it's possible to set start and end date as UTC, gbm, heasarc use that?
> 
> (c) This search is based on publication date of Circulars, may be it's obvious or may be mention with date range button.

addressed point (a) with most recent commit

app/routes/circulars/circulars.server.ts
```diff
@@ -52,6 +52,40 @@ function parseDate(date?: string) {
...
+      new Date().getFullYear(),
+      new Date().getMonth(),
+      1
+    ).getTime()
+  if (fuzzyTime === 'ytd')
+    return new Date(new Date().getFullYear(), 0).getTime()
+  else return undefined // invalid fuzzyTime defaults to fuzzless time range
+}
+
+function checkFuzzyDate(startDate?: string, endDate?: string) {
```
Courey, line 76: this is super nit picky, so feel free to ignore. The checkFuzzyDate function is doing more than just checking for a fuzzy date. It's validating and using the fuzzy date sometimes. I would name it something like getValidDates.

app/routes/circulars/index.tsx
```diff
@@ -210,6 +210,7 @@ export default function () {
         <Link to="/docs/circulars">documentation</Link> for help with
         subscribing to or submitting Circulars.
       </p>
+      <h2> number of results: {totalItems}</h2>
```
tylerbarna, line 213:
```suggestion
```

app/routes/circulars/circulars.server.ts
```diff
@@ -52,6 +52,40 @@ function parseDate(date?: string) {
...
+  if (fuzzyTime === 'month') return now - 86400000 * 30
+  if (fuzzyTime === 'year') return now - 86400000 * 365
+  if (fuzzyTime === 'mtd')
+    return new Date(
+      new Date().getFullYear(),
+      new Date().getMonth(),
+      1
+    ).getTime()
+  if (fuzzyTime === 'ytd')
+    return new Date(new Date().getFullYear(), 0).getTime()
```
lpsinger, line 72: Rewrite as a `switch` block.

app/routes/circulars/circulars.server.ts
```diff
@@ -52,6 +52,40 @@ function parseDate(date?: string) {
...
+  else return undefined // invalid fuzzyTime defaults to fuzzless time range
```
lpsinger, line 73: Not necessary. If the function never executes a `return` statement, then the return value is undefined.

```suggestion
```

app/routes/circulars/circulars.server.ts
```diff
@@ -52,6 +52,40 @@ function parseDate(date?: string) {
...
+  // regex for YYYY-MM-DD and another for MM/DD/YYYY
+  const usDateRegex = /(\d{4})-(\d{2})-(\d{2})|(\d{2})\/(\d{2})\/(\d{4})/
+  const normalDateRegex = /(\d{4})-(\d{2})-(\d{2})/
+  const usDateMatch = startDate?.match(usDateRegex)
+  const normalDateMatch = startDate?.match(normalDateRegex)
+  if (usDateMatch || normalDateMatch) {
+    return [parseDate(startDate), parseDate(endDate) + 86400000]
+  } else {
+    return [fuzzyTimeRange(startDate), fuzzyTimeRange(endDate)]
+  }
```
lpsinger, line 86: There's no need for these regular expressions. You can first call `fuzzyTimeRange`, and parse the dates if its return value is undefined.

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
tylerbarna: ah, good point. Addressed in [8f2e6ef](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1172/commits/8f2e6efbdd8af3b4492be45cdc7de9b38276185b)

app/routes/circulars/circulars.server.ts
tylerbarna: changed to a switch block in 28caefc

lpsinger (approved):

PR #1183, opened 2023-08-02
Adds docs pages for notices

dakota002:


dakota002:
@jracusin do you have any suggestions for expanding these? I think notices could probably use a general About page similar to circulars and the New Notice Producers docs might make sense to move under Notices

jracusin:
Sorry, I actually have a branch where I've been working on something similar. I should have submitted it as a draft PR a couple of weeks ago.  I'll do that and we should merge this one with what I was doing.

jracusin:
Superseded by #1196.

PR #1190, opened 2023-08-03
Fixes the 'Illegal characters in path' error when running npm run clean

dakota002:


lpsinger (approved):
Must be a Windows thing.

PR #1204, opened 2023-08-08
adding origin import

Courey:


dakota002 (approved):

lpsinger:
Hmm, why didn't TypeScript or eslint catch this?

dakota002:
@lpsinger had the same thoughts, looks like it may have been interpreted as a global origin variable

PR #1229, opened 2023-08-14
Shortens Schema-Browser to Schema

dakota002:


dakota002:
This was for demo purpose, please ignore

Issue #1202, opened 2023-08-07
Remove "Submitted By" from circulars archive


jracusin:
useful for our own internal tracking, not needed for public archive

lpsinger:
Is there any reason _not_ to make this information public?

My model for this feature had been GitHub Issues, which as far as I recall do annotate visibly which comments were submitted by email.

jracusin:
I feel like it's a little weird to display how people submit them, as it felt like publicly shaming people for using the legacy email, but maybe that'll encourage them to use the new system.  I suppose there's no harm.  Okay, I'll close this.  Also, I'm happy to see that many users are using the web form.

PR #1241, opened 2023-08-16
Remove 'submitted by' display

### LINKED ISSUE #1240: GCN Circulars: Remove submitted by
Just a quick fix, we had a discussion over removing the submitted by (legacy/web-form etc.) in the circulars, as we have inclusive policy.
But it would be nice to keep the information as a hidden file.

dakota002:
resolves #1240

### DIFF
--- a/app/routes/circulars.$circularId.tsx
+++ b/app/routes/circulars.$circularId.tsx
@@ -44,14 +44,8 @@ export async function loader({ params: { circularId } }: DataFunctionArgs) {
 export const headers: HeadersFunction = ({ loaderHeaders }) =>
   pickHeaders(loaderHeaders, ['Link'])
 
-const submittedHowMap = {
-  web: 'Web form',
-  email: 'email',
-  'email-legacy': 'legacy email',
-}
-
 export default function () {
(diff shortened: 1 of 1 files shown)


Vidushi-GitHub:
Thanks! Instead of removing totally, is it possible to store this info in some file? This information would be helpful for us at least for a year, to know how many users are still submitting to capella2, comfort with web-form etc.

dakota002:
Yep! That should still happen. I only removed the display of the data

lpsinger:
As I said in #1202, I see no reason for this information not to be public. In fact, I was thinking of adding some public statistics derived from this data.

Vidushi-GitHub:
Ok, I see. Sorry my bad, I didn't follow. I have two reasons for my concern: (a) It feels like hacking someone's alert, especially when opening the JSON file submitted by doesn't fit. (b) Users should feel at ease for quickly sending the GCN, without the sense that their submission process is publicly known, and should switch to a specific method. If our intention is to encourage the use of the web form, then it's fine.
I agree it's good to show some statistics but I feel it should be done without identification of the specific users involved.

PR #1252, opened 2023-08-20
Add GCN Circulars to sitemap

lpsinger:
Note, most of the added code in this PR has to do with paginating the sitemap due to a limit of 50,000 URLs per sitemap file.


lpsinger:
Replaced by #1428.

Issue #1110, opened 2023-07-13
clean variable in circulars index causing issues with displaying results


tylerbarna:
The use of the clean constant in index.tsx for the circulars page is causing an issue with other form elements that are being added which causes the results to not be displayed. This is because clean is evaluated as false in cases where it should be evaluating as true. Manually setting clean to be true resolves the issue, but I'm unsure how to modify the current implementation of the clean constant to both preserve the intent of the clean constant and also prevent this issue

This behaviour can be replicated in the current public site by searching a query, submitting it, and, after loading, clicking on the circulars navigation link. The results will no longer appear, though the search form will still be populated (there will be no url parameter string, and printing it to console will show it's not defined, so it should just display the full results list). I don't believe the search form being populated is the underlying issue but rather the clean constant being evaluated as false, as mentioned above.

It can also be replicated by running PR #1104 locally and changing the number of results to display with or without submitting a query, though the results will load correctly if one refreshes the browser page without altering the url query parameters

nulfrost:
Hey there 👋🏿 may I take a stab at this? So I believe I understand what's going on.

- Before you start typing in anything, inputQuery and query are undefined so the 'clean' variable will be true
- When you start typing in the search input, inputQuery is no longer undefined and is a string. string != undefined so clean is false. This is why the search results disappear as well.
- Hitting enter submits the form and so now inputQuery and query are true since they now have the same value.
- Clicking the circulars link in the navbar clears the query since you're navigating away from "?query=thing" to "/circulars", so now query is undefined and "clean" will evaluate to false. So no search results.

A few questions about the desired UX.

1. Does the query need to be updated on every keystroke? Or is just submitting the form after the user hits enter or clicks the search icon fine? I feel like the latter would be better UX
2. Should the text in the input be retained or be cleared whenever someone searches for something after the initial search?
3. Similar question for the limit search param, should that be retained in the URL when someone does a new search?

PR #1265, opened 2023-08-21
Circulars Search Date Range Frontend

tylerbarna:
Fresh PR, as the old PR (#1101) was from prior to some updates that made merging difficult. This includes all the frontend changes made in that PR, along with the associated existing issues, namely the anchoring of the details dropdown (see #1171 for fix-in-progress) as well as an issue where selecting a specific fuzzy date range doesn't properly update the url parameters - rather, the fuzzy date range lags by one (ie, if I select option A, the page will reload without applying option A. If I then proceed to select option B, option A will now be present in the url)

### DIFF
--- a/app/routes/circulars._index.tsx
+++ b/app/routes/circulars._index.tsx
@@ -233,167 +233,168 @@ export default function () {
         subscribing to or submitting Circulars.
       </p>
       <ButtonGroup className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1">
-        <Form
-          className="display-inline-block usa-search usa-search--small"
-          role="search"
-          id="searchForm"
-        >
-          <Label srOnly={true} htmlFor="query">
-            Search
-          </Label>
-          <TextInput
-            autoFocus
-            id="query"
-            name="query"
-            type="search"
-            defaultValue={inputQuery}
-            placeholder="Search"
-            aria-describedby="searchHint"
-            onChange={({ target: { form, value } }) => {
-              setInputQuery(value)
-              if (!value) submit(form)
-            }}
-          />
-          <Button type="submit">
-            <img
-              src={searchImg}
-              className="usa-search__submit-icon"
-              alt="Search"
-            />
-          </Button>
-          <DetailsDropdownButton
-            className="height-4 padding-top-0 padding-bottom-0"
-            onClick={() => setShowContent(!showContent)}
-            aria-expanded={showContent}
-            aria-controls="searchFilters"
+        <Form id="searchForm">
+          <div
+            className="display-inline-block usa-search usa-search--small"
+            role="search"
           >
-            <Icon.CalendarToday />
-          </DetailsDropdownButton>
-          {showContent && (
-            <DetailsDropdownContent>
-              {!showDateRange && (
-                <>
-                  <Button
-                    type="button"
-                    className="usa-button usa-button--unstyled"
-                    onClick={(value) => {
-                      setInputDateGte('hour')
-                      setInputDateLte('now')
-                      setDateRange()
-                    }}
-                  >
-                    Last Hour
-                  </Button>
-                  <Button
-                    type="button"
-                    className="usa-button usa-button--unstyled"
-                    onClick={(value) => {
-                      setInputDateGte('today')
-                      setInputDateLte('now')
-                      setDateRange()
-                    }}
-                  >
-                    Today
-                  </Button>
-                  <Button
-                    type="button"
-                    className="usa-button usa-button--unstyled"
-                    onClick={(value) => {
-                      setInputDateGte('day')
-                      setInputDateLte('now')
-                      setDateRange()
-                    }}
-                  >
-                    Last 24 Hours
-                  </Button>
-                  <Button
-                    type="button"
-                    className="usa-button usa-button--unstyled"
-                    onClick={(value) => {
-                      setInputDateGte('month')
-                      setInputDateLte('now')
-                      setDateRange()
-                    }}
-                  >
-                    Last Month
-                  </Button>
-                  <Button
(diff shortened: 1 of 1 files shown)


tylerbarna:
reposting the UI concept from the previous PR
image

tylerbarna:
This is also dependent on https://github.com/trussworks/react-uswds/issues/2537

lpsinger:
@tylerbarna, would you please post a few screenshots?

tylerbarna:
> @tylerbarna, would you please post a few screenshots?

sure:

image
image
image

I think the UI of the dropdown buttons could use some work, and I'd also prefer if the left side of the dropdown was aligned with the dropdown button, but that seems to potentially be a limitation of the buttonGroup

lpsinger (changes requested):
- [ ] **Rely on passive HTML form data behavior rather than dynamically constructed form data whenever possible.** Instead of actively, programmatically constructing the date range params for the form submission from React states, give the fuzzy date range inputs names and values that result in appropriate form data. They can be either radio buttons or submit buttons; in either case, you can have multiple buttons with the same name but different values, and the button that is clicked is the one whose value is present in the form data.
- [ ] **Use the default appearances of react-uswds components whenever possible.** The USWDS components are intended to have typography and (generous) spacing that provides a consistent look and feel. Use CSS utility classes and custom styles _sparingly_. This will also fix the issue that the design looks quite cramped due to too little whitespace.
- [ ] Display the current date range limits (if any are set) in the dropdown button in its collapsed form.

lpsinger:
@tylerbarna, any updates on this?

tylerbarna:
> @tylerbarna, any updates on this?

Hey Leo, my parents were visiting last week, so I'm getting caught up; will be working on addressing your comments today

tylerbarna:
May need to tweak the styling further, but I reverted the buttons to their standard styling 
image
image

Also added a basic implementation of the dynamic button label, which definitely needs refinement.
image
image
image

tylerbarna:
> * [ ]  **Rely on passive HTML form data behavior rather than dynamically constructed form data whenever possible.** Instead of actively, programmatically constructing the date range params for the form submission from React states, give the fuzzy date range inputs names and values that result in appropriate form data. They can be either radio buttons or submit buttons; in either case, you can have multiple buttons with the same name but different values, and the button that is clicked is the one whose value is present in the form data.

Could you expand on this? Would we no longer be utilizing a react state to store/read the start and end date parameters in the url? @dakota002  and I spent some time in a huddle earlier brainstorming how to potentially alter the repetitive onClick behaviour for the fuzzy date ranges, but we were encountering some trouble when trying to make alterations such that the following conditions were satisfied:
1. The fuzzy times and user-defined dates use the same url parameters
2. The fuzzy times were human readable in the url 
3. The fuzzy times were dynamic (ie if, in February, I copy the url of a search that specifies only results from the past month, navigating to that same url in September will return results from August, not January)

lpsinger:
@tylerbarna, you raise some good points that I hadn't considered. I am thinking about this.

tylerbarna:
@lpsinger I am tinkering with using radio buttons; after spending some time working on it with Courey, we were able to address that parameter lagging by one submission issue by setting the parameter by selecting the item and then having a separate "submit" button, which is a design choice I would think compliments using radio buttons. I still need to see how the radio buttons work with the custom date ranges, however. Is [this](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1265/commits/72b7565e149579fe03e0c3c55fa5fe518d20ae35) along the lines of what you were thinking for the radio buttons?

lpsinger:
> Is [this](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1265/commits/72b7565e149579fe03e0c3c55fa5fe518d20ae35) along the lines of what you were thinking for the radio buttons?

I thought that the radio buttons would have been a way to avoid needing a React state for the fuzzy date, so no, not really. Don't worry about that too much for now. I'll take a fresh look at this PR at my next opportunity.

However, from the screen shots I still see that the button content is extremely compressed and crowded. Please use the default spacing for button groups as much as possible. We aim to use default USWDS styling as much as possible so that whitespace, fonts, and layout is consistent across the entire UI.

tylerbarna:
> However, from the screen shots I still see that the button content is extremely compressed and crowded. Please use the default spacing for button groups as much as possible. We aim to use default USWDS styling as much as possible so that whitespace, fonts, and layout is consistent across the entire UI.

Yeah, I've been trying to figure out how to have the width of the details dropdown button be dynamic to prevent it from feeling so crowded, but the default behaviour seems to be overflowing out of view rather than widening the button

lpsinger:
> > However, from the screen shots I still see that the button content is extremely compressed and crowded. Please use the default spacing for button groups as much as possible. We aim to use default USWDS styling as much as possible so that whitespace, fonts, and layout is consistent across the entire UI.
> 
> Yeah, I've been trying to figure out how to have the width of the details dropdown button be dynamic to prevent it from feeling so crowded, but the default behaviour seems to be overflowing out of view rather than widening the button

@Courey, I think that @tylerbarna could use some CSS help here.

tylerbarna:
@lpsinger given the addition of other advanced search options, do you think it might be better to move the date range filter to also be in an advanced search filters tab?

image

image

image

tylerbarna:
@lpsinger added commits that revert to a dropdown menu but using the radio buttons; they still need to be realigned or have the dropdown menu width altered, and the issue with the button label overflowing persists. #1110 and #1366 are also still present

tylerbarna:
Closed due to staleness, replaced by #1482

tylerbarna:
Closed due to staleness, replaced by https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482

PR #1268, opened 2023-08-21
Fix link, remove .schema.json from display name, pull correct example files

dakota002:


dakota002:
Folders UI
image

Mobile:
image
image


Schema UI
image

Mobile:
image

app/routes/docs_.schema.tsx
```diff
@@ -73,6 +76,9 @@ export default function Schema() {
         <BreadcrumbBar className="tablet:grid-col-fill">
           <>{...breadCrumbs}</>
         </BreadcrumbBar>
+        {windowSize.width < 480 && !isSchema && (
+          <h2 className="margin-y-0">{path.split('/').slice(-1)[0]}</h2>
+        )}
```
lpsinger, line 81: Can we please show the full cookie crumb path, even on mobile?

app/routes/docs_.schema.tsx
```diff
@@ -39,26 +43,46 @@ export async function loader({
...
+            {item.replace('.schema.json', '')}
+          </h2>
+        ) : (
+          <BreadcrumbLink href={`/docs/schema/${version}/${crumbPath}`}>
+            <span>{item}</span>
+          </BreadcrumbLink>
+        )}
+      </Breadcrumb>
+    )
+  })
```
lpsinger, line 69: - [ ] I think that you can use spread syntax here so that you can avoid the comparisons.
- [ ] You might as well have this return the breadcrumb bar itself rather than a list of components, so that this is a proper React component.

```jsx
const [...breadcrumbs, lastBreadcrumb] = path.split('/')
return (
  <BreadcrumbBar>
    {
      breadcrumbs.map(
        (breadcrumb) => <Breadcrumb>fill me in for path components</Breadcrumb>
      )
    }
   <Breadcrumb>fill me in for filename</Breadcrumb>
  </BreadcrumBar>
)
```

lpsinger (changes requested):

app/routes/docs_.schema.tsx
dakota002: We do on larger sizes, but the USWDS component drops all but the previous nav point on smaller screens

app/routes/docs_.schema.tsx
dakota002: I think I got this idea working on my latest commit

app/routes/docs_.schema.$version.$/route.tsx
```diff
@@ -112,7 +85,28 @@ export default function () {
             examples={examples}
           />
         ) : (
-          <SchemaDefinition />
+          <>
+            <SchemaDefinition />
+            <CardGroup>
+              {...data.map((x: GitContentDataResponse) => (
```
lpsinger, line 91: Change `useLoaderData()` to `useLoaderData<typeof loader>()` so that this type is automatically inferred.

app/components/BreadcrumbNav.tsx
```diff
@@ -0,0 +1,40 @@
...
+  path,
+  pathPrepend,
+  className,
+}: {
+  path: string
+  pathPrepend?: string
+  className?: string
+}) {
+  const [lastBreadcrumb, ...breadcrumbs] = path.split('/').reverse()
+  const sortedCrumbs = breadcrumbs.reverse()
```
lpsinger, line 17: Ah, I didn't realize that spread syntax requires the element with ellipses to be the last one.

Here's a simpler approach:

```ts
const breadcrumbs = path.split('/')
const lastBreadcrumb = breadcrumbs.pop()
```

app/components/BreadcrumbNav.tsx
```diff
@@ -0,0 +1,40 @@
...
+            .slice(0, sortedCrumbs.indexOf(item) + 1)
+            .join('/')
+          return (
+            <Breadcrumb key={item}>
+              <BreadcrumbLink href={[pathPrepend, `${crumbPath}`].join('/')}>
+                {item}
+              </BreadcrumbLink>
+            </Breadcrumb>
+          )
+        })}
```
lpsinger, line 33: Here's a simpler approach:

```jsx
let cumulativePath = prependPath
...
return (
  <BreadcrumbBar>
    {breadcrumbs.map((breadcrumb) => {
      cumulativePath = `${cumulativePath}/${breadcrumb}`
      return (
        <Breadcrumb key={cumulativePath}>
          <BreadcrumbLink href={cumulativePath}>
            {breadcrumb}
          </BreadcrumbLink>
        </Breadcrumb>
      )
    })}
    <Breadcrumb>
      <BreadcrumbLink href={`${cumulativePath}/${lastBreadcrumb}`}>
        ${lastBreadcrumb}
      </BreadcrumbLink>
    </Breadcrumb>
  </BreadcrumbBar>
)
```

Also, note that `item` is not a valid key because it is possible for the path to contain two path components that are equal strings, hence not unique.

app/components/BreadcrumbNav.tsx
```diff
@@ -0,0 +1,40 @@
+import {
```
lpsinger, line 1: - [ ] Add missing copyright and license header.
- [ ] Move to the route module folder, since this component is only used once.

app/lib/schema-data.server.ts
```diff
@@ -27,40 +28,46 @@ function isErrnoException(e: unknown): e is NodeJS.ErrnoException {
...
+        await loadSubSchema(body.allOf, body.$id)
+      }
+      if (body.anyOf?.find((x) => x.$ref)) {
+        await loadSubSchema(body.anyOf, body.$id)
+      }
+      if (body.oneOf?.find((x) => x.$ref, body.$id)) {
+        await loadSubSchema(body.oneOf, body.$id)
+      }
+    } catch (e) {
+      if (isErrnoException(e) && e.code === 'ENOENT') {
```
lpsinger, line 61: How would the GitHub API return an OS error like this? This code is probably a remnant of when you were reading from the filesystem.

app/routes/docs_.schema.$version.$/components.tsx
```diff
@@ -36,9 +36,16 @@ export type Schema = {
   required?: string[]
 }
 
+function FormatLinkString(path: string) {
```
lpsinger, line 39:
```suggestion
function formatLinkString(path: string) {
```

Function names generally start with a lowercase letter.

lpsinger (changes requested):

app/routes/docs_.schema.$version.$/route.tsx
dakota002: This change makes the following message appear:
Parameter 'x' implicitly has an 'any' type.

app/lib/schema-data.server.ts
dakota002: Yup that sounds right, I'll remove that

app/lib/schema-data.server.ts
```diff
@@ -118,53 +118,59 @@ export type GitContentDataResponse = {
...
-      ...repoData,
-      path: path ?? 'gcn',
-      ref,
-    })
-  ).data as GitContentDataResponse[]
-}
+export const loadSchemaExamples = memoizee(
+  async function (path: string, ref: string): Promise<ExampleFiles[]> {
+    const dirPath = path.substring(0, path.lastIndexOf('/'))
+    const schemaName = path.substring(path.lastIndexOf('/') + 1)
```
lpsinger, line 124: Use `path.dirname` and `path.basename` rather than manual string processing. https://nodejs.org/api/path.html

app/lib/schema-data.server.ts
```diff
@@ -118,53 +118,59 @@ export type GitContentDataResponse = {
...
+    const exampleFiles = (await getGithubDir(dirPath, ref)).filter(
+      (x) =>
+        x.name.startsWith(`${schemaName.split('.')[0]}.`) &&
+        x.name.endsWith('.example.json')
+    )
+
+    const result: ExampleFiles[] = []
+    for (const exampleFile of exampleFiles) {
+      const exPath = join(dirPath, '/', exampleFile.name)
```
lpsinger, line 133: Use path.join instead of manual string processing. https://nodejs.org/api/path.html

app/routes/docs_.schema.$version.$/route.tsx
lpsinger: What is it inferring the type of `data` is?

lpsinger (changes requested):

app/routes/docs_.schema.$version.$/route.tsx
dakota002: It is inferring the type as `any`, could that be because the loader function returns JSON?

app/lib/schema-data.server.ts
```diff
@@ -23,44 +24,43 @@ const repoData = {
...
-      await loadSubSchema(body.oneOf, body.$id)
-    }
-  } catch (e) {
-    if (isErrnoException(e) && e.code === 'ENOENT') {
+export const getVersionRefs = memoizee(
+  async function () {
+    const releases = (await octokit.rest.repos.listReleases(repoData)).data.map(
+      (x) => ({ name: x.name, ref: x.tag_name })
+    )
+    return [...releases, { name: 'main', ref: 'main' }]
```
lpsinger, line 32: Use getDefaultBranch here instead of hardcoding the branch.

lpsinger (changes requested):

app/routes/docs_.schema.$version.$/route.tsx
lpsinger: No, I don't think so. You declared it like this in the loader:

```
let data: GitContentDataResponse[]
```

app/lib/schema-data.server.ts
lpsinger: This doesn't look like it's resolved. Did you forget to push some changes?

app/routes/docs_.schema.$version.$/route.tsx
dakota002:
```
return json(
    { data, jsonContent, examples, versions },
    { headers: publicStaticShortTermCacheControlHeaders }
  )
```
seems to break the ties of the typing in the loader, as the returned object from the loader is a `Response`:
image


 where as 

```
return{ data, jsonContent, examples, versions }
```
keeps the types:
image

The latter allows the type inference to work in the component, is it possible to return the headers with this method, or do i have to use the former in order to appropriately return the cache control headers?

lpsinger: Type inference for a loader that uses the `json()` helper works here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/circulars.%24circularId.tsx

dakota002: I figured it out, I was importing json from react-router, the example you provided was importing it from `'@remix-run/node'`. Should be all set in the next commit

app/routes/docs_.schema.$version.$/components.tsx
```diff
@@ -36,9 +36,17 @@ export type Schema = {
   required?: string[]
 }
 
+function formatLinkString(path: string) {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
lpsinger, line 40:
```suggestion
function useLinkString(path: string) {
```

app/lib/schema-data.server.ts
dakota002: Was waiting to push until i figured out the loader type issue, but should be good now

lpsinger (changes requested):

app/routes/docs_.schema.$version.$/components.tsx
lpsinger: I don't think that this is resolved. Did you forget to push something?

dakota002: Oops, this time yes, pushed now

lpsinger (approved):

PR #1285, opened 2023-08-24
Updates c# code sample

### LINKED ISSUE #1249: C# client sample code should not throw an exception for topic not available
The C# client sample code throws an exception and terminates if one of the requested topics is not available:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/4b6f441833c2869d0c87e0de73fd5934b997fcef/app/components/ClientSampleCode.tsx#L473-L479

This is inconsistent with the other client samples. It should print a warning or error message to stderr and  continue processing of the other topics.

dakota002:
Closes #1249

Prints the error message when topic is not available

### DIFF
--- a/app/components/ClientSampleCode.tsx
+++ b/app/components/ClientSampleCode.tsx
@@ -475,13 +475,13 @@ export function ClientSampleCode({
                 }
                 catch (Exception ex)
                 {
-                  if (!ex.Message.Contains("Subscribed topic not available"))
+                  if (ex.Message.Contains("Subscribed topic not available"))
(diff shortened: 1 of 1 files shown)


lpsinger (changes requested):
Remove the logical not (`!`) by switching the `if` and `else` clauses.

```c#
if (ex.Message.Contains("Subscribed topic not available"))
{
  Console.WriteLine(ex.Message);
}
else
{
  throw;
}
```

lpsinger (approved):

Issue #1286, opened 2023-08-24
Download Circulars archive fails with HTTP error 500


lpsinger:
Both https://dev.gcn.nasa.gov/circulars/archive.json.tar and https://dev.gcn.nasa.gov/circulars/archive.txt.tar return an HTTP 500 response with the text `{"message":"Internal Server Error"}`.

I don't think that our hosting target currently supports streaming responses --- it has to do with using AWS Lambda + AWS API Gateway; see https://github.com/remix-run/remix/discussions/6037.

Let's try returning the full tarball as one big blob rather than streaming it.

If this is not going to be a quick fix, then we might need to hide the download button behind a feature flag.

PR #1287, opened 2023-08-24
converting download back to blob instead of stream

Courey:
fixes issue 1286 https://github.com/nasa-gcn/gcn.nasa.gov/issues/1286


lpsinger (changes requested):
I think that there's still value in iterating over the circulars table page by page, even though all of the resultant _blobs_ will end up in memory.

lpsinger (approved):

PR #1316, opened 2023-08-31
Courey/download circulars s3

Courey:


lpsinger (changes requested):
High level comments:

- [ ] You _should_ be able to remove the two resource routes, because the static bucket will already be mounted at a URL in the API Gateway --- that's why we're trying this approach of using the existing static bucket.
- [ ] Study whether you can stream the tarball chunks to S3 as they are generated, to avoid holding them in memory.
- [ ] By iterating over the DynamoDB scan pages twice, we are incurring twice as much cost for DynamoDB usage as necessary. And I anticipate adding _more_ actions that are done for each Circular. Can we structure this so that we iterate over it _once_?

Courey:
closed because we took a different approach in https://github.com/nasa-gcn/gcn.nasa.gov/pull/1319

PR #1319, opened 2023-09-01
initial design for scheduled circular handler

### LINKED ISSUE #1289: Tar archive endpoints are returning HTTP 500 error because they are timing out
The Lambda timeout is currently set to 10 seconds.

We should figure out where the bottleneck is: reading from DynamoDB? rendering the Circulars in the appropriate format? adding them to the tar archive? concatenating all of the buffers?

Courey:
Fixes #1289.


app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  for await (const page of pages) {
+    const items: Circular[] = page.Items as Circular[]
+    yield items
+  }
+}
+
+// FIXME: must use module.exports here for OpenTelemetry shim to work correctly.
+// See https://dev.to/heymarkkop/how-to-solve-cannot-redefine-property-handler-on-aws-lambda-3j67
+module.exports.handler = async () => {
+  const actions: CircularAction[] = [{ callback: uploadTar }]
```
Courey, line 35: you put your actions in here after importing them. all that should need to be passed in is the array of circulars that are coming back in chunks.

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { tables } from '@architect/functions'
+import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
+import { paginateScan } from '@aws-sdk/lib-dynamodb'
+
+import { uploadTar } from './uploadTar'
+import { type Circular } from '~/routes/circulars/circulars.lib'
+
+interface CircularAction {
+  callback: (circularArray: Circular[]) => void
```
lpsinger, line 16: The callback should definitely be an async function:

```suggestion
  callback: (circularArray: Circular[]) => Promise<void>
```

Also, let's call it something other than "callback", because callbacks have a confusing different meaning when you talk about promises in JavaScript.

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+async function mapCirculars(...actions: CircularAction[]) {
+  for await (const circularArray of getAllRecords()) {
+    for (const action of actions) {
+      action.callback(circularArray)
+    }
```
lpsinger, line 23:
```suggestion
    Promise.all(actions.map(action) => action(circularArray))
```

lpsinger (changes requested):
We should probably define CircularAction in a file other than `index.ts` in order to avoid circular (ha ha) references.

Courey:
okay, it's changed a LOT since you looked at it. It is now at a hypothetically reviewable state. It's all just an educated guess, but do you at least like the concept behind it? I didn't try to convert yours because it was challenging enough to do my own and I didn't have any context for your cron yet.

app/scheduled/circulars/circularAction.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+export interface CircularAction<CircularActionContext> {
+  initialContext: () => CircularActionContext
+  currentContext: CircularActionContext
+  action: (
+    circularArray: Circular[],
```
lpsinger, line 14:
```suggestion
    circulars: Circular[],
```

app/scheduled/circulars/circularAction.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+export interface CircularAction<CircularActionContext> {
+  initialContext: () => CircularActionContext
+  currentContext: CircularActionContext
```
lpsinger, line 12: Remove the context object from the interface.

```suggestion
```

app/scheduled/circulars/circularAction.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+export interface CircularAction<CircularActionContext> {
+  initialContext: () => CircularActionContext
```
lpsinger, line 11:
```suggestion
  initialize: () => Promise<CircularActionContext>
```

app/scheduled/circulars/circularAction.ts
```diff
@@ -0,0 +1,22 @@
...
+  currentContext: CircularActionContext
+  action: (
+    circularArray: Circular[],
+    context: CircularActionContext
+  ) => Promise<CircularActionContext>
```
lpsinger, line 16: The callback probably shouldn't modify the context.```suggestion
  ) => Promise<void>
```

File: app/scheduled/circulars/circularAction.ts [hunk 8]
```diff
@@ -0,0 +1,22 @@
...
+  finalize: (context: CircularActionContext) => Promise<void>
+}
+
+export interface CircularActionContext {
+  context: object
+}
```
lpsinger, line 22: This interface doesn't seem very useful.

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        ? action.currentContext
+        : action.initialContext()
+      const results = await action.action(circularArray, context)
+      action.currentContext = results
+    }
+  }
+  for (const action of actions) {
+    // the finalize context is based the final results of the callback function
+    await action.finalize(action.currentContext)
+  }
```
lpsinger, line 32: Use `Promise.all` to do as much work as possible concurrently.

```suggestion
  const contexts = await Promise.all(actions.map({initialize} => initialize()))
  for await (const circulars of getAllRecords()) {
    for (const action of actions) {
      await Promise.all(actions.map(({action}, i) => action(circulars, contexts[i])))
    }
  }
  await Promise.all(actions.map(({finalize}, i) => finalize(contexts[i])))
```

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+async function* getAllRecords(): AsyncGenerator<Circular[], void, unknown> {
+  const db = await tables()
+  const client = db._doc as unknown as DynamoDBDocument
+  const TableName = db.name('circulars')
+  const pages = paginateScan({ client }, { TableName })
+
+  for await (const page of pages) {
+    const items: Circular[] = page.Items as Circular[]
+    yield items
```
lpsinger, line 43:
```suggestion
    yield page.Items as Circular[]
```

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      initialContext: setupTar,
+      finalize: finalizeTar,
+      currentContext: { context: {} },
+    },
+    {
+      action: uploadTxtTar,
+      initialContext: setupTar,
+      finalize: finalizeTar,
+      currentContext: { context: {} },
+    },
```
lpsinger, line 62: Define these two `CircularAction`s in uploadTar.ts and import them here, rather than importing every single one of their methods.

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,119 @@
...
+
+interface TarContextObject {
+  pack: Pack
+  readableTar: Readable | ReadableStream<Uint8Array>
+  fileType: string
+}
+
+async function uploadStream(tarContext: TarContextObject) {
+  const s3 = new S3Client({})
+  const Bucket = getEnvOrDie('ARC_STATIC_BUCKET')
```
lpsinger, line 27: Move these two constants to file scope.

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,119 @@
...
+      delete circular.sub
+      const json_entry = pack.entry(
+        { name: `archive.json/${circular.circularId}.json` },
+        JSON.stringify(circular, null, 2)
+      )
+      json_entry.end()
+    }
+  }
+  return { context: { pack, readableTar } }
+}
```
lpsinger, line 66: There are some refactorings in #1288 that make this a lot simpler. See https://github.com/nasa-gcn/gcn.nasa.gov/blob/c3b53c2086d550457813a3acfd5a0ad772db2e0b/app/routes/circulars/circulars.server.ts#L273-L318.

lpsinger (changes requested):

app/scheduled/circulars/circularAction.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+export interface CircularAction {
+  initialize: () => Promise<object>
+  action: (circulars: Circular[], context: object) => Promise<void>
+  finalize: (context: object) => Promise<void>
+}
```
lpsinger, line 14: Make this a generic type:

```suggestion
export interface CircularAction<T = unknown> {
  initialize: () => Promise<T>
  action: (circulars: Circular[], context: T) => Promise<void>
  finalize: (context: T) => Promise<void>
}
```

app/scheduled/circulars/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    yield page.Items as Circular[]
+  }
+}
+
+// FIXME: must use module.exports here for OpenTelemetry shim to work correctly.
+// See https://dev.to/heymarkkop/how-to-solve-cannot-redefine-property-handler-on-aws-lambda-3j67
+module.exports.handler = async () => {
+  const actions: CircularAction[] = [jsonUploadAction, txtUploadAction]
+
+  await mapCirculars(...actions)
```
lpsinger, line 44:
```suggestion
  await mapCirculars(jsonUploadAction, txtUploadAction)
```

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,124 @@
...
+import { createReadableStreamFromReadable } from '@remix-run/node'
+import { Readable } from 'stream'
+import type { Pack } from 'tar-stream'
+import { pack as tarPack } from 'tar-stream'
+
+import { getEnvOrDie } from '~/lib/env.server'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+import { formatCircular } from '~/routes/circulars/circulars.lib'
+
+interface TarContextObject {
```
lpsinger, line 18:
```suggestion
interface TarContext {
```

lpsinger (changes requested):

app/scheduled/circulars/uploadTar.ts
Courey: COUREY: Add context here so it's {context: {pack.....}}

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,124 @@
...
+  pack: Pack
+  readableTar: Readable | ReadableStream<Uint8Array>
```
Courey, line 20: COUREY: make this tarStream

app/scheduled/circulars/circularAction.ts
lpsinger: Actually, that should be `export interface CircularAction<T = any>`.

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { Circular } from '~/routes/circulars/circulars.lib'
+import { formatCircular } from '~/routes/circulars/circulars.lib'
+
+interface TarContext {
+  context: {
+    pack: Pack
+    tarStream: Readable | ReadableStream<Uint8Array>
+    fileType: string
+  }
+}
```
lpsinger, line 24: Why do you have an extra object nested inside? Why not the following:

```suggestion
interface TarContext {
  pack: Pack
  tarStream: Readable | ReadableStream<Uint8Array>
  fileType: string
}
```

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  pack.on('end', () => {
+    tarStream.push(null)
+  })
+
+  return { context: { pack, tarStream } }
+}
+
+export async function finalizeTar(context: object) {
+  const finalContext = context as unknown as TarContext
```
lpsinger, line 85: Why not strongly type the context parameter?

```suggestion
export async function finalizeTar(context: TarContext) {
```

app/scheduled/circulars/uploadTar.ts
```diff
@@ -0,0 +1,125 @@
...
+  })
+}
+
+export const jsonUploadAction = {
+  action: uploadJsonTar,
+  initialize: setupTar,
+  finalize: finalizeTar,
+}
+
+export const txtUploadAction = {
```
lpsinger, line 121:
```suggestion
export const txtUploadAction: CircularAction<TarContext> = {
```

lpsinger (changes requested):

app/scheduled/circulars/uploadTar.ts
Courey: I think that's a holdover from when I was passing data back in the action. it had to match the context object shape. I'll adjust it.

app/scheduled/circulars/uploadTar.ts
lpsinger: Why would you need a type for each job?

Courey: it was a holdover from how it was before. when I was passing in the functions to the index instead of declaring them in the job file itself.

Courey: I changed it.

lpsinger (approved):

PR #1343, opened 2023-09-08
Add redirect to Circulars archive tarballs in static bucket

lpsinger:
API Gateways have a limit on response size of 10 MB, so we cannot rely on the API Gateway to server the Circulars archive.


app/routes/circulars.archive.$suffix[.]tar.ts
```diff
@@ -0,0 +1,21 @@
+/*!
```
Comment (Courey) [inline app/routes/circulars.archive.$suffix[.]tar.ts:1]: i really like how you made $suffix dynamic here.

Courey (approved):

PR #1346, opened 2023-09-08
fixing dynamic routing

Courey:


lpsinger:
#1345 was supposed to fix this, but it didn't. I'm not sure why.

Courey:
> #1345 was supposed to fix this, but it didn't. I'm not sure why.

was it supposed to fix the bucket address or the fact that it wasn't hitting the loader? I don't think it was even hitting the loader before because it was 404ing. I'm still digging into the bucket address. It looks funny in the url
Screenshot 2023-09-08 at 10 27 54 AM

Courey:
locally:
/Users/courey/dev/gcn.nasa.gov/build/static --bucket
us-east-1 --region
circulars/archive.json.tar --key
the url is funny because the bucket is the static folder instead of the actual bucket name since it's local.

lpsinger:
The problem is that the the `circulars.$circularId.tsx` route module is handling the /circulars/archive.json.tar route.

Courey:
> The problem is that the the `circulars.$circularId.tsx` route module is handling the /circulars/archive.json.tar route.

ah because it's treating "archive.json.tar" as one thing. that's why changing it to be /circulars/archive/json.tar worked

Courey:
it will also work if you make the files circulars.archive[.]json[.]tar.ts and circulars.archive[.]txt[.]tar.ts

lpsinger:
Ah, I see. But I'd like the URLs to be /circulars/archive.tar.json and /circulars/archive.txt.json.

Courey:
okay, reverted to the way that keeps the url intact.

PR #1368, opened 2023-09-13
Draft: incremental adoption of Astro Flavored Markdown

lpsinger:
This is an idea for incremental adoption and development of Astro Flavored Markdown. The idea is that as people are continuing to submit plaintext GCN Circulars, we can parse _just_ the astronomy related syntax but not the general Markdown syntax. So, for example, whitespace in GCN Circualrs is treated the same as it is now, and none of the Markdown syntax such as `# Headings` or `**emphasis**` are processed, but GCN Circulars cross references _are_ processed.

The way this works is that we create a fake Markdown syntax tree for the GCN Circular that puts all of its text in a single paragraph, then we pass that through the Astro Flavored Markdown processor.

This patch also demonstrates a method of enriching Astro Flavored Markdown using React components. In the Markdown to HTML processing stage ([remark-rehype](https://github.com/remarkjs/remark-rehype)), we transform the Astro Flavored Markdown AST node into an HTML [`<data>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/data) with the `class` attribute set to the Astro Flavored Markdown type and the `value` attribute set to the data value. The content is the original text.

For example, the following Markdown text:

```md
See GCN Circular 123a
```

is transformed to this Markdown AST:

```json
{
  "type": "root",
  "children": [
    {
      "type: "paragraph",
      "children": [
        {
          "type": "text",
          "value": "see GCN Circular"
        },
        {
          "type": "text",
          "value": "123a",
          "data": {
            "astro": {
              "type": "gcn-circular",
              "value": 123.5
            }
          }
        }
      ]
    }
  ]
}
```

And this, in turn, is transformed to the following HTML:

```html
<p>
  See GCN Circular
  <data class="gcn-circular" value="123.5">
    123a
  </data>
</p>
```

Finally, at the HTML to React stage (which is done by [rehype-react](https://github.com/rehypejs/rehype-react)) we can transform HTML `<data>` elements to React components of our choosing.


jracusin:
How does it recognize the syntax that you're referencing a GCN Circular?  People sometimes use "GCN 12345" or "GCN #12345" or "GCN Circ. 12345", etc.  Also, is there any validation that that circular number is valid?

lpsinger:
> How does it recognize the syntax that you're referencing a GCN Circular? People sometimes use "GCN 12345" or "GCN #12345" or "GCN Circ. 12345", etc.

That's set by this regular expression: https://github.com/nasa-gcn/remark-rehype-astro/blob/main/src/replacements/gcn-circular/index.ts#L8

> Also, is there any validation that that circular number is valid?

No, that would be harder.

Vidushi-GitHub:
Looks good! Why it has wrong GCN Circular no instead of 33372 in heading?

lpsinger:
> Looks good! Why it has wrong GCN Circular no instead of 33372 in heading?

Because this is a screen shot of the application in local sandbox mode, and it using the [sandbox seed data](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/sandbox-seed.json).

Vidushi-GitHub:
For hyperlink it will first link with GCN arXiv then other arXiv. Could you please share the example of how paper link is working?

lpsinger:
> For hyperlink it will first link with GCN arXiv then other arXiv. Could you please share the example of how paper link is working?

I'm sorry, what is "GCN arXiv"?

Vidushi-GitHub:
> > For hyperlink it will first link with GCN arXiv then other arXiv. Could you please share the example of how paper link is working?
> 
> I'm sorry, what is "GCN arXiv"?

My bad, I was referring to circulars page (https://gcn.nasa.gov/circulars). I was trying to understand app/routes/circulars.$circularId/body.tsx, if condition is working well for circulars page, else condition is directed to nasa ads arXiv. I was wondering the result of else condition as paper citation.

Vidushi-GitHub (approved):
Good to go! May be we can add paper links as well in future.

lpsinger:
> Good to go! May be we can add paper links as well in future.

Well, it already supports arXiv links. And I'll add DOIs shortly.

PR #1390, opened 2023-09-19
Display Circular text in monospace font

lpsinger:
Circulars frequently contain ASCII art tables which only display correctly with a monospace font.

When we start supporting Markdown, we'll want to display Circulars that are in plain text in a monospaced font but Circulars that are entered as Markdown in a variable-width sans-serif font (i.e., the default body text of the web site).


Vidushi-GitHub (approved):
sounds right.

PR #1423, opened 2023-09-25
Fix for windows build

dakota002:
Can't build on my machine without this package

```
Error: The package "@esbuild/win32-x64" could not be found, and is needed by esbuild.

```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@architect/functions": "^5.3.4",
+        "@esbuild/win32-x64": "^0.19.3",
         "@nasa-gcn/architect-functions-search": "^1.0.0",
         "@nasa-gcn/dynamodb-autoincrement": "^1.0.0",
         "@nasa-gcn/remark-rehype-astro": "^0.1.3",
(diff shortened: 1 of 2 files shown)


lpsinger:
@dakota002, would you please open a separate PR to make the node.js GitHub workflow a matrix that runs on Ubuntu, macOS, and Windows, so that we catch things like in advance in the future?

And does rolling back to an earlier version of esbuild fix this? I would rather not add an unconditional dependency on binaries for a specific platform, which won't be used on other platforms.

dakota002:
Sure, did we remove esbuild as a dependency in general? I dont see it in the dev deps, but I also just tried `npm install --save-exact --save-dev esbuild` and it is working again too

lpsinger:
> Sure, did we remove esbuild as a dependency in general? I dont see it in the dev deps, but I also just tried `npm install --save-exact --save-dev esbuild` and it is working again too

It's a dependency of `@remix-run/dev`:

```
$ npm ls esbuild
@nasa-gcn/gcn.nasa.gov@ /Users/lpsinger/src/gcn.nasa.gov
└─┬ @remix-run/dev@2.0.1
  ├─┬ @vanilla-extract/integration@6.2.1
  │ ├── esbuild@0.17.6 deduped
  │ └─┬ vite@4.3.9
  │   └── esbuild@0.17.15
  ├─┬ esbuild-plugins-node-modules-polyfill@1.6.1
  │ └── esbuild@0.17.6 deduped
  └── esbuild@0.17.6
```

lpsinger:
@dakota002, does it work if you do a fully clean build from the main branch? i.e., blow away `node_modules` first?

dakota002:
Wow, yea apparently it did. I ran `npm run clean` without luck. Okay that just solved the problem. Thank you! We can just close this and that other matrix PR out

PR #1424, opened 2023-09-25
Workflow matrix

dakota002:


lpsinger (approved):
How odd that Prettier checks are failing on Windows only. Some Windows-specific quirk around newlines or BOMs, perhaps?

dakota002:
Yea, if I run `npx prettier -w .` It actually shows that there are changes in like every file, but they get cleaned up when I run `git add`

lpsinger:
> Yea, if I run `npx prettier -w .` It actually shows that there are changes in like every file, but they get cleaned up when I run `git add`

Please study https://prettier.io/docs/en/options.html#end-of-line and propose a fix.

PR #1425, opened 2023-09-25
Circular maintenance

dakota002:


dakota002:
@lpsinger This is becoming a pretty hefty PR, are there any parts that you think could be broken out into some smaller PRs to avoid crazy future merge conflicts?

lpsinger:
> @lpsinger This is becoming a pretty hefty PR, are there any parts that you think could be broken out into some smaller PRs to avoid crazy future merge conflicts?

I would start with the back end.

lpsinger:
Did we get all of the pieces of this onto main? Can we close it now?

dakota002:
> Did we get all of the pieces of this onto main? Can we close it now?

Yup! I'll close it

PR #1427, opened 2023-09-26
Fixes unexpected server error from trying to hit the text page for circulars

dakota002:
https://dev.gcn.nasa.gov/circulars/33638.txt 

The error:
image

### DIFF
--- a/__tests__/circulars.ts
+++ b/__tests__/circulars.ts
@@ -5,7 +5,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import dedent from 'ts-dedent'
+import { dedent } from 'ts-dedent'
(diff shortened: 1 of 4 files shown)


lpsinger (changes requested):
We have three other instances of ts-dedent using the default import. Would you please fix them too?

dakota002:
There is one remaining, in the tests, however this is not throwing an error and the tests seem to still run successfully

lpsinger:
> There is one remaining, in the tests, however this is not throwing an error and the tests seem to still run successfully

Sure, but I figure we might as well be consistent. Would you update that one too, please?

lpsinger (approved):

PR #1455, opened 2023-09-28
Add new tables for versioning and changes, respective CRUD stuff

dakota002:
Backend breakout for further updates with versioning


app/routes/circulars/circulars.lib.ts
```diff
@@ -116,6 +126,7 @@ export const validSubjectKeywords = [
   'Baksan Neutrino Observatory Alert',
   'CALET',
   'Chandra',
+  'Duplicate Submission',
```
lpsinger, line 129: Move these two keyword additions to a separate PR.

lpsinger (changes requested):
- Move circular keywords into a separate table.

app/routes/circulars/circulars.lib.ts
lpsinger: I don't understand. Do people request changes to a Circular by submitting a Circular with these magic works in the subject?

dakota002: Ah I see, no I think I am seeing an error in my implementation. I don't want the users to submit circulars where the value is literally 'Duplicate Submission', I was checking that the string was valid, but it shouldn't be the case that these are just considered valid keywords. 

In the [big PR](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1425) I had a couple simplified workflows that would be shortcuts for mods to just click a button and it would handle the workflow for duplicates/mistaken submissions

I have some ideas for how to improve this, the functionality as I added it is wrong, I'll update this shortly

lpsinger (changes requested):
- What is the purpose of `erroneousSubmissionKeywords`? Do you want to require users to insert one of these keywords into the subject lines for proposed edits of GCN Circulars? Why? The fact that a Circular has a version number that is greater than one will indicate that it has been edited, and we can use that to mark it differently in the UI.
- We should be careful to generate new versions using transactions to avoid race conditions. Fortunately, the logic should be almost identical to what we do in dynamodb-autoincrement. Would you please propose a patch for dynamodb-autoincrement to either (1) generalize it to support version control applications or (2) add a second class that implements version control?

app/routes/circulars/circulars.server.ts
```diff
@@ -263,3 +273,280 @@ export async function circularRedirect(query: string) {
...
+  if (!user?.groups.includes(moderatorGroup))
+    throw new Response('User is not in the submitters group', {
+      status: 403,
+    })
+
+  if (
+    !subjectIsValid(circular.subject) &&
+    !erroneousSubmissionKeywords.includes(circular.subject)
+  )
+    throw new Response('subject is invalid', { status: 400 })
```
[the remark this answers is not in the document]
dakota002: Not if the subject contains an `erroneousSubmissionKeyword`, but that `erroneousSubmissionKeywords` includes the subject.

If the subject is not valid and the subject is not literally one of the `erroneousSubmissionKeywords`: throw error

so if either the subject is valid, or the subject is one of the listed items in the `erroneousSubmissionKeywords`, it will continue, no?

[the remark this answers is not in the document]
lpsinger: Oh, I get it now.

app/routes/circulars/circulars.lib.ts
```diff
@@ -158,6 +168,11 @@ export const validSubjectKeywords = [
   'ZTF',
 ]
 
+export const erroneousSubmissionKeywords = [
```
lpsinger, line 171: Rename to make it clear that these are not partial matches, these are whole subjects.

```suggestion
export const erroneousSubmissionSubjects = [
```

lpsinger (changes requested):

app/routes/circulars/circulars.lib.ts
dakota002: Good idea, I'll have that in the next commit

dakota002:
Converting to draft PR, need to revise the implementation further

lpsinger:
You've requested a review from me, but this is still marked as a draft. Is it ready for review? If so, then please remove the draft status.

lpsinger:
> Converting to draft PR, need to revise the implementation further

Oh, OK. Let me know when you are ready for a review.

dakota002:
Okay, should be ready now. Had to strip out the stuff that the auto-incrementer was updated to handle

dakota002:
Rebased, should resolve the conflicts

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -46,6 +54,28 @@ export const getDynamoDBAutoIncrement = memoizee(
...
+  return new DynamoDBHistoryAutoIncrement({
+    doc,
+    counterTableName,
+    counterTableKey: { circularId },
+    attributeName: 'version',
+    tableName,
+    initialValue: 1,
+    dangerously,
+  })
+})
```
lpsinger, line 77: This will store a memoized autoincrement object _for each circular ID_. That's too much!

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+  circularId: number
+): Promise<Circular[]> {
+  const db = await tables()
+  const result = await db.circulars_history.query({
+    KeyConditionExpression: 'circularId = :circularId',
+    ExpressionAttributeValues: {
+      ':circularId': circularId,
+    },
+  })
+  return result.Items as Circular[]
```
lpsinger, line 340:
```suggestion
  return result.Items
```

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+  const result = await db.circulars_history.get({
+    circularId,
+    version,
+  })
+  if (!result)
+    throw new Response(null, {
+      status: 404,
+    })
+  return result
+}
```
lpsinger, line 323: Let's merge this functionality into the existing `get()` function, adding an optional version positional argument.

Also: since the _current_ version is not stored in the history table, if the caller requests the current version by its number, then we need to correctly handle that case.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+
+/**
+ * Gets all entries in circulars_history for a given circularId
+ * @param circularId
+ * @returns an array of previous versions of a Circular sorted by version
+ */
+export async function getCircularHistory(
```
lpsinger, line 330: - No need for "Circular" to be in the name... it's clear because this is in circulars.server.ts.
- Consider `getVersions()` instead of `getHistory()` to reduce the number of different similar but distinct terms in function names (history, version, change)
- Is it more useful to return the content of all versions, or simply a list of their version numbers? Perhaps version numbers and titles, like `CircularMetadata[]`? I'm thinking that the main use of this function will be to implement an index route that lists all of the versions.

```suggestion
export async function getVersions(
```

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+  if (!subjectIsValid(subject))
+    throw new Response('subject is invalid', { status: 400 })
+  if (!bodyIsValid(body)) throw new Response('body is invalid', { status: 400 })
+
+  const user = await getUser(request)
+  if (!user)
+    throw new Response('User is not signed in', {
+      status: 403,
+    })
+  const requestor = formatAuthor(user)
```
lpsinger, line 372: This duplicates much of the logic from the existing `put` function. Please refactor.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+  const db = await tables()
+  return (
+    await db.circulars_change_requests.query({
+      KeyConditionExpression: 'circularId = :circularId',
+      ExpressionAttributeValues: {
+        ':circularId': circularId,
+      },
+    })
+  ).Items
+}
```
lpsinger, line 400: Again, do we want this to return the entire content of all of the change requests, or just a list of them?

app.arc
```diff
@@ -72,6 +72,16 @@ auto_increment_metadata
...
 
+circulars_history
+  circularId *Number
+  version **Number
+  PointInTimeRecovery true
+
+circulars_change_requests
+  circularId *Number
+  requestorSub **String
+  PointInTimeRecovery true
```
lpsinger, line 83: Will this data model limit us to one change request per circular per user at a time?

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +291,198 @@ export async function circularRedirect(query: string) {
...
+) {
+  if (!subjectIsValid(subject))
+    throw new Response('subject is invalid', { status: 400 })
+  if (!bodyIsValid(body)) throw new Response('body is invalid', { status: 400 })
+
+  const user = await getUser(request)
+  if (!user)
+    throw new Response('User is not signed in', {
+      status: 403,
+    })
```
lpsinger, line 371: Should we only permit users who are submitters to enter change requests?

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
lpsinger: I don't know. @jracusin?

app/routes/_gcn.circulars/circulars.server.ts
lpsinger: Please _don't_ memoize this.

lpsinger (changes requested):
I see that you requested another review from me, but there are still unresolved comments from the last review.

app.arc
dakota002: Yes. I think that makes the most sense for this use case. Once approved, this just sets the subject and body to whatever the request has in it, so if they make multiple requests at the same they will result in overwriting. 

If we see more of a need for some granularity in the future we could implement more of a diff based approach

app/routes/_gcn.circulars/circulars.server.ts
dakota002: Yup! updated in latest commit

dakota002:
> I see that you requested another review from me, but there are still unresolved comments from the last review.

Oh that shouldn't have triggered the re-review until now, I was still working on a commit. It should be ready for review now though, with the exception of the point waiting for @jracusin's feedback

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -173,12 +201,30 @@ export async function search({
...
+    if (!result)
+      throw new Response('Specified version does not exist for this circular', {
+        status: 404,
+      })
+  } else {
+    result = await db.circulars.get({
+      circularId,
+    })
+  }
```
lpsinger, line 227: A couple issues with this:

- The code within the `if (version) { if (isNaN(version)) { ... }}` block cannot be reached because `NaN` is falsey.
- When getting a Circular by version, we _still_ need to handle the case that the current version _is_ the requested version.
- The new `if (!result)` section is redundant because that check already exists below.

One possible refactoring:

```suggestion
  if (version) {
    result = await db.circulars_history.get({
      circularId,
      version,
    })
  }
  if (!result && !isNaN(version)) {
    result = await db.circulars.get({
      circularId
    })
  }
  if (version && result?.version !== version)
  {
    result = undefined
  }
```

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +304,169 @@ export async function circularRedirect(query: string) {
...
+export async function getVersions(circularId: number): Promise<Circular[]> {
+  const db = await tables()
+  const result = await db.circulars_history.query({
+    KeyConditionExpression: 'circularId = :circularId',
+    ExpressionAttributeValues: {
+      ':circularId': circularId,
+    },
+  })
+  return result.Items
+}
```
lpsinger, line 322: Do we want to return the _content_ of all of the circulars, or just a list of their version numbers?

Also, this does not include the newest or current version number.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +304,169 @@ export async function circularRedirect(query: string) {
...
+ * @param requestorSub
+ * @param request
+ */
+export async function verifyAndDeleteChangeRequest(
+  circularId: number,
+  requestorSub: string,
+  request: Request
+): Promise<void> {
+  const user = await getUser(request)
+  if (!user) throw new Response('User must be signed in')
```
lpsinger, line 399:
```suggestion
  if (!user) throw new Response(null, { status: 403 })
```

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +304,169 @@ export async function circularRedirect(query: string) {
...
+ * Throws an HTTP error if:
+ *  - The current user is not signed in
+ *  - The current user is not the same as the user who requested
+ *    the changed, OR the current user is not a moderator
+ *
+ * @param circularId
+ * @param requestorSub
+ * @param request
+ */
+export async function verifyAndDeleteChangeRequest(
```
lpsinger, line 393: None of the other exported functions have names that start with `verifyAnd`. Suggest renaming this `verifyAndDeleteChangeRequest` and the internal function `deleteChangeRequestInternal` or `deleteChangeRequestRaw`.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +304,169 @@ export async function circularRedirect(query: string) {
...
+    })
+
+  const db = await tables()
+  const changeRequest = (await db.circulars_change_requests.get({
+    circularId,
+    requestorSub,
+  })) as CircularChangeRequest
+
+  if (!changeRequest)
+    throw new Response('No change request found', { status: 404 })
```
lpsinger, line 454: Factor out as `getChangeRequest`?

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
dakota002: It should include all versions that exist, since the auto increment puts into both tables, right? I'll need to test a bit. I think in the previous iteration I assumed that the history didn't contain the latest, and I just joined the history and the current entry in the circulars table. I can edit this to make sure it returns the whole list

lpsinger: Remember that in @nasa-gcn/dynamodb-autoincrement we landed on a design in which all versions _except for the most recent one_ are stored in the history table.

I suggest the following:

- Change the signature of this function so that it returns a list of version numbers.
- Add a projection to the query so that it only returns the circularId attribute.
- Before returning, sort the versions, and push one additional implied version.

dakota002: Ah right, okay I updated this and another bug in the get method in this latest commit

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -173,12 +201,31 @@ export async function search({
...
+  let result = version
+    ? await db.circulars_history.get({
+        circularId,
+        version,
+      })
+    : await db.circulars.get({
+        circularId,
+      })
+
+  if (!result && version && !isNaN(version)) {
```
lpsinger, line 220: Remember that `Boolean(NaN) === false`. This `if` statement is not correct.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +306,181 @@ export async function circularRedirect(query: string) {
...
+    await db.circulars_history.query({
+      KeyConditionExpression: 'circularId = :circularId',
+      ExpressionAttributeValues: {
+        ':circularId': circularId,
+      },
+      ProjectionExpression: 'version',
+    })
+  ).Items
+
+  return [...versions.map((x) => x.version), versions.length + 1]
```
lpsinger, line 327: Note that this will return `[1]` if (a) the Circular exists but has only one version or (b) the Circular does not exist. Do we care to be correct for the latter corner case?

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
dakota002: ah good catch, I think I am misunderstanding what this case should be catching. Is this correct:

If the result isn't found in the initial check given a `circularId` and `version` (which is a number), just get whatever the latest/current version is. 

Then in the following block, if the provided version doesn't match on the found result, clear the result

app/routes/_gcn.circulars/circulars.server.ts
dakota002: I think the latter shouldn't be that much of an issue, since it is just returning an array, it won't really mean anything until we put it into further context, like generating links to view the versions

app/routes/_gcn.circulars/circulars.server.ts
lpsinger: Correct.

The following possible values of type `number | undefined` are falsey:

- undefined: no version passed as an argument, so don't even bother looking up by version
- NaN: an impossible value, should always throw 404
- 0: another impossible value, should always throw 404

dakota002: Gotcha, I had made the change to that if statement because of this error:
image

lpsinger: To satisfy type inference, you chould check that `number !== undefined && !number`.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -173,12 +201,31 @@ export async function search({
...
+
+  if (!result && version !== undefined && !version && !isNaN(version)) {
+    result = await db.circulars.get({
+      circularId,
+    })
+  }
+
+  if (version && result?.version !== version) {
+    result = undefined
+  }
```
lpsinger, line 228: So, the first ternary expression here still isn't quite right if we want a value of `undefined` to trigger lookup of the latest version but `0` or `NaN` to trigger an error.

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
dakota002: I see, do we just want the checks to short out in those cases? So move the check for If `0` or `NaN` ... up before doing any gets?

lpsinger: You could... or you could structure the body of the function like this.

```ts
if (version === undefined) {
  // get most recent version
} (else if !version) {
  // throw 404
} else {
  // look up by version number
}
```

dakota002: Ah okay I didn't see this comment until after I pushed, what do you think of the latest refactor? It moves the errors up to handle it all at the beginning

lpsinger: Ugh. You know what, if `version === 0` or `version === NaN`, I don't think you need to do anything special, since it will never match a real circular in the table.

Maybe this would be simpler using [function overloading](https://dev.to/zirkelc/how-to-overload-functions-in-typescript-22c3)?

```ts
export async function get(circularId: number) {
  ...
}

export async function get(circularId: number, version: number) {
  ...
}
```

lpsinger (changes requested):

dakota002:
True, maybe we can just scrap some of those lines? 

As for the overloading, I don't think that will help in this case. It looks like the way that TS supports overload methods is that you just define all the overloads at the top, then account for all cases within the function. So it would look like this (i removed some unneeded curly brackets to shrink it for this code block):

```ts
export async function get(circularId: number): Promise<Circular>
export async function get(
  circularId: number,
  version?: number
): Promise<Circular> {
  if (isNaN(circularId)) throw new Response(null, { status: 404 })
  if (version === 0 || (version !== undefined && isNaN(version)))
    throw new Response(null, { status: 404 })

  const db = await tables()
  let result = version
    ? await db.circulars_history.get({ circularId, version })
    : await db.circulars.get({ circularId })

  if (!result && version)
    result = await db.circulars.get({
      circularId,
    })

  if (version && result?.version !== version) result = undefined

  if (!result)
    throw new Response(null, {
      status: 404,
    })
  return result
}
```
 
Attempting to define 2 distinct methods with the same name throws these errors:

image

lpsinger:
> True, maybe we can just scrap some of those lines?

I'll make a PR to propose a version of this function. One moment.

lpsinger:
@dakota002, I tried myself to write list and get methods that include the current version, and I found corner cases that were difficult to correctly deal with performantly. I think that your original approach of excluding the current version from the list is better.

I opened https://github.com/nasa-gcn/dynamodb-autoincrement/pull/39 to implement this in the library.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -173,17 +201,24 @@ export async function search({
 }
 
 /** Get a circular by ID. */
-export async function get(circularId: number): Promise<Circular> {
+export async function get(
```
lpsinger, line 204: You can make this function a lot simpler now:

```
/** Get a circular by ID. */
export async function get(circularId: number, version?: number) {
  const circularVersions = await getDynamoDBVersionAutoIncrement(circularId)
  const result: Circular = await circularVersions.get(version)
  if (!result)
    throw new Response(null, {
      status: 404,
    })
  return result
}
```

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +294,174 @@ export async function circularRedirect(query: string) {
...
+ * @param circularId
+ * @returns an array of previous versions of a Circular sorted by version
+ */
+export async function getVersions(circularId: number): Promise<number[]> {
+  const circularVersionsAutoIncrement =
+    await getDynamoDBVersionAutoIncrement(circularId)
+  const versions = await circularVersionsAutoIncrement.list()
+
+  return [...versions, versions.length + 1]
+}
```
lpsinger, line 309:
```suggestion
export async function getVersions(circularId: number) {
  const circularVersionsAutoIncrement =
    await getDynamoDBVersionAutoIncrement(circularId)
  return await circularVersionsAutoIncrement.list()
}
```

package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@architect/functions": "^5.3.4",
     "@nasa-gcn/architect-functions-search": "^1.0.0",
-    "@nasa-gcn/dynamodb-autoincrement": "^2.1.0",
+    "@nasa-gcn/dynamodb-autoincrement": "^2.2.0",
```
lpsinger, line 31: Please rebase from the main branch. This change is already there.

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
dakota002: Sweet, updated in the latest commit with a rebase

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -261,3 +287,172 @@ export async function circularRedirect(query: string) {
...
+ * @param request
+ */
+export async function deleteChangeRequest(
+  circularId: number,
+  requestorSub: string,
+  request: Request
+): Promise<void> {
+  const user = await getUser(request)
+  if (!user) throw new Response(null, { status: 403 })
+  if (requestorSub != user.sub || !user.groups.includes(moderatorGroup))
```
lpsinger, line 378:
```suggestion
  if (requestorSub !== user.sub || !user.groups.includes(moderatorGroup))
```

lpsinger (changes requested):

PR #1456, opened 2023-09-28
Join values with pipe if object

### LINKED ISSUE #1453: Types for union fields are rendered as run-on words
Screenshot 2023-09-28 at 10 19 35

Instead of `numbernull`, this should be shown as `number | null`.

dakota002:
Resolves #1453

image

### DIFF
--- a/app/routes/docs_.schema.$version.$/components.tsx
+++ b/app/routes/docs_.schema.$version.$/components.tsx
@@ -17,7 +17,7 @@ export type ReferencedSchema = {
 
 export type SchemaProperty = {
   description: string
-  type?: string
+  type?: string | string[]
   enum?: string[]
   $ref?: string
 }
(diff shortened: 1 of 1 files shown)


app/routes/docs_.schema.$version.$/components.tsx
```diff
@@ -145,7 +145,8 @@ export function formatFieldName(name: string, requiredProps?: string[]) {
 }
 
 export function formatFieldType(item: SchemaProperty): string {
-  if (item.type) return item.type
+  if (item.type)
+    return typeof item.type == 'object' ? item.type.join(' | ') : item.type
```
lpsinger, line 149: Fewer nested `if` statements:

```suggestion
  if (item.type === 'object') return item.type.join(' | ')
  if (item.type) return item.type
```

lpsinger (changes requested):

lpsinger (changes requested):
Please fix CI failures.

lpsinger (approved):

PR #1457, opened 2023-09-28
Fix for table resizing

### LINKED ISSUE #1454: Columns in "Properties from all of..." table jump when expanding and collapsing rows
https://github.com/nasa-gcn/gcn.nasa.gov/assets/728407/8222b4cf-57ac-43dc-bb13-e220c38fb8b1

dakota002:
Collapsed:
image

Expanded:
image

Resolves #1454

### DIFF
--- a/app/routes/docs_.schema.$version.$/TableColumns.module.scss
+++ b/app/routes/docs_.schema.$version.$/TableColumns.module.scss
@@ -1,0 +1,3 @@
+.TableColumns {
+  width: 50%;
+}
--- a/app/routes/docs_.schema.$version.$/components.tsx
+++ b/app/routes/docs_.schema.$version.$/components.tsx
@@ -9,6 +9,8 @@ import { Link, useResolvedPath } from '@remix-run/react'
 import { Icon, Table } from '@trussworks/react-uswds'
 import { useState } from 'react'
 
+import styles from './TableColumns.module.css'
+
 export type ReferencedSchema = {
   $ref: string
   type: string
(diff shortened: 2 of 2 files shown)


lpsinger (commented):
I would rather avoid unnecessary custom CSS. The reason that the layout is jumping is that it's suddenly gaining an extra column. Can we just display the column always?

dakota002:
> I would rather avoid unnecessary custom CSS. The reason that the layout is jumping is that it's suddenly gaining an extra column. Can we just display the column always?

That is not the case. The resizing is happening because some content in the descriptions are much larger than others. The `colspan={2}` was making sure the additional column was catching the new columns 

With additional columns the table still resizes ("filler" added to show that it is there):
image

image

Using the `table-layout:fixed` also looks pretty bad, since it enforces equal sizing on all columns:
image

lpsinger:
> > I would rather avoid unnecessary custom CSS. The reason that the layout is jumping is that it's suddenly gaining an extra column. Can we just display the column always?
> 
> That is not the case. The resizing is happening because some content in the descriptions are much larger than others. The `colspan={2}` was making sure the additional column was catching the new columns

The react-uswds `<Table>` component has a `fixed` attribute that I think has the same effect as your custom CSS. Would you please try it?

dakota002:
> > > I would rather avoid unnecessary custom CSS. The reason that the layout is jumping is that it's suddenly gaining an extra column. Can we just display the column always?
> > 
> > 
> > That is not the case. The resizing is happening because some content in the descriptions are much larger than others. The `colspan={2}` was making sure the additional column was catching the new columns
> 
> The react-uswds `<Table>` component has a `fixed` attribute that I think has the same effect as your custom CSS. Would you please try it?

I did, the `fixed` attribute applies the `table-layout: fixed` style (https://github.com/trussworks/react-uswds/blob/main/src/components/Table/Table.module.css), which _should_ give the intended effect of enforcing two equally sized columns, however it does not seem to work correctly, at the moment, maybe a bug in their component?
image

image

dakota002:
I can look into their repo and see if I can contribute a fix if this is the case

dakota002:
> Try `<Table fixed fullWidth>`.

I did, thats where the `width-full` class is coming from in the picture^

lpsinger:
:shrug: I guess this will have to do then.

lpsinger (approved):

PR #1458, opened 2023-09-28
Add API for third parties to post GCN Circulars on behalf of users

lpsinger:


app/routes/api.circulars.ts
```diff
@@ -0,0 +1,190 @@
...
+ *    c. Save the resulting refresh token and auth token on the server side in
+ *       a record associated with the user's account.
+ *
+ *    d. To post a GCN Circular on behalf of the user, make a POST request to
+ *       https://<stage>.gcn.nasa.gov/api/circulars. Provide the access token
+ *       in the Authorization: Bearer header. The request body should be a JSON
+ *       document of the form '{"subject": "...", "body": "..."}'.
+ *
+ *    e. The application is responsible for renewing the access token as needed
+ *       using the refreshing token.
```
lpsinger, line 152: @jnation3406, @cmccully, these are the instructions for Hermes. Do you see any issues with these?

jnation3406: What does step A / "do the [authorization code flow]" mean for us? I.e. is there an endpoint we hit with user provided credentials (username/password? gcn api token?) that then returns their bearer token? Or Is this a step the user does on the GCN website, i.e. login with GCN's IdP and then generate a bearer / refresh token and then they copy those into their hermes account profile to enable GCN submission?

lpsinger: The third party application (i.e., HERMES, in this case), will initiate an [OpenID Connect authorization code flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow) with our IdP. This is exactly the same procedure that you do with CILogon for single sign on.

[Authlib Web OAuth Clients](https://docs.authlib.org/en/latest/client/frameworks.html) should mostly automate this for you with Django.

jnation3406: Okay I read up on authorization code flow and it looks like that requires the browser and user involvement (clicking)... Not ideal for api based usage in hermes. I guess we could force users to go through the flow in the hermes-frontend one time and then store their bearer/refresh tokens in their profile forever after that. Do the refresh tokens expire or will they be valid forever - just want to make sure users dont need to continually go to the hermes frontend to "reauthorize" the gcn credentials...

lpsinger: Exactly, your user would sign into GCN once, and then HERMES would store the refresh token (not the access token) in its database row associated with the user.

According to [our IdP's documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-refresh-token.html) on refresh tokens:

> You can use the refresh token to retrieve new ID and access tokens. By default, the refresh token expires 30 days after your application user signs into your user pool. When you create an application for your user pool, you can set the application's refresh token expiration to any value between 60 minutes and 10 years.

lpsinger: On the lifetime of the refresh token, GCN _might_ be bound by the [NIST Digital Identity Guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html#aal1reauth) which I interpret as limiting the lifetime of refresh token to 30 days:

> 4.1.3 Reauthentication
>
> Periodic reauthentication of subscriber sessions SHALL be performed as described in [Section 7.2](https://pages.nist.gov/800-63-3/sp800-63b.html#sessionreauthn). At AAL1, reauthentication of the subscriber SHOULD be repeated at least once per 30 days during an extended usage session, regardless of user activity. The session SHOULD be terminated (i.e., logged out) when this time limit is reached.

Certainly that's how we have it set up for the GCN web app itself.

jnation3406: uggg. Once a year is probably fine but once every 30 days is going to be pretty annoying for users. Most users that want to submit to gcn through hermes will be doing so from the TOM programmatically, through the api. So remembering to go to the hermes UI once every 30 days to 'reauthenticate' is going to be a problem for them

jnation3406: Right now once a user registers with you and register's their email address, they can submit to the GCN listserv forever without reauthenticating every 30 days, right?

lpsinger: > uggg. Once a year is probably fine but once every 30 days is going to be pretty annoying for users. Most users that want to submit to gcn through hermes will be doing so from their TOM programmatically, through the api. So remembering to go to the hermes UI once every 30 days to 'reauthenticate' is going to be a problem for them

That sounds like it could be addressed through good UX in both HERMES and the TOM Toolkit. How do TOM Toolkit users authenticate to HERMES? Are you not using OIDC for that as well?

lpsinger: GCN Circular submission by email is a legacy feature and it is far less secure than going through the web application. It's not suitable for delegated submission by third parties.

jnation3406: Hermes has an api token associated with a users account. All users must do is send their api token with a request to authenticate it as that user. They can revoke the token and get a new one if it is compromised. To initially create a hermes account you have to go through the scimma/cilogin oidc stuff one time, but after that there is an api token that is used for all hermes programatic access. For web access, the oidc login is used to authenticate the user there every time (well it expires with the session every 2 weeks or so).

jnation3406: So to finish answering your question, TOM users store their api_token in their TOM's deployment secrets which get into the settings of the django app, where it is used for requests to hermes. API tokens are the same way the TOM authenticates with many other services it supports for submitting observing requests or retrieving targets or data, so it seems like its a well supported model for programmatic access over an API.

lpsinger: I'll talk to my colleagues here at NASA and see if I can find more nuanced guidance around long-lived refresh tokens. I agree, 30 days seems too short. Can you work with this for now, though?

jnation3406: Yeah I think this is a workable solution with longer-lived refresh tokens. Thanks for implementing it so quickly!

dakota002 (approved):

Courey (approved):

PR #1471, opened 2023-10-02
Circulars filter by date button

lpsinger:
Screenshot 2023-10-02 at 13 40 02


dakota002 (approved):

tylerbarna:
currently trying to review it, but getting a lambda init error

tylerbarna (approved):
so is the intent that we'll have a card object or something along those lines expand when the expanded bool is set to true? Would we still be using Dakota's details dropdown menu?

lpsinger:
> so is the intent that we'll have a card object or something along those lines expand when the expanded bool is set to true? Would we still be using Dakota's details dropdown menu?

Yes.

PR #1479, opened 2023-10-04
improvements to the search functionality

Courey:
PR 2 of 3, but this can go out independent of the other work as it is just an improvement to the existing search functionality.

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -113,14 +113,13 @@ export async function search({
     : { circularId: { order: 'desc' } }
 
   const [startTime, endTime] = getValidDates(startDate, endDate)
-
   const esQuery = query
     ? {
         function_score: {
           query: isLuceneSyntax
             ? {
                 query_string: {
-                  query,
+                  query: `${query}~`,
                   fuzziness: 'AUTO',
                 },
               }
@@ -129,7 +128,7 @@ export async function search({
                   must: [
                     {
                       query_string: {
-                        query,
+                        query: `*${query}*`,
                         default_field: '*',
                         fuzziness: 'AUTO',
                       },
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -130,6 +130,7 @@ export async function search({
                       query_string: {
                         query,
                         default_field: '*',
+                        fuzziness: 'AUTO',
                       },
                     },
                     {
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -121,6 +121,7 @@ export async function search({
             ? {
                 query_string: {
                   query,
+                  fuzziness: 'AUTO',
                 },
               }
             : {
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -145,19 +145,19 @@ export async function search({
               },
           functions: [
             {
-              filter: { exists: { field: 'eventId' } },
+              filter: { wildcard: { eventId: `*${query}*` } },
               weight: 4,
             },
             {
-              filter: { exists: { field: 'synonyms' } },
+              filter: { wildcard: { synonyms: `*${query}*` } },
               weight: 3,
             },
             {
-              filter: { exists: { field: 'subject' } },
+              filter: { wildcard: { subject: `*${query}*` } },
               weight: 2,
             },
             {
-              filter: { exists: { field: 'body' } },
+              filter: { wildcard: { body: `*${query}*` } },
               weight: 1,
             },
           ],
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -107,39 +107,42 @@ export async function search({
   totalItems: number
 }> {
   const client = await getSearch()
+  const isLuceneSyntax = query ? query.includes(':') : false
+  const sortOrder = query
+    ? { _score: 'desc' }
+    : { circularId: { order: 'desc' } }
 
   const [startTime, endTime] = getValidDates(startDate, endDate)
 
   const esQuery = query
     ? {
         function_score: {
-          query: {
-            bool: {
-              must: [
-                {
-                  query_string: {
-                    query: `*${query}*`,
-                    fields: [
-                      'eventId^4',
-                      'synonyms^3',
-                      'subject^2',
-                      'body^1',
-                      'submitter',
-                    ],
-                    fuzzy_max_expansions: 50,
-                  },
+          query: isLuceneSyntax
+            ? {
+                query_string: {
+                  query,
                 },
-                {
-                  range: {
-                    createdOn: {
-                      gte: startTime,
-                      lte: endTime,
+              }
+            : {
+                bool: {
+                  must: [
+                    {
+                      query_string: {
+                        query,
+                        default_field: '*',
+                      },
                     },
-                  },
+                    {
+                      range: {
+                        createdOn: {
+                          gte: startTime,
+                          lte: endTime,
+                        },
+                      },
+                    },
+                  ],
                 },
-              ],
-            },
-          },
+              },
           functions: [
             {
               filter: { exists: { field: 'eventId' } },
@@ -179,7 +182,7 @@ export async function search({
       query: esQuery,
       fields: ['subject'],
       _source: false,
-      sort: { _score: 'desc' },
+      sort: sortOrder,
       from: page && limit && page * limit,
       size: limit,
       track_total_hits: true,
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -110,6 +110,62 @@ export async function search({
 
   const [startTime, endTime] = getValidDates(startDate, endDate)
 
+  const esQuery = query
+    ? {
+        function_score: {
+          query: {
+            bool: {
+              must: [
+                {
+                  query_string: {
+                    query: `*${query}*`,
+                    fields: [
+                      'eventId^4',
+                      'synonyms^3',
+                      'subject^2',
+                      'body^1',
+                      'submitter',
+                    ],
+                    fuzzy_max_expansions: 50,
+                  },
+                },
+                {
+                  range: {
+                    createdOn: {
+                      gte: startTime,
+                      lte: endTime,
+                    },
+                  },
+                },
+              ],
+            },
+          },
+          functions: [
+            {
+              filter: { exists: { field: 'eventId' } },
+              weight: 4,
+            },
+            {
+              filter: { exists: { field: 'synonyms' } },
+              weight: 3,
+            },
+            {
+              filter: { exists: { field: 'subject' } },
+              weight: 2,
+            },
+            {
+              filter: { exists: { field: 'body' } },
+              weight: 1,
+            },
+          ],
+          score_mode: 'sum',
+          boost_mode: 'sum',
+        },
+      }
+    : {
+        match_all: {},
+      }
+
   const {
     body: {
       hits: {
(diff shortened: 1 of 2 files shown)


app/routes/circulars/circulars.server.ts
```diff
@@ -110,6 +110,62 @@ export async function search({
...
   const [startTime, endTime] = getValidDates(startDate, endDate)
 
+  const esQuery = query
+    ? {
+        function_score: {
+          query: {
+            bool: {
+              must: [
+                {
+                  query_string: {
```
lpsinger, line 120: According to https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax, `query_string` activates Lucene query syntax, like `author:"John Smith"`. That's probably a very good thing, but the UI should explain the supported syntax. Also, do we want Lucene syntax always on?

app/routes/circulars/circulars.server.ts
```diff
@@ -110,6 +110,62 @@ export async function search({
...
+                    query: `*${query}*`,
```
lpsinger, line 121: The wildcards before and after will do unexpected and unpredictable things if `query` actually consists of multiple terms or Lucene syntax.

app/routes/circulars/circulars.server.ts
```diff
@@ -110,6 +110,62 @@ export async function search({
...
+              filter: { exists: { field: 'subject' } },
+              weight: 2,
+            },
+            {
+              filter: { exists: { field: 'body' } },
+              weight: 1,
+            },
+          ],
+          score_mode: 'sum',
+          boost_mode: 'sum',
```
lpsinger, line 162: Please explain what this does. I can't find it in the documentation.

app/routes/circulars/circulars.server.ts
```diff
@@ -120,33 +176,10 @@ export async function search({
...
-      },
+      query: esQuery,
       fields: ['subject'],
       _source: false,
-      sort: {
-        circularId: {
-          order: 'desc',
-        },
-      },
+      sort: { _score: 'desc' },
```
lpsinger, line 182: If there is no query text, then the matches should be in descending chronological order.

If there is a query text, then it's OK to rank them by descending relevance.

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
Courey: If you look at the next level up, you will see that it is part of the function_score. function_score can take an array called functions. functions are applied to each result. It always has two properties in this case, filter and weight. Filter sets a filter that determines whether or not the function should be applied and weight assigns a weight to the function. This ranks the importance of the function being applied. The higher the weight, the more importance. These are then used to calculate the final relevance score. If it has a higher weight, that contributes more to the score.

score_mode specifies how the score should be calculated to come up with the final relevance score. So all the scores from all the functions are added to make the final_score. 

boost_mode specifies how the scores should be combined. Since I am using sum, the individual function scores are summed. 
    
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html
https://opster.com/guides/elasticsearch/search-apis/elasticsearch-function-score/

app/routes/circulars/circulars.server.ts
Courey: If it were not always on, on what criteria would we determine when it should be applied?

Courey: would this be sufficient?
const isLuceneSyntax = query.includes(':')

app/routes/circulars/circulars.server.ts
Courey: got this.

app/routes/circulars/circulars.server.ts
Courey: accounted for this

app/routes/circulars/circulars.server.ts
Courey: I did change this, but leaving this comment open to make sure that the check I put in place is acceptable. I ran three tests:
1. I used lucene syntax
2. I used a search phrase with multiple words that occurred together
3. I used multiple search terms that did not appear in the same place in the text body

lpsinger: > would this be sufficient? const isLuceneSyntax = query.includes(':')

Well, those wildcards wouldn't do anything if the user employs Lucene syntax. And they still won't do what you intend (which is match partial words) if the user inputs multiple terms separated by spaces.

Does OpenSearch have any query mode that does partial word matches?

app/routes/circulars/circulars.server.ts
lpsinger: I think that what you intend here is to apply a given weight depending on whether the search term matches the body, or the subject, or the event ID, etc. But is that what this actually does? As I read it, I would guess that it adds the weight if the _matching document_ contains a body, or a subject, or an event ID, etc. --- and most documents will have all of those fields.

app/routes/circulars/circulars.server.ts
Courey: https://opensearch.org/docs/2.1/opensearch/ux/
https://opensearch.org/docs/latest/query-dsl/term/fuzzy/

lpsinger: As I understand the documentation on [Lucene query syntax](https://opensearch.org/docs/latest/query-dsl/full-text/index/#query-string), you don't need wildcards in order to perform fuzzy matches. You can do that just with the [fuzzy query options](https://opensearch.org/docs/latest/query-dsl/full-text/index/#fuzzy-query-options).

app/routes/circulars/circulars.server.ts
```diff
@@ -107,8 +107,68 @@ export async function search({
   totalItems: number
 }> {
   const client = await getSearch()
+  const isLuceneSyntax = query ? query.includes(':') : false
```
lpsinger, line 110: That's not sufficient to detect Lucene syntax. Even without colons, the query string may contain boolean operators (`AND`, `OR`) or parentheses.

I suggest that for simplicity and a consistent user experience, Lucene query syntax is either always on or always off. I suggest always on, because it's a nice feature to have!

app/routes/circulars/circulars.server.ts
```diff
@@ -107,8 +107,68 @@ export async function search({
...
+            : {
+                bool: {
+                  must: [
+                    {
+                      query_string: {
+                        query: `*${query}*`,
+                        default_field: '*',
+                        fuzziness: 'AUTO',
+                      },
+                    },
```
lpsinger, line 135: Why do we need to insert wildcards into the query string? Can we not accomplish what we want by using the [fuzziness options](https://opensearch.org/docs/latest/query-dsl/full-text/index/#fuzzy-query-options)?

app/routes/circulars/circulars.server.ts
```diff
@@ -107,8 +107,68 @@ export async function search({
...
+              weight: 2,
+            },
+            {
+              filter: { wildcard: { body: `*${query}*` } },
+              weight: 1,
+            },
+          ],
+          score_mode: 'sum',
+          boost_mode: 'sum',
+        },
```
lpsinger, line 167: Are we sure that the Lucene query mode does not support boosted lists of fields, as in `"fields": ["title^4", "plot"]`? See https://opensearch.org/docs/latest/query-dsl/full-text/index/#multi-match.

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
Courey: Screenshot 2023-10-11 at 12 57 43 PM
This is removing wildcards. the above is without lucene syntax and has the exact same results with lucene syntax. 
compare to the images I commented above and you will see that with wildcards that returned several hits. I felt like that was a highly requested feature.

Courey: Screenshot 2023-10-11 at 1 02 07 PM

app/routes/circulars/circulars.server.ts
Courey: it's not because it's a lucene query, it's because it's a function_score. I could not use a function_score.

lpsinger: What does `function_score` have to do with the `fields` property?

app/routes/circulars/circulars.server.ts
Courey: ok, added a regex that gets all these conditions:
: 
There should be text on either side of the :
Example:
Subject: something

AND
There should be text on either side of the AND
Example:
Dogs AND cats

OR
There should be text on either side of the OR
Example:
Dogs OR cats

?
This can occur anywhere in the query
Examples:
Thi?s
This?
?this

*
This can occur anywhere in the query
Examples:
thi*
*this
th*s

~
This symbol will have text before it
Example:
Term~

[TO]
This will have [ then some text and then TO and then more text and then ]
Example:
Date: [ some string TO some string]

+
This will have text following it
Example:
+dogs

NOT
This will have text before and after NOT
Example:
Dogs NOT cats

-
This will have text following it
Examples:
Dogs -cats
-cats bowls

()
This will have text inside the parens
Example:
(Something will be in here)


^, which can be supported as part of lucene syntax is taken care of with our weighting. This is not a supported search character.

app/routes/circulars/circulars.server.ts
lpsinger: @Courey, the reason that "230802" doesn't match what you expect it to may be that it is being treated as a GCN Circular ID here:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/2ad4607f922529d7fa2a2867c99e1d0e11467aff/app/routes/circulars._archive._index.tsx#L41

and here:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/circulars/circulars.server.ts#L251-L264

app/routes/circulars/circulars.server.ts
lpsinger: As I said, I suggest that we either have Lucene syntax always on or always off.

lpsinger: I'm now convinced that it _does_ make sense to preprocess the query string. However, I think it has to be a bit more subtle:

- The query string may consist of multiple words (such as `x-shooter GRB follow-up`). If so, then we need to correctly insert wildcards, fuzzy modifiers, etc. for each term.
- If a term consists _only_ of digits, then we don't want to fuzz it.

I think that it will help to use one of the several [Lucene query syntax parser/formatters on NPM](https://www.npmjs.com/search?q=lucene) to tokenize the query string.

app/routes/circulars/circulars.server.ts
Courey: nothing, the function_score has the functions property and is how it ranks results. that's why the ranking is done in functions here. I could remove the function score

app/routes/circulars/circulars.server.ts
Courey: how should people search for an event by numbers as requested without it assuming it's a circular id? Do people use the GCN circular id more than the event id? I think that the results that someone would expect if they typed 1234 in the search bar would be to get results that contain 1234, not circular 1234. But I also don't know how people typically use this, so the gcn circular id could be what people are searching for.

Courey: also, searching for a year would bring up a circular then. like if I wanted to search for results in 2023 it would redirect me to circular 2023

Courey: for example, the naming convention of GRBs make it really easy to search for all GRBs in august of this year. I can get all the circulars for august by searching 2308
Screenshot 2023-10-13 at 1 41 07 PM

lpsinger: > how should people search for an event by numbers as requested without it assuming it's a circular id? Do people use the GCN circular id more than the event id?

Without user testing, I do not know the answer to that.

app/routes/circulars/circulars.server.ts
lpsinger: The `function_score` method seems more verbose than field boosting method. Can you use the field boosting method?

Issue #1366, opened 2023-09-11
Footer draws over floating elements


tylerbarna:
This might be specific to the [detailsDropdownButton](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/components/DetailsDropdownButton.tsx)/[detailsDropdownContent](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/components/DetailsDropdownContent.tsx), but if a page is short enough, the footer will draw over the floating element. An example of this can be seen if there are no results in the circulars page and the date range selector dropdown (currently #1265) is expanded:
![image](https://github.com/nasa-gcn/gcn.nasa.gov/assets/57914086/bcb9a322-8f3a-4eb8-baad-4577e90fb7d5)

For context, this is the full dropdown as is currently implemented in that pr:
![image](https://github.com/nasa-gcn/gcn.nasa.gov/assets/57914086/d2dee742-2517-4f15-bfad-1babac61fe7e)

lpsinger:
I'm not sure this _is_ a bug. You have full control of the z-order of your new control, don't you?

tylerbarna:
> I'm not sure this _is_ a bug. You have full control of the z-order of your new control, don't you?

This might be an inheritance issue, but setting the z-order of the details drop-down wasn't setting it above the footer. May have to experiment with where the z-index is set

lpsinger:
@tylerbarna, I think you mentioned you found a workaround for this. Can we close it?

tylerbarna:
> @tylerbarna, I think you mentioned you found a workaround for this. Can we close it?

Ah yes, it's fixed by adding `z-top` to the search button group 
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/circulars._archive._index.tsx#L263

I have this change present in my date selector PR, but should I submit it as a separate PR?

lpsinger:
> I have this change present in my date selector PR, but should I submit it as a separate PR?

No.

PR #1482, opened 2023-10-05
Implement Circular Archive date range search frontend with new DateSelectorButton

tylerbarna:
This is a PR to replace #1265 to bring it to a more recent commit and also to make use of the recent Date selector button added by @lpsinger in #1471. The appearance is now as such:

image
image

Known issues:

- [x] Date dropdown menu remains expanded after page is reloaded following a selection and submission of one of the date options. However, the div that contains the date range selector element is collapsed upon page submission and reloading.

- [x] Radio buttons are misaligned when viewed on mobile or when a desktop window is made sufficiently narrow
image

- [x] Selecting an option, submitting and then switching back to the all time option before submitting again results in the button label being listed as 'undefined-now' rather than the default 'filter by date' 
image

- [x] When selecting a custom date range, if the start date is set, the results are cleared from the page due to the `clean` variable being set to false (related: #1110, #1366), resulting in the footer obscuring the end date and submit buttons 
image

- [x] The buttons to download circulars as a tarball are set to a higher z-index than the date range dropdown 
image

- [ ] On mobile, most of the vertical space is occupied by the date range menu when the custom range option is selected (example is an iPhone 12 Pro) 
image

- [ ] On mobile, smaller screens may not be able to use the date selector for the end date due to screen height limitations. This might be an issue with the DateRangeSelector element. (example is an iPhone SE) 
image

- [ ] start and end date values are not shown in input fields after setting them and submitting them (they are set as parameters, but they are not visually shown in the input fields). Having a UseEffect set the values of the inputs results in the dates not changing visually after selecting a new date
image
- [x] setting a custom date range and then submitting will result in custom range radio button being selected but the showDateRange div not being expanded 
image

- [x] selecting a fuzzy date (eg last year), submitting, and then switching to only submitting a custom end date will result in the startDate parameter persisting as the prior value when it should be undefined


app/routes/circulars._archive._index.tsx
```diff
@@ -239,10 +242,40 @@ export default function () {
...
+  const [showDateRange, setShowDateRange] = useState(false)
+  const clean =
+    inputQuery === query &&
+    inputDateGte === startDate &&
+    inputDateLte === endDate
 
   const submit = useSubmit()
 
+  const setFuzzyTime = (startDate?: string) => {
+    // console.log(startDate)
```
lpsinger, line 257:
```suggestion
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
           </Button>
         </Form>
         {featureCircularsFilterByDate && (
-          <DateSelectorButton startDate={startDate} endDate={endDate} />
+          <>
+            <DateSelectorButton
+              startDate={startDate}
+              endDate={endDate}
+              onClick={() => setShowContent(!showContent)}
```
lpsinger, line 331:
```suggestion
              onClick={() => { setShowContent((shown) => !shown) }}
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
...
+                        id="radio-alltime"
+                        name="radio-date"
+                        value="undefined"
+                        defaultChecked={true}
+                        label="All Time"
+                        className="usa-search__filter-button"
+                        onChange={(e) => {
+                          setFuzzyTime(e.target.value)
+                        }}
+                      ></Radio>
```
lpsinger, line 349: Use self-closing element syntax:

```suggestion
                     / >
```

and the same for the other radio buttons.

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
...
+                        name="radio-date"
+                        value="today"
+                        label="Today"
+                        className="usa-search__filter-button"
+                        onChange={(e) => {
+                          setFuzzyTime(e.target.value)
+                        }}
+                      ></Radio>
+                    </div>
+                    <div className="display-flex flex-column margin-1">
```
lpsinger, line 371: You should be able to use the `<Grid>` component to structure this in such a way that you don't need to manually break the radio buttons into rows.

lpsinger (changes requested):
Can you auto-submit the form if any of the fuzzy date radio buttons are clicked?

lpsinger:
> * [ ]  On mobile, smaller screens may not be able to use the date selector for the end date due to screen height limitations. This might be an issue with the DateRangeSelector element. (example is an iPhone SE)

I agree. I think that we may handle this by displaying the dropdown content as a block element on mobile and as a dropdown on desktop. But don't worry about that now; we can fix that in a future PR.

lpsinger:
BTW, this looks very nice!

lpsinger:
> > Can you auto-submit the form if any of the fuzzy date radio buttons are clicked?
> 
> So that was the original intent when I was originally using buttons rather than radios, but I was encountering a weird error where the Parameters were delayed in being set by one cycle (eg if on submission 1, I set the params to time A, the url would still be blank, but upon trying to set them to time B on submission 2, the url params would be set to time A). @Courey and I were unable to debug that when we initially tried, which is why I added the submit button

Alright, we can tackle that in a future PR.

app/routes/circulars._archive._index.tsx
tylerbarna: implemented in [8a26a4e](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482/commits/8a26a4ee4ab4d1e98fc5a50458438633f7ddf4fb)

app/routes/circulars._archive._index.tsx
tylerbarna: @lpsinger I've been experimenting with this, and I've found that enclosing them into a single grid component doesn't result in the 3x3 of radio buttons I was aiming for. However, I tried to do a similar thing to the responsive layout in the react components library (see [here](https://trussworks.github.io/react-uswds/?path=/docs/components-grid--responsive)), which resulted in something like this:
image

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
...
+                <div className="usa-radio maxw-card-xlg">
+                  <div className="display-flex flex-row">
+                    <div className="display-flex flex-column flex-align-start margin-1">
+                      <Radio
+                        id="radio-alltime"
+                        name="radio-date"
+                        value="undefined"
+                        defaultChecked={true}
+                        label="All Time"
+                        className="usa-search__filter-button"
```
lpsinger, line 345: Remove for all radio buttons. This is not a [USWDS Search component](https://designsystem.digital.gov/components/search/).

```suggestion
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
...
+                      startDateLabel="Start Date"
+                      className="margin-bottom-2"
+                      startDatePickerProps={{
+                        id: 'event-date-start',
+                        name: 'event-date-start',
+                        defaultValue: 'startDate',
+                        onChange: (value) => {
+                          setInputDateGte(value)
+                        },
+                        // style: { height: '15px' }, // Adjust the height as needed
```
lpsinger, line 450: Remove commented out code. Same for all other radio buttons.

```suggestion
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +324,160 @@ export default function () {
...
+          <>
+            <DateSelectorButton
+              startDate={startDate}
+              endDate={endDate}
+              onClick={() => { setShowContent((shown) => !shown) }}
+              expanded={showContent}
+            />
+            {showContent && (
+              <DetailsDropdownContent className="margin-top-1">
+                <div className="usa-radio maxw-card-xlg">
```
lpsinger, line 336:
```suggestion
                <div className="maxw-card-xlg">
```

lpsinger (changes requested):

app/routes/circulars._archive._index.tsx
tylerbarna: ah, must've been vestigial from when I was trying to use buttons in the dropdown

app/routes/circulars._archive._index.tsx
```diff
@@ -239,10 +243,55 @@ export default function () {
...
+        radioSelected = true
+      }
+    })
+    if (!radioSelected) {
+      const customButton = document.getElementById('radio-custom')
+      if (customButton) {
+        ;(customButton as HTMLInputElement).checked = true
+      } // does not address the need for the custom date range button to be expanded if this button is checked
+    }
+  })
```
lpsinger, line 293: No, don't use `useEffect` here. You can generate the value for `defaultChecked` from the search params.

tylerbarna: I had thought defaultChecked was just the way to specify with radio to have selected by default, is that not the case?

I had tried doing something like `checked={isChecked(buttonValue)}` for each of the radio buttons, where `isChecked()` passed the specific button value to a function and checked if it was the same as the search param. This worked for having it show the correct radio button as selected after submitting, but it wouldn't update to reflect choosing a new radio button prior to submission. 

In this example, the page has been loaded with the radio button selected for the last day. Upon trying to click on the option for the last year, it shows that the radio has been selected, but the last year button doesn't actually show as being checked until you submit and reload the page:
image

lpsinger: > I had thought defaultChecked was just the way to specify with radio to have selected by default, is that not the case?

Correct.

> I had tried doing something like `checked={isChecked(buttonValue)}` for each of the radio buttons, where `isChecked()` passed the specific button value to a function and checked if it was the same as the search param. This worked for having it show the correct radio button as selected after submitting, but it wouldn't update to reflect choosing a new radio button prior to submission.

You generally should _not_ explicitly set the value or the checked state of controls that the user is interacting with.

lpsinger:
@tylerbarna, I have played around with this locally and I see the z-ordering glitches in the date selector that you were describing. I haven't come up with a way to fix that. I'd really like to wrap this up and free you to work on something new. Could you pare this down to a PR that _just_ supports the fuzzy date ranges and no custom date ranges?

tylerbarna:
> @tylerbarna, I have played around with this locally and I see the z-ordering glitches in the date selector that you were describing. I haven't come up with a way to fix that. I'd really like to wrap this up and free you to work on something new. Could you pare this down to a PR that _just_ supports the fuzzy date ranges and no custom date ranges?

@lpsinger oh, apologies for not including this in a message, but @Courey was actually able to resolve the z-order issues I was referring to by setting the z-index to z-top for the button group. Are you referring to a different index issue?

lpsinger:
> > @tylerbarna, I have played around with this locally and I see the z-ordering glitches in the date selector that you were describing. I haven't come up with a way to fix that. I'd really like to wrap this up and free you to work on something new. Could you pare this down to a PR that _just_ supports the fuzzy date ranges and no custom date ranges?
> 
> @lpsinger oh, apologies for not including this in a message, but @Courey was actually able to resolve the z-order issues I was referring to by setting the z-index to z-top for the button group. Are you referring to a different index issue?

Oh, I see! Very nice.

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +340,152 @@ export default function () {
...
+              endDate={endDate}
+              onClick={() => {
+                setShowContent((shown) => !shown)
+                setShowDateRange(false)
+              }}
+              expanded={showContent}
+            />
+            {showContent && (
+              <DetailsDropdownContent className="margin-top-1">
+                <div className="maxw-card-xlg">
```
lpsinger, line 355:
```suggestion
              <DetailsDropdownContent className="maxw-card-xlg">
                <CardBody>
```

- Remove explicit margin-top; see #1508.
- Enclose content in CardBody.

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +340,152 @@ export default function () {
...
+                      className=""
+                      form="searchForm"
+                      onClick={() => {
+                        setDateRange()
+                      }}
+                    >
+                      <Icon.CalendarToday /> Submit
+                    </Button>
+                  </div>
+                </div>
```
lpsinger, line 485:
```suggestion
                </CardBody>
                <CardFooter>
                  <Button
                    type="button"
                    className=""
                    form="searchForm"
                    onClick={() => {
                      setDateRange()
                    }}
                  >
                    <Icon.CalendarToday /> Submit
                  </Button>
                </CardFooter>
```

app/routes/circulars._archive._index.tsx
lpsinger: Yes, that's what I'm suggesting:

```jsx
<Grid row>
  <Grid col={4}
    <Radio ... />
  </Grid>
  <Grid col={4}
    <Radio ... />
  </Grid>
  <Grid col={4}
    <Radio ... />
  </Grid>
  <Grid col={4}
    <Radio ... />
  </Grid>
  ...
</Grid>
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +340,152 @@ export default function () {
...
+                      />
+                      <Radio
+                        id="radio-today"
+                        name="radio-date"
+                        value="today"
+                        label="Today"
+                        onChange={(e) => {
+                          setFuzzyTime(e.target.value)
+                        }}
+                      />
```
lpsinger, line 385: To reduce duplication of code, please generate all of these radio buttons for the fuzzy date ranges by looping over `dateSelectorLabels`:

```jsx
{Object.entries(dateSelectorLabels).map(([key, label]) => <Radio ... />)}

**Review by lpsinger — CHANGES_REQUESTED** [review]

File: app/routes/circulars._archive._index.tsx (hunk 10 above)
Comment (tylerbarna) [reply app/routes/circulars._archive._index.tsx]: implemented this in 53a30ab, though I did notice that the submit button is slightly further right now 
image

File: app/routes/circulars._archive._index.tsx (hunk 11 above)
Comment (tylerbarna) [reply app/routes/circulars._archive._index.tsx:387]: > To reduce duplication of code, please generate all of these radio buttons for the fuzzy date ranges by looping over `dateSelectorLabels`:
> 
> ```js
> {Object.entries(dateSelectorLabels).map(([key, label]) => <Radio ... />)}
> ```

would I break dateSelectorLabels into sublists to account for them needing to be in different columns? I may be misunderstanding how the Grid object works though

Comment (tylerbarna) [reply app/routes/circulars._archive._index.tsx]: switched to using the Grid structure in 8c3016d, this is the result:
image

It actually fixes the issue I had described regarding the horizontal alignment of the radio buttons when on mobile, which is nice:
image

Comment (lpsinger) [reply app/routes/circulars._archive._index.tsx]: > would I break dateSelectorLabels into sublists to account for them needing to be in different columns?

No, it will wrap automatically.

**Comment by lpsinger** [comment]
@tylerbarna, there are now a large number of unrelated changes on this branch. It looks like the aftermath of a failed merge attempt. Please fix.

**Comment by tylerbarna** [comment]
> @tylerbarna, there are now a large number of unrelated changes on this branch. It looks like the aftermath of a failed merge attempt. Please fix.

apologies, must've happened when I tried merging the upstream changes to add the requested change in https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482#discussion_r1355582069, I'll get right on that

**Comment by tylerbarna** [comment]
b479e2d and d91d10b should fix that merge issue

File: app/routes/circulars._archive._index.tsx (hunk 11 above)
Comment (tylerbarna) [reply app/routes/circulars._archive._index.tsx]: > To reduce duplication of code, please generate all of these radio buttons for the fuzzy date ranges by looping over `dateSelectorLabels`:
> 
> ```js
> {Object.entries(dateSelectorLabels).map(([key, label]) => <Radio ... />)}
> ```

Should be added as of 88b8669

File: app/routes/circulars._archive._index.tsx [hunk 12]
```diff
@@ -291,7 +328,89 @@ export default function () {
...
+                          setInputDateLte(value)
+                        },
+                      }}
+                    />
+                  )}
+
+                  <CardFooter>
+                    <Button
+                      type="button"
+                      className=""
```
lpsinger, line 401:
```suggestion
```

app/routes/circulars._archive._index.tsx
```diff
@@ -291,7 +328,89 @@ export default function () {
...
+                            onChange={(e) => {
+                              if (option.value === 'custom') {
+                                setShowDateRange(e.target.checked)
+                              } else {
+                                setFuzzyTime(option.value)
+                              }
+                            }}
+                          />
+                        </Grid>
+                      ))}
```
lpsinger, line 370: Combine the two `.map` calls into one:

```suggestion
                      .map(([value, label]) => (
                        <Grid col={4} key={value}>
                          <Radio
                            id={`radio-${value}`}
                            name="radio-date"
                            value={value}
                            label={label}
                            checked={value === inputDateGte}
                            defaultChecked={value === 'alltime'}
                            onChange={(e) => {
                              if (value === 'custom') {
                                setShowDateRange(e.target.checked)
                              } else {
                                setFuzzyTime(option.value)
                              }
                            }}
                          />
                        </Grid>
                      ))}
```

lpsinger (changes requested):
The `'alltime'` and `undefined` keys in `dateSelectorLabels` result in a lot of branches. They are such special cases that I think it will be simpler if you remove those keys and handle them outside of the loop:

```jsx
<Grid row>
  <Grid col={4}>
    <Radio label="All Time" ... />
  </Grid>
  {Object.entries(dateSelectorLabels).map(([value, label]) => ...)}
  <Grid col={4}>
    <Radio label="Custom Range..." ... />
  </Grid>
</Grid>
```

tylerbarna:
> The `'alltime'` and `undefined` keys in `dateSelectorLabels` result in a lot of branches. They are such special cases that I think it will be simpler if you remove those keys and handle them outside of the loop:
> 
> ```js
> <Grid row>
>   <Grid col={4}>
>     <Radio label="All Time" ... />
>   </Grid>
>   {Object.entries(dateSelectorLabels).map(([value, label]) => ...)}
>   <Grid col={4}>
>     <Radio label="Custom Range..." ... />
>   </Grid>
> </Grid>
> ```

So I had originally set the value of the All Time radio button to be set to undefined so that it would default to displaying the all time label, but I ended up swapping the value to alltime so the radio mapping would line up; I mostly left 'undefined' there to cover any cases, but it should be fine to be removed. The custom button is actually set to the value 'custom'

I tried out removing the All Time label from dateSelectorLabels, but it causes the button labels to default to the behaviour of the parameters being two dates, so it ends up looking like:
image

app/routes/circulars._archive._index.tsx
tylerbarna: resolved in [3c05f3c](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482/commits/3c05f3cfd8ebc1718968e58f7c180ab0bb09bf67)

lpsinger:
> > The `'alltime'` and `undefined` keys in `dateSelectorLabels` result in a lot of branches. They are such special cases that I think it will be simpler if you remove those keys and handle them outside of the loop:
> > ```
> > <Grid row>
> >   <Grid col={4}>
> >     <Radio label="All Time" ... />
> >   </Grid>
> >   {Object.entries(dateSelectorLabels).map(([value, label]) => ...)}
> >   <Grid col={4}>
> >     <Radio label="Custom Range..." ... />
> >   </Grid>
> > </Grid>
> > ```
> 
> So I had originally set the value of the All Time radio button to be set to undefined so that it would default to displaying the all time label, but I ended up swapping the value to alltime so the radio mapping would line up; I mostly left 'undefined' there to cover any cases, but it should be fine to be removed. The custom button is actually set to the value 'custom'
> 
> I tried out removing the All Time label from dateSelectorLabels, but it causes the button labels to default to the behaviour of the parameters being two dates, so it ends up looking like: image

Probably because you also have special cases in your onClick handlers.

tylerbarna:
New edge case:

selecting a fuzzy date (eg last year), submitting, and then switching to only submitting a custom end date will result in the startDate parameter persisting as the prior value when it should be undefined

tylerbarna:
> > > The `'alltime'` and `undefined` keys in `dateSelectorLabels` result in a lot of branches. They are such special cases that I think it will be simpler if you remove those keys and handle them outside of the loop:
> > > ```
> > > <Grid row>
> > >   <Grid col={4}>
> > >     <Radio label="All Time" ... />
> > >   </Grid>
> > >   {Object.entries(dateSelectorLabels).map(([value, label]) => ...)}
> > >   <Grid col={4}>
> > >     <Radio label="Custom Range..." ... />
> > >   </Grid>
> > > </Grid>
> > > ```
> > 
> > 
> > So I had originally set the value of the All Time radio button to be set to undefined so that it would default to displaying the all time label, but I ended up swapping the value to alltime so the radio mapping would line up; I mostly left 'undefined' there to cover any cases, but it should be fine to be removed. The custom button is actually set to the value 'custom'
> > I tried out removing the All Time label from dateSelectorLabels, but it causes the button labels to default to the behaviour of the parameters being two dates, so it ends up looking like: image
> 
> Probably because you also have special cases in your onClick handlers.

Isn't it because the DateSelectorButton is defined such that the startDate variable needs to be defined in dateSelectorLabels in order for it to not default to displaying {startDate}-{endDate}?
https://github.com/nasa-gcn/gcn.nasa.gov/blob/8d85b362777267d02fa9aace4207847a8c32b497/app/routes/circulars._archive._index.tsx#L204

lpsinger:
> Isn't it because the DateSelectorButton is defined such that the startDate variable needs to exist in dateSelectorLabels in order for it to not default to displaying {startDate}-{endDate}?

Right. So for those radio buttons for which you do _not_ want to show a custom date range, set their value to an empty string (`""`).

tylerbarna:
> > Isn't it because the DateSelectorButton is defined such that the startDate variable needs to exist in dateSelectorLabels in order for it to not default to displaying {startDate}-{endDate}?
> 
> Right. So for those radio buttons for which you do _not_ want to show a custom date range, set their value to an empty string (`""`).

Ah, I see

app/routes/circulars._archive._index.tsx
```diff
@@ -176,14 +182,14 @@ function Pagination({
...
-  year: 'Last year',
+  hour: 'Last Hour',
   today: 'Today',
-  mtd: 'Month to date',
-  ytd: 'Year to date',
+  day: 'Last Day',
+  week: 'Last Week',
+  month: 'Last Month',
+  year: 'Last Year',
+  // mtd: 'Month to date',
```
lpsinger, line 191: Remove commented out code...

lpsinger (changes requested):

tylerbarna:
I think the component now meets most of the specifications we were aiming for

app/routes/circulars._archive._index.tsx
```diff
@@ -239,10 +246,37 @@ export default function () {
...
+
+  const setDateRange = () => {
+    setShowContent(false)
+    const params = new URLSearchParams(location.search)
+    if (inputDateGte) params.set('startDate', inputDateGte)
+    else params.delete('startDate')
+    if (inputDateLte) params.set('endDate', inputDateLte)
+    else params.delete('endDate')
+    window.location.search = params.toString()
+  }
```
lpsinger, line 278: In React, we don't modify the location search params directly. We would employ the [`useSearchParams` hook](https://reactrouter.com/en/main/hooks/use-search-params).

However, you shouldn't need to do that either, because the user is submitting a form to perform the navigation, and _that_ updates the search params.

lpsinger (changes requested):

app/routes/circulars._archive._index.tsx
tylerbarna: ah, okay, I can revert it to the previous implementation then

lpsinger:
Please rebase. there are now merge conflicts due to #1524.

tylerbarna:
> Please rebase. There are now merge conflicts due to #1524.

should be rebased now!

app/routes/circulars._archive._index/route.tsx
```diff
@@ -75,14 +81,54 @@ export default function () {
...
+    } else {
+      setShowDateRange(false)
+      setInputDateGte(startDate)
+    }
+    setInputDateLte('')
+  }
+
+  const setDateRange = () => {
+    setShowContent(false)
+    const params = new URLSearchParams(location.search)
```
lpsinger, line 107: Reuse `searchParams`, defined above.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -75,14 +81,54 @@ export default function () {
...
+    if (startDate === 'custom') {
+      setShowDateRange(true)
+    } else {
+      setShowDateRange(false)
+      setInputDateGte(startDate)
+    }
+    setInputDateLte('')
+  }
+
+  const setDateRange = () => {
```
lpsinger, line 105:
```suggestion
  function setDateRange() {
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -75,14 +81,54 @@ export default function () {
...
+  const [showContent, setShowContent] = useState(false)
+  const [showDateRange, setShowDateRange] = useState(false)
+  const clean =
+    inputQuery === query &&
+    inputDateGte === startDate &&
+    inputDateLte === endDate
 
   const submit = useSubmit()
 
+  const setFuzzyTime = (startDate?: string) => {
```
lpsinger, line 95:
```suggestion
  function setFuzzyTime(startDate?: string) {
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -75,14 +81,54 @@ export default function () {
...
+    if (startDate === 'custom') {
+      setShowDateRange(true)
+    } else {
+      setShowDateRange(false)
+      setInputDateGte(startDate)
+    }
+    setInputDateLte('')
```
lpsinger, line 102: Factor this into two _different_ onClick handlers so that you don't need the `if (startDate === 'custom')` conditional logic.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -75,14 +81,54 @@ export default function () {
...
   const [inputQuery, setInputQuery] = useState(query)
-  const clean = inputQuery === query
+  const [inputDateGte, setInputDateGte] = useState(startDate)
+  const [inputDateLte, setInputDateLte] = useState(endDate)
+  const [showContent, setShowContent] = useState(false)
+  const [showDateRange, setShowDateRange] = useState(false)
+  const clean =
+    inputQuery === query &&
+    inputDateGte === startDate &&
+    inputDateLte === endDate
```
lpsinger, line 91: I suggest that we don't want to wipe away the search results if the date range is dirty (unsubmitted).

```suggestion
  const clean = inputQuery === query
```

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
tylerbarna: Ah, I actually think this might have been a holdover from when the radios were combined into one map and custom was one of the values it had to handle ([3c05f3c](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482/commits/3c05f3cfd8ebc1718968e58f7c180ab0bb09bf67)). `setFuzzyTime()` doesn't actually handle the custom date radio button any more. I removed the conditional logic, but the call to `setShowDateRange(false)` is still present in [c9266c7](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1482/commits/c9266c7804380f15401426ec8519481bdd2c6142) to handle the case of selecting the custom date range but switching to a fuzzy date prior to submitting.

lpsinger (approved):

PR #1481, opened 2023-10-05
Redirect on missing version and path

dakota002:
Closes #1450

PR #1483, opened 2023-10-05
Clean up URL routing for schema browser

lpsinger:
- Redirect from /docs/schema to /docs/schema/stable, from /docs/schema/stable to /docs/schema/<latest>, and from /docs/schema/$version to /docs/schema/$version/gcn/notices.
- The loader for /docs/schema can now be safely cached because its return value does not depend on the dynamic portions of the path from its child routes.
- The /docs/schema/<version> index route, which was only ever used as a component import, is gone.


dakota002 (approved):
Looks good, I think we should use this instead of #1481

PR #745, opened 2023-04-05
Add backend for event name synonyms synced with OpenSearch

lpsinger:
This adds the backend implementation for #611.

PR #1505, opened 2023-10-09
Base grouping functionality

Courey:
This is PR ~~3 of 3~~ 4 of 4

### DIFF
--- a/sandbox-seed.json
+++ b/sandbox-seed.json
@@ -5505,9 +5505,15 @@
     }
   ],
   "auto_increment_metadata": [
-    { "tableName": "circulars", "circularId": 34776 }
+    {
+      "tableName": "circulars",
+      "circularId": 34776
+    }
   ],
   "legacy_users": [
-    { "email": "example.receive@example.com", "receive": 1 },
+    {
+      "email": "example.receive@example.com",
+      "receive": 1
+    }
   ]
 }


app/routes/circulars._archive_index/Pagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+              return (
+                <li
+                  className="usa-pagination__item usa-pagination__overflow"
+                  role="presentation"
+                  key={i}
+                >
+                  <span>…</span>
+                </li>
+              )
+            case 'next':
```
Courey, line 86: I have found an issue with pagination on grouping page. looking into this

lpsinger (changes requested):
I like refactoring the (now fairly complicated) `circulars._archive._index.tsx` route into a route folder. Would you please split that into a separate PR?

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+                  {circulars.map(({ circularId, subject }, index) => (
+                    <li
+                      id={circularId.toString()}
+                      value={circularId}
+                      key={`li-${groupIndex}-${index}`}
+                      className="border-base-lighter"
+                    >
+                      <Link
+                        to={`/circulars/${circularId}`}
+                        key={`link-${groupIndex}-${index}`}
```
lpsinger, line 68:
```suggestion
```

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+              <details open={detailsToggle} key={`details-${groupIndex}`}>
+                <summary key={`summary-${groupIndex}`}>
+                  {circulars[0].synonyms?.join(', ')}
+                </summary>
+                <ol key={`ol-${groupIndex}`}>
+                  {circulars.map(({ circularId, subject }, index) => (
+                    <li
+                      id={circularId.toString()}
+                      value={circularId}
+                      key={`li-${groupIndex}-${index}`}
```
lpsinger, line 63:
```suggestion
                      key={index}
```

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </summary>
+        </details>
+        {totalItems > 0 &&
+          allItems.map(({ circulars }, groupIndex) => (
+            <>
+              <details open={detailsToggle} key={`details-${groupIndex}`}>
+                <summary key={`summary-${groupIndex}`}>
+                  {circulars[0].synonyms?.join(', ')}
+                </summary>
+                <ol key={`ol-${groupIndex}`}>
```
lpsinger, line 58:
```suggestion
                <ol>
```

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+            key={`open-close-summary`}
+          >
+            {expandAllText}
+          </summary>
+        </details>
+        {totalItems > 0 &&
+          allItems.map(({ circulars }, groupIndex) => (
+            <>
+              <details open={detailsToggle} key={`details-${groupIndex}`}>
+                <summary key={`summary-${groupIndex}`}>
```
lpsinger, line 55:
```suggestion
                <summary>
```

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+            className="text-base border-base-lighter"
+            key={`open-close-summary`}
+          >
+            {expandAllText}
+          </summary>
+        </details>
+        {totalItems > 0 &&
+          allItems.map(({ circulars }, groupIndex) => (
+            <>
+              <details open={detailsToggle} key={`details-${groupIndex}`}>
```
lpsinger, line 54:
```suggestion
              <details open={detailsToggle} key={groupIndex}>
```

app/routes/circulars._archive_index/GroupedView.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <summary
+            className="text-base border-base-lighter"
+            key={`open-close-summary`}
+          >
+            {expandAllText}
+          </summary>
+        </details>
+        {totalItems > 0 &&
+          allItems.map(({ circulars }, groupIndex) => (
+            <>
```
lpsinger, line 53: Remove unnecessary fragment.

app/routes/circulars/circulars.server.ts
```diff
@@ -262,3 +279,156 @@ export async function circularRedirect(query: string) {
...
+
+  const tableName = db.name('synonyms')
+
+  doc.put({
+    TableName: tableName,
+    Item: {
+      synonyms: synonyms.join(','),
+    },
+  })
+}
```
lpsinger, line 298:
```suggestion
  await db.synonyms.put({
    TableName: tableName,
    Item: {
      synonyms: synonyms.join(','),
    },
  })
}
```

lpsinger: According to the app.arc file, the partition key of this table is called `id`, but this field is not specified here.

lpsinger (changes requested):

Courey:
> I like refactoring the (now fairly complicated) `circulars._archive._index.tsx` route into a route folder. Would you please split that into a separate PR?

why a separate PR? it was the complexity increase in this PR that necessitated the breaking apart of the route. It was done as part of the campsite rule.

Courey:
Here it is on it's own. I still don't understand the why behind it. The other PR will have to go in first and that will bork this PR so I'll have to fix it once the other branch is in main. https://github.com/nasa-gcn/gcn.nasa.gov/pull/1524

lpsinger:
> Here it is on it's own. I still don't understand the why behind it. The other PR will have to go in first and that will bork this PR so I'll have to fix it once the other branch is in main. #1524

Because the refactoring comprises the bulk of this PR, because reviewing the refactoring separately allows us to get it onto the main branch sooner, because it reduces the cognitive load of reviewing the synonyms work, because we have other PRs in progress that will touch the same files.

Courey:
blocked by https://github.com/nasa-gcn/gcn.nasa.gov/pull/1524

app/table-streams/synonyms/index.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const handler = createTriggerHandler(
+  async ({ eventName, dynamodb }: DynamoDBRecord) => {
+    const id = unmarshallTrigger(dynamodb!.Keys).id as string
+    const promises = []
+
+    if (eventName === 'REMOVE') {
+      promises.push(removeIndex(id))
+    } /* (eventName === 'INSERT' || eventName === 'MODIFY') */ else {
```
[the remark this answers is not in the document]
lpsinger: No, this is OK. We have a similar comment in the other table stream handlers.

app/routes/circulars._archive._index/GroupedView.tsx
```diff
@@ -0,0 +1,68 @@
...
+          className="margin-bottom-2 margin-top-1"
+          aria-label="view all toggle"
+          onClick={() => {
+            setDetailsToggle(!detailsToggle)
+          }}
+        >
+          <summary className="text-base border-base-lighter">
+            {expandAllText}
+          </summary>
+        </details>
```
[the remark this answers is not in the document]
lpsinger: As far as I recall, `<details>`/`<summary>` supports expanding and collapsing 'out of the box' --- the browser does it for us, and we don't have to script it. Is the purpose here to support an "expand all" button?

Courey:
I was thinking through what you asked me for feedback and I have a question. You were asking how synonyms get linked. Currently the way they are linked is when someone creates a string list of eventIds on a specific circular. then when the grouping page is viewed, it gets all events with an eventId in that array of string synonyms. So if you have not marked circular 123 event abc as a synonym for the group "abc, xyz", it will still get pulled into that grouping as it has an eventId in that existing grouping. If the list already exists, it doesn't make a new one, it should just overwrite the old one or do nothing since there is no change.
This does make me feel like it's a little weird that the way a synonym is created is on a circular individually but it impacts all circulars with those ids. Would it be better to have an entirely separate page for moderators to just look up and make synonym links? I feel like it might make more sense that way.
if we do it that way, we don't have to put the synonyms on the circular, it would just be it's own table that leverages the eventId on the circular
actually, I think either way we could take the synonyms off the circulars

Courey:
# Group by event id
## tables
| synonyms  |
| --------- |
| id        |

| circulars |
| --------- |
| circularId|
| subject   |
| body      |
| submitter |
| eventId   |

## lifecycle of synonyms
- synonym lists are created by moderators
- they are a string of comma separated values of event ids
- they are stored in the id field of the synonyms table
- they can be created, edited, viewed, and deleted

### creation:
- format the array of synonym strings
    - this trims whitespace
    - this natural sorts the synonyms list so they are always in a predictable order
- create the synonym with the formatted, ordered, synonyms list.
- the list of event ids is combined into one comma separated string as it is a primary id so can not have an array of strings

```
  const synonymIdString = formatSynonyms(synonyms).join(',').toString()
  return  await db.synonyms.put({ id: synonymIdString })
```

### update:
- since we are using the primary key for this, we delete the old one
- we create the formatted ordered synonyms list and turn it into a string of comma separated values
- create the new synonym

```
  await db.synonyms.delete({ id: synonymId })
  const synonymIdString = formatSynonyms(synonyms).join(',').toString()
  return await db.synonyms.put({ id: synonymIdString })
```

### deletion:
- delete the synonym by id

```
  await db.synonyms.delete({ id: synonymId })
```

## the general concept
- you have moderators that want to create synonymous events lists and NOT have to individually add circulars to that.
- they create a comma separated list of eventIds
- that is added to the synonyms table as a csv string.
- when you want to find these groupings created by moderators, you query the synonyms table with normal paginated queries to get distinct synonym groups. 
- you take these distinct synonyms and use those to query the circulars index by eventId where the circulars.eventId is one of the values in the array of synonym id components. 
- i.e. we have a synonym "apple,cat,purple", that gets converted into ["apple", "cat", "purple"] and we search the circulars index where eventId is in ["apple", "cat", "purple"]
- this means that without having to mark any individual circulars, you are able to group them by their eventIds by creating independent synonyms

## accessing circulars by synonym groups
- around 95% of circulars will have an eventId after we backfill those values
- the grouping page queries the synonyms table to get the groups of event ids (splitting on ',' on the csv in synonyms.id to get an array of string event ids)
    - this is very important for pagination. aggregates use a key to paginate so all you have is previous or next. You do NOT have ordinal pagination with aggregates.
    - instead of doing an aggregation, since the synonyms table is it's own table with unique records, we can grab them using typical ordinal pagination which allows for a seamless integration with existing pagination. 
    - the table is queried by primary id which is the least impactful of queries
- When viewing grouped circulars or finding what circulars are related, the eventId on the circular is used in conjunction with the synonyms
    - circular has an eventID on it
    - synonym has list of event ids
    - circulars are queried like so with an array of string event ids passed in (it's shorter and clearer to write my intent in psql):
    ```
        SELECT * from circulars WHERE eventId in (['event 1', 'event 2', 'event 3']);
    ```
- To find any synonyms associated with a particular circular, you search like so:
    ```
        SELECT * from synonyms where id ilike '%eventId%'
    ```

if you want to get an index of grouped circulars:
1. you get a synonym id, which is a csv string of eventIds paginated ordinally for an index (with the standard limit and page method)
2. you turn that csv string into an array of strings (this makes it easier to manipulate and query by, but an array of strings can't be a primary id)
3. you pass that into the function to get circulars by eventId where eventId is in that array of string eventIds. 
4. This returns all related circulars where eventId is in the array of synonyms

if you want to find groups with a certain eventId or partial eventId:
1. you can search by eventId for specific groups that contain that eventId and only those that do (supports partial match) will be returned
```
query: {
  match: {
    id: `*${eventId}*`
  }
}
```
2. you turn that csv string into an array of strings (this makes it easier to manipulate and query by, but an array of strings can't be a primary id)
3. you pass that into the function to get circulars by eventId where eventId is in that array of string eventIds. 
4. This returns all related circulars where eventId is in the array of synonyms

## why?
- synonyms and synonym grouping are completely different domains/concepts/models than circulars. Trying to make synonyms fit into circulars when they are something that can be manipulated independently and have their own lifecycle did not keep the domains clean. Synonyms are inherently a separate entity. As such, they deserve their own table. All that I needed to store in that table though is a string id. This allows us to have a unique index of synonyms to manage independently. One thing we can change here is that we could use a different field and give an ordinal id for the synonyms. This leads to a bit of a redundant "synonyms.synonyms" naming pattern, whereas right now it's "synonyms.id"
- Pagination was one of the biggest challenge since using values exclusively on the circulars table would require an aggregation. Once synonyms were broken out on their own, normal ordinal pagination was a possibility.
- having a separate table allows us to NOT have to keep the circular up to date or in sync with anything
    - it's expensive to update a document in a document store, especially multiple at once when it impacts multiple records which is common in this scenario.
    - too many moving parts lead to bugs if data doesn't match in both places. If we minimize the places synonyms are repeated, it leads to better consistency.
    - we can infer the synonyms by querying the synonym table for anything with the eventId in it.

app/email-incoming/circulars/index.ts
```diff
@@ -84,12 +85,16 @@ export const handler = createEmailIncomingMessageHandler(
...
+    const eventId = parseEventFromSubject(parsed.subject)
+    const synonyms = eventId ? [eventId] : []
     const circular = {
       subject: parsed.subject,
       body: parsed.text,
       sub: userData.sub,
       submitter: formatAuthor(userData),
       submittedHow,
+      eventId,
+      synonyms,
```
lpsinger, line 97: Looks like you _are_ storing the synonyms in the Circulars table. Is this a mistake?

Courey: you said not to update code before explaining the methodology. so I explained and did not update the code.

lpsinger:
Thanks for writing that up. This is more clear to me now. What if we combine this approach with #745? Namely:

- Add eventId field to circulars table in DynamoDB
- Add synonyms table to DynamoDB with a the eventId as the partition key and a UUID field which is used to bind terms together as synonyms
- Upload synonyms to OpenSearch so that they are respected for full text searches
- Aggregate by group in JavaScript, but using the synonyms DynamoDB table instead of OpenSearch

Also, I see a limitation of this approach:

- In the grouped view, it is not guaranteed that all of the Circulars for a given event will appear on the same page. We can fix that by adding a DynamoDB index on the circulars.eventId column and doing a second (batch) DynamoDB query to get all of the Circulars for the events that are shown on the current page.

app/email-incoming/circulars/index.ts
lpsinger: Oh, I see. But you're planning on removing that?

Courey:
> Ping...

pong. at a conference this week, so trying to pay attention to that. I'll get back to this monday.

app/components/Header.tsx
```diff
@@ -125,6 +125,14 @@ export function Header() {
               >
                 Circulars
               </NavLink>,
+              <NavLink
```
Courey, line 128: this needs to be moderator only. will put this in

PR #1519, opened 2023-10-11
Enable fuzzy full text searches for GCN Circulars

lpsinger:
# Before

Screenshot 2023-10-11 at 15 40 04

# After

Screenshot 2023-10-11 at 15 39 48

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -127,6 +127,7 @@ export async function search({
                 multi_match: {
                   query,
                   fields: ['submitter', 'subject', 'body'],
+                  fuzziness: 'AUTO',
                 },
               }
             : undefined,


Courey (approved):

lpsinger:
> is there an actual circular with that event id in the subject or body?

No, there isn't. If the search term looks like a number _and it matches the ID of an existing GCN Circular_, then this function simply redirects the user to that GCN Circular: https://github.com/nasa-gcn/gcn.nasa.gov/blob/2ad4607f922529d7fa2a2867c99e1d0e11467aff/app/routes/circulars/circulars.server.ts#L251-L264

> the results still don't look very accurate.

I see... it's taking the string `230802` and computing the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between words in the text, and finding small changes (single transpositions or substitutions of characters) which result in a match.

I think that what we are finding here is that while we may want fuzzy matches for words (i.e., to catch misspellings, spelling variations, different verb conjugations, etc.) we do _not_ want fuzzy matches against digits.

PR #1520, opened 2023-10-11
Remove rel="external" from AstroData links

lpsinger:
GCN Circulars frequently rely on monospaced ASCII tables, and the external link icon disrupts layout.

# Example

Screenshot 2023-10-11 at 15 47 18


dakota002 (approved):
Looks like a simple change, but the checks are mad at something

lpsinger:
I think that GitHub was just glitching.

PR #1543, opened 2023-10-18
Adds a delete option to the edit page for email notifications

### LINKED ISSUE #1541: Can't turn off Notices
While experimenting with the new GCN, I tried turning on Fermi notices.  When trying to turn these off, I can't save the setting ('Save' button is greyed out). It's a specific corner case where I want to deactivate *all* notices.

For now, I turned off the full set of Fermi notices and am just subscribed to GBM alert notices.

I heard about the latest GCN at the Tucson MMA conference---thanks for setting this up, it looks like a great system!

dakota002:
Adds a delete button to the email notification edit page. It is only visible if the uuid exists
Resolves #1541 

image

### DIFF
--- a/app/routes/user.email.edit.tsx
+++ b/app/routes/user.email.edit.tsx
@@ -179,16 +179,6 @@ export default function () {
           >
             Cancel
           </Link>
-          <Button
-            name="intent"
-            value={notification.uuid ? 'update' : 'create'}
-            disabled={
-              !(nameValid && recipientValid && alertsValid && recaptchaValid)
-            }
-            type="submit"
-          >
-            {notification.uuid ? 'Update' : 'Save'}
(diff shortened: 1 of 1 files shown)


lpsinger (changes requested):
Suggestion: make this a single Form with multiple submit buttons. Each button has `name="intent"` and a `value="delete"`, `value="create"`, or `value="update"`, as appropriate.

The form data will contain only the value corresponding to the submit button that was clicked.

lpsinger:
Also, could we have the "Delete" button between the Cancel and Save buttons?

lpsinger (changes requested):
Now you can delete the hidden `intent` input.

lpsinger (approved):

Issue #1584, opened 2023-10-25
Select Python API framework


lpsinger:
Select a Python framework for URL routing. Criteria:

- Integration with OpenAPI and Swagger for automated generation of schema and language bindings
- Integration with AWS Lambda
- Project health (contributors, documentation, supply chain risk management)

lpsinger:
@jak574 and I seem to be converging on [FastAPI](https://fastapi.tiangolo.com) + [Mangum](https://mangum.io).

PR #1599, opened 2023-10-27
Add Python Lambda

### LINKED ISSUE #1583: Create placeholder Python Lambda

lpsinger:
Fixes #1583.

Depends on architect/hydrate#181 and architect/sandbox#727.


requirements.txt
```diff
@@ -0,0 +1,4 @@
+astropy
```
Courey, line 1: where is this being used?

lpsinger: It isn't yet, but it will be.

lpsinger: Would you prefer that I remove this?

dakota002 (approved):
Seems straight-forward enough, though I am not familiar with FastAPI and Magnum. I did some reading on the former and I get that part now. Is [this](https://pypi.org/project/magnum/) the latter?

lpsinger:
> Seems straight-forward enough, though I am not familiar with FastAPI and Magnum. I did some reading on the former and I get that part now.

They were new to me too, but @jak574 has used them (see #1584). The advantage of FastAPI over more common Python web frameworks like Flask or Django is that FastAPI provides an auto-generated OpenAPI spec and API browser (derived from Python type annotations) whereas Flask or Django require third-party extensions and a bunch of boilerplate to do that.

> Is [this](https://pypi.org/project/magnum/) the latter?

Yes. It is an adapter that allows you to run any Python ASGI app in a serverless context (e.g. AWS Lambda).

lpsinger:
> Is [this](https://pypi.org/project/magnum/) the latter?

Ah, no, actually, it's [mangum](https://pypi.org/project/mangum/) (note the deliberately odd spelling).

dakota002:
Ah now that makes more sense. I was reading it as magnum this whole time

jak574:
> Seems straight-forward enough, though I am not familiar with FastAPI and Magnum. I did some reading on the former and I get that part now. Is [this](https://pypi.org/project/magnum/) the latter?

It's "Mangum" not Magnum. https://mangum.io

Basically a ASGI layer to allow FastAPI to work with AWS Lambda.

PR #1606, opened 2023-10-31
added note about attachments

jracusin:
addresses concern raised in support ticket 48


lpsinger (changes requested):
Please use GitHub Flavored Markdown footnote syntax: https://docs.github.com/en/enterprise-cloud@latest/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes

jracusin:
I don't like how it puts the footnote all the way at the bottom of the page.  I want users to see it right below the table.  I tried using [!NOTE], but it doesn't render properly, it just prints out "[!NOTE]".

lpsinger:
> I don't like how it puts the footnote all the way at the bottom of the page. I want users to see it right below the table. I tried using [!NOTE], but it doesn't render properly, it just prints out "[!NOTE]".

OK. This is fine, then.

lpsinger (approved):

PR #1617, opened 2023-11-02
Set Python tab size to 4

lpsinger:
A tab size of 2 is pretty normal for JavaScript, but Python codes typicaly use 4. We are using Black to format Python code, and Black uses 4-space tabs.


dakota002 (approved):

jak574 (approved):
Sounds good. FYI I use black for formatting.

Courey (approved):

PR #1671, opened 2023-11-17
Replace links that were previously to wikipedia resources

### LINKED ISSUE #1590: Remove links to Wikipedia
> It's a NASA policy when it comes to public communications to only link to trusted other agency or institution websites. Anyone can change the content of wikipedia.

Is there a link to that policy? I haven't heard of it before.

_Originally posted by @lpsinger in https://github.com/nasa-gcn/gcn.nasa.gov/pull/1540#discussion_r1364558007_

dakota002:
Resolves #1590

### DIFF
--- a/app/routes/_gcn.docs.contributing.deployment/route.mdx
+++ b/app/routes/_gcn.docs.contributing.deployment/route.mdx
@@ -11,13 +11,13 @@ import deploymentScreenshot from './screenshot.png'
 
 This web site is deployed **in the cloud** on [Amazon Web Services (AWS)](https://aws.amazon.com) in order resist outages due to local network conditions or distasters.
 
-Site assets are replicated and cached through a global **[content delivery network](https://en.wikipedia.org/wiki/Content_delivery_network)** ([AWS CloudFront](https://aws.amazon.com/cloudfront/)) to provide the fastest experience possible to scientists everywhere regardless of their country and location.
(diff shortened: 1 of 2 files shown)


lpsinger (commented):
This deletes some Wikipedia references. For others, it replaces a a Wikipedia reference with an article from one vendor --- for example, it replaces the Wikipedia article on CI/CD with GitHub's own page. I am not sure the these are better sources --- they are certainly not balanced sources.

dakota002:
Fair, aside from the policy that warranted this issue in the place, I do actually think the Wikipedia links are the best source for information on the topics here. But if not them, my thought was to either replace them with links to the specifics of what we are using (like the GitHub ci-cd page) or just remove them if they seemed redundant (like the doubly defined snake_case). The latter is pretty simple enough to define right in the text as it is. 

Should we maybe add some kind of site glossary? Something like a reference for people who are curious about how it works, but don't know the jargon? Then we could just replace all the external links with links to that page

lpsinger:
I don't think that there _is_ such a policy --- only the stated opinion of one person in the Astrophysics Science Division communications office.

A glossary is not a bad idea, but the glossary would need links to sources too.

lpsinger:
Unfortunately, I think that in these remaining cases, Wikipedia _is_ the best resource. Is it OK if we close this?

CC @jracusin.

Issue #1670, opened 2023-11-17
Non-working links on page detailing Swift GUANO Kafka alerts


jak574:
The following page:

https://gcn.nasa.gov/missions/swift#json-serialized-gcn-notice-types-in-gcn-kafka

contains several links that do not function:
* [GCN Schema](https://gcn.nasa.gov/schema/stable/gcn/notices/swift/bat/guano.schema.json) leads to a "404: Not Found".
* [JSON Message](https://raw.githubusercontent.com/nasa-gcn/gcn-schema/v3.0.0/gcn/notices/swift/bat/guano.example.json) leads to a "404: Not Found"
* [Schema Browser](https://gcn.nasa.gov/docs/schema/v3.0.0/gcn/notices/swift/bat/guano.example.json) leads to a GCN page with the error "Unexpected error. An unexpected error occurred."
* [Detailed Description and Examples](https://github.com/nasa-gcn/gcn-schema/tree/main/gcn/notices/swift/bat/guano) leads to a "404 - page not found"

lpsinger:
Good catch! It looks like in many cases the problem is a matter of capitalization: https://github.com/nasa-gcn/gcn-schema/tree/main/gcn/notices/swift/bat

@jak574, would you be interested in attempting a PR for this, to get some practice with editing Markdown content in Remix? If not, then I can write a PR later this afternoon.

jak574:
@lpsinger I'll give it a go.

jak574:
I submitted https://github.com/nasa-gcn/gcn.nasa.gov/pull/1673

lpsinger:
Fixed by #1673.

PR #1673, opened 2023-11-17
Fix for broken Swift/BAT GUANO schema links.

jak574:
Links to the GUANO schema were incorrect. This pull attempts to fix that, fixing three issues:

* Links to the schema on GitHub were corrected so guano -> Guano (note uppercase G)
* Link to Schema Browser was to the incorrect filename "guano.example.json", which is fixed to "Guano.schema.json".
* [Detailed Description and Examples](https://github.com/nasa-gcn/gcn-schema/tree/main/gcn/notices/swift/bat) linked to a non-existent "guano". This now links to the "bat".

~~Quick note: Schema Browser does not work in my dev environment, so I cannot verify the fix.~~ Schema Browser link verified to work.

### DIFF
--- a/app/routes/_gcn.missions.swift/route.mdx
+++ b/app/routes/_gcn.missions.swift/route.mdx
@@ -285,11 +285,11 @@ These are bursts that did not result in any of the automatic GCN Notices describ
 
 These notices are published on the GCN Kafka topic `gcn.notices.swift.bat.guano`.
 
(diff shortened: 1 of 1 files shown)


lpsinger (approved):

lpsinger (approved):
Looks great! Note that in the future if you add `Fixes #1670` to your commit message, then it will automatically resolve the issue on merge.

PR #1680, opened 2023-11-21
Circulars history UI

dakota002:


dakota002:
This update should all be behind the `CIRCULAR_VERSIONS` feature flag. There is still some more to come, but i am updating quite a few file names, so I think this part is best merged in before I continue

lpsinger:
Thank for rebasing to take care of the merge conflicts from #1705. I swear I will complete a review today!

Courey:
the code looks fine. the one thing I was wondering about was the change of the route from new to edit for the create/update circulars. I think it's the correct thing to do, but do you think we will interrupt the flow of anyone that may have that bookmarked to just fire off a circular? if so, could we put a redirect from new -> edit?
If you don't think that it's likely to break anyone's workflow, it otherwise looks good to me!

dakota002:
I don't imagine it will be, but I can add one either way to be safe and we can phase it out over time to be sure

Courey (approved):

lpsinger:
I'll get to reviewing this shortly.

lpsinger (changes requested):
- [x] Let's not rename the /circulars/new route to /circulars/edit.
- [x] Can you add a layout route for the version number? You'd have routes `_gcn.circulars.$circularId` and `_gcn.circulars.$circularId.($version)`. That way the version number list doesn't have to be queried for every navigation between versions.

dakota002:
I should be able to make those work, i'll get a new commit out shortly

dakota002:
@lpsinger I made some updates for the layout route, but I think there is some duplicated stuff that could still be revised. Any suggestions?

Also, not sure if it is related, but recently I noticed that when running locally, there are a LOT of canceled requests any time there is a code change, maybe something is up with the hot reloading? 
image

This is just one chunk, but I saved 1 file and saw this pattern repeat at least 4 times over in the network tab

app/routes/_gcn.circulars/route.tsx
```diff
@@ -8,12 +8,15 @@
...
 
+import { action as parentAction } from '../_gcn.circulars._archive._index/route'
 import type { BreadcrumbHandle } from '~/root/Title'
 
 export const handle: BreadcrumbHandle = {
   breadcrumb: 'Circulars',
 }
 
+export const action = parentAction
+
```
lpsinger, line 19: This is weird. Why do we have two routes with the same action?

app/routes/_gcn.circulars.$circularId/route.tsx
```diff
@@ -7,43 +7,76 @@
...
-import { origin } from '~/lib/env.server'
-import { getCanonicalUrlHeaders, pickHeaders } from '~/lib/headers.server'
+import { getUser } from '../_gcn._auth/user.server'
+import {
+  get,
+  getVersions,
+  moderatorGroup,
+} from '../_gcn.circulars/circulars.server'
+// import { PlainTextBody } from './Body'
+// import { FrontMatter } from './FrontMatter'
```
lpsinger, line 20: Remove commented out code.

app/routes/_gcn.circulars.$circularId.($version)[.]json.ts
```diff
@@ -10,12 +10,18 @@ import invariant from 'tiny-invariant'
...
+export async function loader({
+  params: { circularId, version },
+}: DataFunctionArgs) {
   invariant(circularId)
-  const result = await get(parseFloat(circularId))
+
+  const result = await get(
+    parseFloat(circularId),
+    version && feature('CIRCULAR_VERSIONS') ? parseFloat(version) : undefined
+  )
```
Comment (lpsinger) [inline app/routes/_gcn.circulars.$circularId.($version)[.]json.ts:24]: If the `CIRCULAR_VERSIONS` feature is disabled and the `version` route param is defined, then we should throw a 404 error.

lpsinger (changes requested):

app/routes/_gcn.circulars/route.tsx
dakota002: This was a weird one, but without it I was getting an error about an error about the action not being defined. I'll see if I can replicate it

dakota002: Couldn't replicate, removed in latest

app/routes/_gcn.circulars.$circularId.($version)[.]json.ts
Comment (dakota002) [reply app/routes/_gcn.circulars.$circularId.($version)[.]json.ts:25]: Resolved in latest

app/routes/_gcn.circulars.$circularId.($version)[.]json.ts
```diff
@@ -10,12 +10,19 @@ import invariant from 'tiny-invariant'
...
+export async function loader({
+  params: { circularId, version },
+}: DataFunctionArgs) {
   invariant(circularId)
-  const result = await get(parseFloat(circularId))
+  if (!feature('CIRCULAR_VERSIONS') && version)
+    throw new Response(null, { status: 404 })
+  const result = await get(
+    parseFloat(circularId),
+    version && feature('CIRCULAR_VERSIONS') ? parseFloat(version) : undefined
```
Comment (lpsinger) [inline app/routes/_gcn.circulars.$circularId.($version)[.]json.ts:24]:
```suggestion
    version ? parseFloat(version) : undefined
```

app/routes/_gcn.circulars.$circularId.($version)[.]txt.ts
```diff
@@ -10,12 +10,19 @@ import invariant from 'tiny-invariant'
...
+export async function loader({
+  params: { circularId, version },
+}: DataFunctionArgs) {
   invariant(circularId)
-  const result = await get(parseFloat(circularId))
+  if (!feature('CIRCULAR_VERSIONS') && version)
+    throw new Response(null, { status: 404 })
+  const result = await get(
+    parseFloat(circularId),
+    version && feature('CIRCULAR_VERSIONS') ? parseFloat(version) : undefined
```
Comment (lpsinger) [inline app/routes/_gcn.circulars.$circularId.($version)[.]txt.ts:24]:
```suggestion
    version ? parseFloat(version) : undefined
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,72 @@
...
+      } = data
+      return `${circularId}: ${subject}`
+    }
+  },
+}
+
+export async function loader({
+  params: { circularId, version },
+}: DataFunctionArgs) {
+  if (!circularId)
```
lpsinger, line 37:
```suggestion
  if (!feature('CIRCULAR_VERSIONS') && version)
    throw new Response(null, { status: 404 })
  if (!circularId)
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,72 @@
...
+    throw new Response('circularId must be defined', { status: 400 })
+  const result = await get(
+    parseFloat(circularId),
+    version ? parseFloat(version) : undefined
+  )
+  return json(
+    { result, version },
```
lpsinger, line 44:
```suggestion
    result,
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,72 @@
...
+  publicStaticShortTermCacheControlHeaders,
+} from '~/lib/headers.server'
+import type { BreadcrumbHandle } from '~/root/Title'
+
+export const handle: BreadcrumbHandle<typeof loader> = {
+  breadcrumb({ data }) {
+    if (data) {
+      const {
+        result: { circularId, subject },
+      } = data
```
lpsinger, line 28:
```suggestion
      const { circularId, subject } = data
```

app/routes/_gcn.circulars.$circularId/route.tsx
```diff
@@ -7,43 +7,69 @@
...
-  const result = await get(parseFloat(circularId))
-  return json(result, {
-    headers: getCanonicalUrlHeaders(
-      new URL(`/circulars/${circularId}`, origin)
-    ),
-  })
+  const result = await get(
+    parseFloat(circularId),
+    version ? parseFloat(version) : undefined
+  )
```
lpsinger, line 45: Why does this route need to get the circular? The route `_gcn.circulars.$circularId.($version)` does that.

lpsinger (changes requested):

app/routes/_gcn.circulars.$circularId/route.tsx
dakota002: I think it's just for the bibcode, but I can probably refactor this

dakota002: @lpsinger actually this has me wondering, will each version have a unique bibcode, or will that persist through the versions?

lpsinger: I don't know. Why does that matter for the bibcode? The ADS link should be in the child route.

dakota002: Gotcha, I updated it to be in the child route

app/routes/_gcn.circulars.$circularId/route.tsx
```diff
@@ -7,42 +7,45 @@
...
-    if (data) {
-      const { circularId, subject } = data
-      return `${circularId}: ${subject}`
-    }
-  },
-}
-
-export async function loader({ params: { circularId } }: DataFunctionArgs) {
+export async function loader({
+  params: { circularId, version },
```
lpsinger, line 21: `version` is present in a child route, but not in _this_ route. According to https://reactrouter.com/en/main/hooks/use-params, "child routes inherit all params from their parent routes." However, the documentation does _not_ say that parent routes have access to their child routes' params. This would rely on undocumented behavior, which might go away as a future "bug fix."

Also, the loader does not need to return URL params that it doesn't actually use for any computations, as these are available to the route component from the `useParams()` hook.

lpsinger (changes requested):

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+    if (data) {
+      const { circularId, subject } = data
+      return `${circularId}: ${subject}`
+    }
+  },
+}
+
+export async function loader({
+  params: { circularId, version },
+}: DataFunctionArgs) {
```
lpsinger, line 34:
```suggestion
}: LoaderFunctionArgs) {
```

See #1754.

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (!feature('CIRCULAR_VERSIONS') && version)
+    throw new Response(null, { status: 404 })
+  if (!circularId)
+    throw new Response('circularId must be defined', { status: 400 })
```
lpsinger, line 38:
```suggestion
  invariant(circularId)
```

It's a precondition that circularId must be a non-empty string because Remix won't route us to this module otherwise.

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,68 @@
...
+  pickHeaders,
+  publicStaticShortTermCacheControlHeaders,
+} from '~/lib/headers.server'
+import type { BreadcrumbHandle } from '~/root/Title'
+
+export const handle: BreadcrumbHandle<typeof loader> = {
+  breadcrumb({ data }) {
+    if (data) {
+      const { circularId, subject } = data
+      return `${circularId}: ${subject}`
```
lpsinger, line 27: The version number, if present, should affect the title. Do you have access to `params` in the breadcrumb callback?

app/routes/_gcn.circulars.$circularId/route.tsx
```diff
@@ -10,44 +10,56 @@ import {
...
     useLoaderData<typeof loader>()
   const searchString = useSearchString()
+
+  const routeMatch = useMatches().find(
+    (x) => x.id == 'routes/_gcn.circulars.$circularId.($version)'
+  )
+
+  const versionRouteMatchData = routeMatch?.data as Circular
+  const bibcode = versionRouteMatchData.bibcode
+  const version = versionRouteMatchData.version
```
lpsinger, line 62: I'm apprehensive of components using route params from child routes. The other way around, a route using params from parents, I think is fine.

lpsinger (changes requested):

dakota002:
Woops, requested a review before seeing some of your latest feedback. I'm reviewing that now

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
dakota002: Will add in next commit

app/routes/_gcn.circulars.$circularId.$version[.]json.ts
```diff
@@ -0,0 +1,30 @@
...
+import { feature, origin } from '~/lib/env.server'
+import { getCanonicalUrlHeaders } from '~/lib/headers.server'
+
+export async function loader({
+  params: { circularId, version },
+}: LoaderFunctionArgs) {
+  invariant(circularId)
+  if (!feature('CIRCULAR_VERSIONS') && version)
+    throw new Response(null, { status: 404 })
+  invariant(version)
```
Comment (lpsinger) [inline app/routes/_gcn.circulars.$circularId.$version[.]json.ts:22]: Let's rename this route to `_gcn.circulars.$circularId.($version)[.]json.ts` so that the version route parameter is optional.

lpsinger (changes requested):

app/routes/_gcn.circulars.$circularId.$version[.]json.ts
Comment (dakota002) [reply app/routes/_gcn.circulars.$circularId.$version[.]json.ts:22]: I had that previously, but this doesn't behave as expected. 

The path /circulars/####.json renders as:
image

Comment (dakota002) [reply app/routes/_gcn.circulars.$circularId.$version[.]json.ts:22]: This only happens if there is not  _gcn.circulars.$circularId[.]json.ts too

app/routes/_gcn.circulars.$circularId/route.tsx
dakota002: Not sure I follow how that would work in this case. The getter for the specific circular only exists in the child route, so the bibcode and version aren't available at the parent level

lpsinger: Right, so this UI must be in the _child_ route.

dakota002: These correspond to the buttons on top (text/json and citation links). These were the ones we talked about previously that I had moved below, Should I move them into the child component then?
image

lpsinger: Whatever order the button bar and the front matter appear in on the page, I think that they are both going to have to be in the child route because both require knowledge of the current version.

I would like you to have the parent route's loader return the list of versions, and employ `useRouteLoaderData` in the child route to retrieve the list. The parent route might end up having no UI content itself.

dakota002: Ahhh i see what you're saying. Okay i can make that work

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -301,6 +301,36 @@ export async function circularRedirect(query: string) {
...
+    ...circular,
+    editedBy: formatAuthor(user),
+    createdOn: Date.now(),
+    bibcode,
+    submitter,
+    submittedHow,
+  }
+  // Remove fields if they are undefined, as this will otherwise throw an error in the PUT
+  if (!bibcode) delete newCircularVersion.bibcode
+  if (!submittedHow) delete newCircularVersion.submittedHow
```
lpsinger, line 330: If we use spread syntax here, then we don't need to manually delete undefined properties:

```suggestion
  // Need to be retrieved otherwise will be absent in latest version
  const oldCircular = await get(circular.circularId)

  const newCircularVersion = {
    ...oldCircular,
    ...circular,
    editedBy: formatAuthor(user),
    createdOn: Date.now(),
  }
  if (!submittedHow) delete newCircularVersion.submittedHow
```

But... uh oh, we probably _don't_ want to modify the createdOn field. Perhaps we need a new revisedOn field. (That should be a separate PR since it affects backend code that is already on the main branch.)

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { useFeature } from '~/root'
+import type { BreadcrumbHandle } from '~/root/Title'
+
+export const handle: BreadcrumbHandle<typeof loader> = {
+  breadcrumb({ data }) {
+    if (data) {
+      const { circularId, subject, version } = data
+      return `${circularId}${version ? ` - v${version}` : ''}: ${subject}`
+    }
+  },
```
lpsinger, line 34:
```suggestion
  breadcrumb({ data }) {
    if (data) {
      const { circularId, subject, version } = data
      return `${circularId}${version ? ` - v${version}` : ''}: ${subject}`
    }
  },
```

- [ ] I suggest presenting the title as `NNNNNN - Version M - Subject`, since I do not want to encourage the notation `NNNNNNvM` for citing Circulars.
- [ ] Given that, the breadcrumb for the circular ID can be returned in the parent route.

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,165 @@
...
+    parseFloat(circularId),
+    version ? parseFloat(version) : undefined
+  )
+
+  return json(result, {
+    headers: {
+      ...publicStaticShortTermCacheControlHeaders,
+      ...getCanonicalUrlHeaders(
+        new URL(
+          `/circulars/${circularId}${version ? `/${version}` : ''}`,
```
lpsinger, line 53:
```suggestion
          `/circulars/${circularId}`,
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,165 @@
...
+            className="usa-button usa-button--outline"
+          >
+            Edit
+          </Link>
+        )}
+      </ButtonGroup>
+      <h1 className="margin-bottom-0">GCN Circular {circularId} </h1>
+      {useFeature('CIRCULAR_VERSIONS') && (
+        <CircularsHistory circular={circularId} history={result?.history} />
+      )}
```
lpsinger, line 134: Perhaps we should only show the version info if there is more than one version (history is not empty)?

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,165 @@
...
+        {history &&
+          history.map((version) => (
+            <span key={version.toString()}>
+              <Link to={`/circulars/${circular}/${version}`}>
+                Version {version}
+              </Link>
+              ,{' '}
+            </span>
+          ))}
+      </div>
```
lpsinger, line 162: How about a dropdown menu in the button bar?

app/routes/_gcn.circulars._archive._index/route.tsx
```diff
@@ -68,10 +70,27 @@ export async function action({ request }: ActionFunctionArgs) {
...
+        body,
+        circularId: parseFloat(circularId),
+        subject,
+      },
+      user
+    )
+    result = await get(parseFloat(circularId))
+    if (intent === 'distribute') {
+      // TODO: Send Email
+    }
```
lpsinger, line 89: What's this for?

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -0,0 +1,353 @@
...
+          <SyntaxExample label="Web address">
+            https://www.swift.psu.edu
+          </SyntaxExample>
+        </tbody>
+      </Table>
+    </>
+  )
+}
+
+export default function CircularEditForm({
```
lpsinger, line 103: I'm trying to get out of the habit of using default exports, except in route modules.

```suggestion
export function CircularEditForm({
```

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -0,0 +1,353 @@
+/*!
```
lpsinger, line 1: Breaking out CircularEditForm into its own module is a good refactoring, period. Given that this PR has turned out to be quite complicated, this refactoring would be a good save point to get onto the main branch ASAP. That way, if any other changes occur to this form before the present PR is merged, there is less for you to rework. Would you please open a separate PR to do that refactoring?

lpsinger (changes requested):

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
dakota002: I think I caught all the references in this one:

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
dakota002: Gotcha, I'll include this in #1784

app/routes/_gcn.circulars._archive._index/route.tsx
dakota002: Woops that was left over from a previous path I was going down. I'll remove it in the next commit

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
dakota002: image

image

app/routes/_gcn.circulars/circulars.server.ts
dakota002: Added https://github.com/nasa-gcn/gcn.nasa.gov/pull/1802 and I will implement the spread syntax next

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
lpsinger: Let's separate this from the Text/JSON/Cite segmented button. Those are grouped together into a segmented button because they are different formats; the Version dropdown doesn't go with them.

dakota002: image

lpsinger: I like it!

dakota002: Sweet, its in the latest commit!

app/routes/_gcn.circulars.$circularId.($version)/FrontMatter.tsx
```diff
@@ -40,17 +40,31 @@ export function FrontMatter({
...
+  const displayDate = editedOn ?? createdOn
   return (
     <>
       <FrontMatterItem label="Subject">{subject}</FrontMatterItem>
       <FrontMatterItem label="Date">
-        {formatDateISO(createdOn)}{' '}
+        {formatDateISO(displayDate)}{' '}
         <small className="text-base-light">
-          (<TimeAgo time={createdOn}></TimeAgo>)
+          (<TimeAgo time={displayDate}></TimeAgo>)
```
lpsinger, line 61: We should display both the date created and the date revised.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -301,6 +301,32 @@ export async function circularRedirect(query: string) {
...
+  // Need to be retrieved otherwise will be absent in latest version
+  const oldCircular = await get(circular.circularId)
+
+  const newCircularVersion = {
+    ...oldCircular,
+    ...circular,
+    editedBy: formatAuthor(user),
+    editedOn: Date.now(),
+  }
+  // if (!submittedHow) delete newCircularVersion.submittedHow
```
lpsinger, line 326: Remove commented out code.

sandbox-seed.json
```diff
@@ -3334,11 +3334,12 @@
     {
       "subject": "AMON Coincidence Alert from the sub-threshold IceCube-HAWC search NuEm-230927A",
       "submittedHow": "email-legacy",
-      "createdOn": 1695837395712,
+      "editedOn": 1695837395712,
```
lpsinger, line 3337: Both the `editedOn` and `createdOn` fields should be present in these examples.

app/components/DetailsDropdownButton.tsx
```diff
@@ -18,7 +18,10 @@ export default function DetailsDropdownButton({
   return (
     <Button type="button" {...props}>
       {children}
-      <Icon.ExpandMore role="presentation" />
+      <Icon.ExpandMore
+        role="presentation"
+        className="margin-top-neg-1 bottom-aligned"
+      />
```
lpsinger, line 24: What does this change do?

app/routes/_gcn.circulars.$circularId/route.tsx
```diff
@@ -10,93 +10,45 @@ import {
...
+  const user = await getUser(request)
+  const userIsModerator = user?.groups.includes(moderatorGroup)
+  return json(
+    { circularId, history, userIsModerator },
+    {
+      headers: getCanonicalUrlHeaders(
+        new URL(`/circulars/${circularId}`, origin)
+      ),
+    }
+  )
```
lpsinger, line 50: This route doesn't need to set a canonical URL. It's impossible to actually navigate to this route since it is a layout route.

```suggestion
  return { circularId, history, userIsModerator }
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,185 @@
...
+            >
+              Cite (ADS)
+            </Button>
+          )}
+        </ButtonGroup>
+        {useFeature('CIRCULAR_VERSIONS') &&
+          result?.history &&
+          result?.history.length > 0 && (
+            <CircularsHistory circular={circularId} history={result?.history} />
+          )}
```
lpsinger, line 124: Does TypeScript like this shorter version?

```suggestion
        {useFeature('CIRCULAR_VERSIONS') &&
          result?.history.length && (
            <CircularsHistory circular={circularId} history={result.history} />
          )}
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+export const headers: HeadersFunction = ({ loaderHeaders }) =>
+  pickHeaders(loaderHeaders, ['Link'])
+
+export default function () {
+  const { circularId, body, bibcode, version, ...frontMatter } =
+    useLoaderData<typeof loader>()
+  const searchString = useSearchString()
+
+  const result = useRouteLoaderData<typeof parentLoader>(
```
lpsinger, line 67:
```suggestion
  const { history, userIsModerator } = useRouteLoaderData<typeof parentLoader>(
```

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -0,0 +1,185 @@
...
+        {useFeature('CIRCULAR_VERSIONS') && result?.userIsModerator && (
+          <Link
+            to={`/circulars/edit/${circularId}`}
+            className="usa-button usa-button--outline"
+          >
+            Edit
+          </Link>
+        )}
+      </ButtonGroup>
+      <h1 className="margin-bottom-0">GCN Circular {circularId} </h1>
```
lpsinger, line 134:
```suggestion
      <h1 className="margin-bottom-0">GCN Circular {circularId}</h1>
```

lpsinger (changes requested):

app/components/DetailsDropdownButton.tsx
dakota002: This one fixes the sizing so that the height of the button with an Icon is the same as one without

sandbox-seed.json
dakota002: It is, this edit just corrects the key:
```json
    {
      "subject": "AMON Coincidence Alert from the sub-threshold IceCube-HAWC search NuEm-230927A",
      "submittedHow": "email-legacy",
      "editedOn": 1695837395712,
      "circularId": 34776,
      "submitter": "Hugo Ayala at Pennsylvania State University <hgayala@psu.edu>",
      "body": "...",
      "editedBy": "Example at Example <example@example.com>",
      "createdOn": 1694188746530,
      "version": 3
    },
```
The createdOn property should be consistent on all versions

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
dakota002: Unfortunately no, 
image

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
lpsinger: Yes, but once you remove the canonical URL thing in the parent route, it should work.

app/routes/_gcn.circulars.$circularId.($version)/FrontMatter.tsx
dakota002: @lpsinger do you think it looks better with the dates next to each other:
image

or the edit info next to each other:
image

lpsinger: Not sure yet: can you show both the original author and the author of the edit, in both styles?

dakota002: The original author is there in both as `From`, what format are you looking for?

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
dakota002: I see, I had removed the headers, but the json function was still there. Removing now

dakota002: I actually still get this error without the json part:
image

app/routes/_gcn.circulars.$circularId.($version)/FrontMatter.tsx
lpsinger: Oh, I see now. I like the first version.

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
lpsinger: OK, I'll take a look.

dakota002: From our other uses of the function it looks like we use it to define a const, then access it's properties

lpsinger: Oh, I see... it's because `useRouteLoaderData` is allowed to return `undefined`.


```ts
/**
 * Returns the JSON parsed data from the current route's `loader`.
 *
 * @see https://remix.run/hooks/use-loader-data
 */
export declare function useLoaderData<T = AppData>(): SerializeFrom<T>;
/**
 * Returns the loaderData for the given routeId.
 *
 * @see https://remix.run/hooks/use-route-loader-data
 */
export declare function useRouteLoaderData<T = AppData>(routeId: string): SerializeFrom<T> | undefined;
```

app/routes/_gcn.circulars.edit.$circularId/route.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const { isAuthorized, formattedEditor, circular } =
+    useLoaderData<typeof loader>()
+
+  const formDefaults = {
+    formattedContributor: formattedEditor,
+    circular,
+    defaultBody: circular?.body || '',
+    defaultSubject: circular?.subject || '',
+    searchString: '',
+    isAuthorized,
```
lpsinger, line 58: This is unused in the form:

```suggestion
```

However, we should definitely throw a 403 Forbidden error if the user is unauthorized.

lpsinger, line 58: This is unused in the form:

```suggestion
```

However, we should definitely throw a 403 Forbidden error if the user is unauthorized.

app/routes/_gcn.circulars.edit.$circularId/route.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const circular = await get(parseFloat(circularId))
+  return json(
+    { isAuthenticated, isAuthorized, formattedEditor, circular },
+    { headers: getCanonicalUrlHeaders(new URL(`/circulars/edit`, origin)) }
+  )
+}
+
+export const headers: HeadersFunction = ({ loaderHeaders }) =>
+  pickHeaders(loaderHeaders, ['Link'])
```
lpsinger, line 46: No need to provide a canonical URL, because this page should not be reachable by search engines.

```suggestion
  return { isAuthenticated, isAuthorized, formattedEditor, circular }
}
```

Please update the `handle` to exclude this page from the sitemap.

app/routes/_gcn.circulars.edit.$circularId/route.tsx
```diff
@@ -0,0 +1,66 @@
...
+    defaultBody: circular?.body || '',
+    defaultSubject: circular?.subject || '',
+    searchString: '',
+    isAuthorized,
+  }
+
+  return (
+    <AstroDataContext.Provider value={{ rel: 'noopener', target: '_blank' }}>
+      <CircularEditForm {...formDefaults} />
+    </AstroDataContext.Provider>
```
lpsinger, line 64: We moved the `AstroDataContext.Provider` inside the form.

```suggestion
    <CircularEditForm {...formDefaults} />
```

lpsinger (changes requested):

app/routes/_gcn.circulars.edit.$circularId/route.tsx
```diff
@@ -0,0 +1,51 @@
...
+  if (!circularId) throw new Response(null, { status: 404 })
+  const user = await getUser(request)
+  if (!user?.groups.includes(moderatorGroup))
+    throw new Response(null, { status: 403 })
+  const isAuthenticated = true
+  const isAuthorized = user.groups.includes(moderatorGroup)
+  const formattedEditor = formatAuthor(user)
+
+  const circular = await get(parseFloat(circularId))
+  return { isAuthenticated, isAuthorized, formattedEditor, circular }
```
lpsinger, line 36: Remove unused props:

```suggestion
  const formattedEditor = formatAuthor(user)

  const circular = await get(parseFloat(circularId))
  return { formattedEditor, circular }
```

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -98,11 +102,13 @@ function SyntaxReference() {
...
   formattedContributor,
+  circular,
   defaultBody,
   defaultSubject,
   searchString,
 }: {
   formattedContributor: string
+  circular?: Circular
   defaultBody: string
   defaultSubject: string
```
lpsinger, line 113: It seems odd that this takes  both the complete Circular _and_ the default subject and body.

lpsinger (changes requested):

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
dakota002: Yea good point, I can refactor this a bit better

app/routes/_gcn.circulars.edit.$circularId/route.tsx
dakota002: is that handle update just including this: `getSitemapEntries: () => null`

lpsinger: Yes.

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -124,21 +131,30 @@ export function CircularEditForm({
...
     <AstroDataContext.Provider value={{ rel: 'noopener', target: '_blank' }}>
-      <h1>New GCN Circular</h1>
+      <h1>{circularId ? 'Edit' : 'New'} GCN Circular</h1>
       <Form method="POST" action={`/circulars${formSearchString}`}>
+        {circularId && (
+          <>
+            <input type="hidden" name="circularId" value={circularId} />
+            <InputGroup className="border-0 maxw-full">
+              <InputPrefix className="wide-input-prefix">From</InputPrefix>
+              <span className="padding-1">{submitter} </span>
```
lpsinger, line 141:
```suggestion
              <span className="padding-1">{submitter}</span>
```

lpsinger (changes requested):

lpsinger (approved):

dakota002:
Good to go ahead and merge this now?

PR #1688, opened 2023-11-21
Implement Semantic Search

ronitagarwala01:
Implement Semantic Search in circulars.server.ts. Add ml.js module for making API calls to OpenSearch cluster

### DIFF
--- a/app/routes/_gcn.circulars/circulars.server.ts
+++ b/app/routes/_gcn.circulars/circulars.server.ts
@@ -47,46 +47,46 @@ export const getDynamoDBAutoIncrement = memoizee(
 )
 
 /** convert a date in format mm-dd-yyyy (or YYYY-MM_DD) to ms since 01/01/1970 */
-function parseDate(date?: string) {
-  return date ? new Date(date).getTime() : NaN
-}
+// function parseDate(date?: string) {
+//   return date ? new Date(date).getTime() : NaN
+// }
 
 /** take input string and return start/end times based on string value */
-function fuzzyTimeRange(fuzzyTime?: string) {
-  const now = Date.now()
-  switch (fuzzyTime) {
-    case 'now': // current time
-      return now
-    case 'hour': // 1 hour ago
-      return now - 3600000
-    case 'today': // 00:00:00 of same day
-      return new Date().setHours(0, 0, 0, 0)
-    case 'day': // 24 hours ago
-      return now - 86400000
-    case 'week': // 7 days ago
-      return now - 86400000 * 7
-    case 'month': // 30 days ago
-      return now - 86400000 * 30
-    case 'year': // 365 days ago
-      return now - 86400000 * 365
-    case 'mtd': // month to date
-      return new Date(
-        new Date().getFullYear(),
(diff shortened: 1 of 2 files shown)


tylerbarna (changes requested):
I think openSearchApi.js and openSearchAPI.js are duplicates

openSearchApi.js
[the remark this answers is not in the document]
ronitagarwala01: You're right. Just deleted openSearchAPI.js

lpsinger (commented):
As an experiment, would you please try writing the collection of API calls as an Architect `deploy.end` plugin in this project, rather than using the openSearchApi.js file? I'd like to see what that looks like.

ronitagarwala01:
I removed the api call file from here and added it to the plugin. Is that what you meant?

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -106,9 +106,65 @@ export async function search({
...
+    ? {
+        neural: {
+          circular_embedding: {
+            query_text: query,
+            model_id,
+            k: 100,
+          },
+        },
+      }
+    : { match_all: {} }
```
tylerbarna, line 144: This should restore the date filtering to the nlp query
```suggestion
const nlpSearchQuery = query
    ? {
        bool: {
          must: [
            {
              neural: {
                circular_embedding: {
                  query_text: query,
                  model_id,
                  k: 100,
                },
              },
            }
          ],
          filter: {
            range: {
              createdOn: {
                gte: startTime,
                lte: endTime,
              },
            },
          },
        },
      }
    : { match_all: {} }
```

tylerbarna (changes requested):

tylerbarna:
was there anything else we wanted to implement for this PR?

ronitagarwala01:
I don't believe so. I want to implement the changes for improving semantic search in another PR.

app/routes/_gcn.circulars/circulars.server.ts
```diff
@@ -94,41 +94,58 @@ export async function search({
...
+        resp.body['nlp-ingest-pipeline'].processors[0].text_embedding.model_id
+    } else {
+      console.log(
+        'Error. Could not deploy model. Returned with response: ',
+        resp
+      )
+    }
+  } catch (e) {
+    console.log('Error: ', e)
+  }
```
lpsinger, line 133: Would you please see if you can use the [raw HTTP request API](https://github.com/opensearch-project/opensearch-js/blob/main/guides/json.md) that was recently added to the OpenSearch client library?

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.server.ts
ronitagarwala01: Sure.

lpsinger:
Please resolve merge conflicts.

ronitagarwala01:
Alright. Will do.

ronitagarwala01:
Done.

Issue #1697, opened 2023-11-21
Create an About page under the labs route


dankocevski:
Create an empty page to contain information about the ACROSS effort under the labs route.

Issue #1696, opened 2023-11-21
Create a Conference page under the labs route


dankocevski:
Create an empty page to contain a summary of TDAMM related conferences under the labs route.

Issue #1695, opened 2023-11-21
Create a Proposals page under the labs route


dankocevski:
Create an empty page to contain a summary of TDAMM related proposal and solicitations under the labs route.

Issue #1694, opened 2023-11-21
Create a Tools page under the labs route


dankocevski:
Create an empty page to contain a summary of TDAMM related Tools under the labs route.

PR #1700, opened 2023-11-22
Adds header, footer, and empty labs pages

dakota002:
This adds empty pages for [Missions](#1693), [Tools](#1694),  [Proposals](#1695 ),  [Conferences](#1696),  and [About](#1697) as well as a loose copy of the existing GCN header with some changes to the nav to point to the new pages 

image

 @dankocevski


Courey:
Should we pop this out of draft so we can get it reviewed and in so we can both start working on the individual routes? The code looks good to me for a solid base to start with.

Courey (approved):

dankocevski (approved):
Looks good.  We may want to go back and change up the style sheets for the final version, but we can do that at a later date and as a separate task.

lpsinger:
There are some opportunities for code reuse here... I'll post a review as soon as I can.

app/routes/labs.missions.tsx
```diff
@@ -0,0 +1,11 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export default function Missions() {
```
lpsinger, line 9:
```suggestion
export default function () {
```

lpsinger (changes requested):
- [ ] By convention, we don't provide function names for default exports in route modules.
- [ ] You might want to create the placeholder routes for labs/about, labs/conferences, etc. as folders rather than files because most of those routes will probably end up needing images or other assets specific to those routes.
- [ ] Create the placeholder routes as .md or .mdx files for easier authoring?

dakota002:
Addressed the first two points in the last commit, as for the 3rd, @dankocevski do you have preference for the placeholder route files to be markdown (.md or .mdx) or typescript components?

lpsinger (approved):

PR #1708, opened 2023-11-28
base schema, interfaces, and dev tools to create grouping

Courey:
This change has the schema, interfaces, and dev tools necessary across multiple branches.


app.arc
```diff
@@ -111,6 +119,10 @@ circulars
   email *String
   name circularsByEmail
 
+circulars
+  eventId *String
+  name circularEventId
```
lpsinger, line 124:
```suggestion
  name circularsByEventId
```

app/routes/_gcn.synonyms/synonyms.lib.ts
```diff
@@ -0,0 +1,11 @@
+// /*!
+//  * Copyright © 2023 United States Government as represented by the
+//  * Administrator of the National Aeronautics and Space Administration.
+//  * All Rights Reserved.
+//  *
+//  * SPDX-License-Identifier: Apache-2.0
+//  */
```
lpsinger, line 7: Weird comment formatting...

sandbox-search.js
```diff
@@ -10,8 +10,14 @@ import { readFile } from 'fs/promises'
 export default async function () {
   const text = await readFile('sandbox-seed.json', { encoding: 'utf-8' })
   const { circulars } = JSON.parse(text)
-  return circulars.flatMap((item) => [
+  const { synonyms } = JSON.parse(text)
```
lpsinger, line 13: Instead of parsing the file twice, consider:

```ts
const { circulars, synonyms } = JSON.parse(text)
```

sandbox-search.js
```diff
@@ -10,8 +10,14 @@ import { readFile } from 'fs/promises'
...
+  const circularResults = circulars.flatMap((item) => [
     { index: { _index: 'circulars', _id: item.circularId.toString() } },
     item,
   ])
+  const groupResults = synonyms.flatMap((item) => [
+    { index: { _index: 'synonyms', _id: item.id } },
+    item,
+  ])
+  return circularResults.concat(groupResults)
```
lpsinger, line 22: Consider instead:

```ts
return [
  ...(circulars.flatMap(/* fill me in */)),
  ...(synonyms.flatMap(/* fill me in */)),
]
```

lpsinger (changes requested):
- [ ] Does the added sandbox data have an exactly one-to-one mapping between event IDs and synonym table entries? If so, perhaps we don't need to create that data at all. If an event ID is an orphan and has no synonyms, then why store that in the database? The synonyms table exists to define _assocations_ between event IDs.
- [ ] What will be the eventual purpose of mirroring the synonyms table to OpenSearch?

Courey:
1. The synonyms are generated using the eventIds in the seed file. It is NOT just one per synonym. I made sure there were at least 2 with the same synonymId, sometimes more, it varies from 2-4 per group I believe.
2. I added the synonyms to opensearch per your request https://github.com/nasa-gcn/gcn.nasa.gov/pull/1505#issuecomment-1787929058 for searching

lpsinger (approved):

PR #1711, opened 2023-11-30
Courey/adding event id to circulars

Courey:
This adds the eventId to the circular on put and beefs up the subject matchers.

### DIFF
--- a/__tests__/circulars.ts
+++ b/__tests__/circulars.ts
@@ -125,6 +125,24 @@ describe('parseEventFromSubject', () => {
       'Swift GRB-230228A detection of a possibly short burst'
     expect(parseEventFromSubject(grbSubjectWithHyphen)).toBe(grbEvent)
   })
+
+  test('handles GRB event in a Fermi Trigger', () => {
+    const grbInsideFermiTrigger =
+      'Fermi TriggNum 576054086 (GRB 1904204.29) / MASTER OT J071404.52+550304.2: CrAO optical observations'
+    expect(parseEventFromSubject(grbInsideFermiTrigger)).toBe('GRB 1904204.29')
+  })
+
+  test('handles fermi GRB event with no space and a decimal', () => {
+    const fermiGrbWithDecimal =
+      'Fermi GRB190515.49 Global MASTER-Net observations report'
+    expect(parseEventFromSubject(fermiGrbWithDecimal)).toBe('GRB 190515.49')
+  })
+
+  test('handles an ingtegral GRB event', () => {
+    const integralGrbSubject =
+      'Integral GRB221013.04: Global MASTER-Net observations report'
+    expect(parseEventFromSubject(integralGrbSubject)).toBe('GRB 221013.04')
+  })
   //#endregion
 
   //#region SGR
@@ -172,6 +190,14 @@ describe('parseEventFromSubject', () => {
     const sgrSubjectWithSwift = 'SGR Swift J1234+5678 refined analysis'
     expect(parseEventFromSubject(sgrSubjectWithSwift)).toBe(
       'SGR Swift J1234+5678'
+    )
+  })
+
+  test('handles SGR event name with a hyphen and two digits', () => {
+    const sgrSubjectWithHyphenTwoDigits =
+      'SGR 1806-20: Low frequency GMRT results; submission to GCN circulars'
+    expect(parseEventFromSubject(sgrSubjectWithHyphenTwoDigits)).toBe(
+      'SGR 1806-20'
     )
   })
   //#endregion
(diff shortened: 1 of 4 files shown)


lpsinger (changes requested):
Note that you will need to also update https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/_gcn.api.circulars.ts.

Courey:
> Note that you will need to also update https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/_gcn.api.circulars.ts.

I do not believe that is necessary because the[ put that it uses](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/_gcn.api.circulars.ts#L202) checks the subject to get the eventId [on the server here](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1711/files#diff-311cd20aeb3b79775a19354235b7e6334b38b6496ab2b3eaa90fd9c5c8a6119bR276)

app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -36,6 +36,20 @@ type SubjectMatcher = [RegExp, (match: RegExpMatchArray) => string]
 
 const subjectMatchers: SubjectMatcher[] = [
   [/GRB[.\s_-]*(\d{6}[a-z|.]\d*)/i, ([, id]) => `GRB ${id.toUpperCase()}`],
+  [/GRB[.\s_-]*(\d{6}[a-z|.]\d*):/i, ([, id]) => `GRB ${id.toUpperCase()}`],
```
lpsinger, line 39: I think that these are duplicates; strings that match the first of these two regexes will also match the second.

app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -36,6 +36,20 @@ type SubjectMatcher = [RegExp, (match: RegExpMatchArray) => string]
...
+  [
+    /IBAS Alert No[.](\d{4})/i,
+    ([, id]) => `IBAS Alert No. ${id.toUpperCase()}`,
+  ],
+  [/IBAS Alert n[.](\d{4})/i, ([, id]) => `IBAS Alert No. ${id.toUpperCase()}`],
+  [
+    /IBAS Alert n[.] (\d{4})/i,
+    ([, id]) => `IBAS Alert No. ${id.toUpperCase()}`,
+  ],
```
lpsinger, line 48: Combine into a single regex.

app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -36,6 +36,20 @@ type SubjectMatcher = [RegExp, (match: RegExpMatchArray) => string]
...
+  [/XRF(\d{6})/i, ([, id]) => `XRF ${id.toUpperCase()}`],
+  [/XRF (\d{6})/i, ([, id]) => `XRF ${id.toUpperCase()}`],
```
lpsinger, line 50:
```suggestion
  [/XRF\s*(\d{6})/i, ([, id]) => `XRF ${id.toUpperCase()}`],
```
Also: `id.toUpperCase()` will have no effect here, because the group `(\d{6})` matches only digits and not letters.

app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -45,14 +59,26 @@ const subjectMatchers: SubjectMatcher[] = [
   [/ZTF[.\s_-]*(\d{2}[a-z]*)/i, ([, id]) => `ZTF${id.toLowerCase()}`],
   [/HAWC[.\s_-]*(\d{6}A)/i, ([, id]) => `HAWC-${id.toUpperCase()}`],
   [
-    /LIGO\/Virgo[.\s_-]*([S|G|GW])(\d{6}[a-z]+)/i,
+    /\.*LIGO\/Virgo[/KAGRA]?[-_ ]?([S|G|GW])(\d{5,6}[a-z]{0,2}).*/i,
     ([, flag, id]) => `LIGO/Virgo ${flag.toUpperCase()}${id.toLowerCase()}`,
   ],
```
lpsinger, line 64: Why the pattern `\.*`, which will match any repeated sequence of periods?

We should not normalize `LIGO/Virgo/KAGRA 123456` to `LIGO/Virgo 123456`. LIGO, Virgo, and KAGRA are three different GW experiments that share data; removing one of them would very likely rub some people the wrong way.

lpsinger (changes requested):

app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -59,9 +55,14 @@ const subjectMatchers: SubjectMatcher[] = [
...
   [
-    /\.*LIGO\/Virgo[/KAGRA]?[-_ ]?([S|G|GW])(\d{5,6}[a-z]{0,2}).*/i,
+    /LIGO\/Virgo[-_ ]?(S|G|GW)(\d{5,6}[a-z]{0,2}).*/i,
     ([, flag, id]) => `LIGO/Virgo ${flag.toUpperCase()}${id.toLowerCase()}`,
   ],
+  [
+    /LIGO\/Virgo\/KAGRA[-_ ]?([S|G|GW])(\d{5,6}[a-z]{0,2}).*/i,
+    ([, flag, id]) =>
+      `LIGO/Virgo/KAGRA ${flag.toUpperCase()}${id.toLowerCase()}`,
+  ],
```
lpsinger, line 65: I think that you can combine _these_ into a single regex too.

__tests__/circulars.ts
```diff
@@ -466,6 +508,50 @@ describe('parseEventFromSubject', () => {
     expect(parseEventFromSubject(baksanSubjectWithHyphen)).toBe(baksanEvent)
   })
   //#endregion
+
+  //#region XRF
```
lpsinger, line 512: I haven't noticed these `#region` comments before, but I guess they have been here for a while. Can we just put these into jest `describe` blocks rather than relying on magic comments?

lpsinger (changes requested):

__tests__/circulars.ts
lpsinger: Sure.

Courey:
replaced by multiple branches:
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1741
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1731
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1729
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1728
etc

PR #1722, opened 2023-12-05
index without working links or create button

Courey:
Screenshot 2023-12-05 at 12 39 51 PM

Links to individual synonym groups don't work yet. 
New button does not work yet

This is behind both a feature flag and a moderator check.


lpsinger:
@Courey, how is this related to #2617?

Courey:
> @Courey, how is this related to #2617?

They

PR #1731, opened 2023-12-07
adding cases where there are six digits and a letters at the end are …

Courey:
…optional

### DIFF
--- a/app/routes/_gcn.circulars/circulars.lib.ts
+++ b/app/routes/_gcn.circulars/circulars.lib.ts
@@ -45,7 +45,7 @@ const subjectMatchers: SubjectMatcher[] = [
   [/ZTF[.\s_-]*(\d{2}[a-z]*)/i, ([, id]) => `ZTF${id.toLowerCase()}`],
   [/HAWC[.\s_-]*(\d{6}A)/i, ([, id]) => `HAWC-${id.toUpperCase()}`],
   [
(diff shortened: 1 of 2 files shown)


app/routes/_gcn.circulars/circulars.lib.ts
```diff
@@ -45,7 +45,7 @@ const subjectMatchers: SubjectMatcher[] = [
   [/ZTF[.\s_-]*(\d{2}[a-z]*)/i, ([, id]) => `ZTF${id.toLowerCase()}`],
   [/HAWC[.\s_-]*(\d{6}A)/i, ([, id]) => `HAWC-${id.toUpperCase()}`],
   [
-    /((?:LIGO|Virgo|KAGRA)(?:[/-](?:LIGO|Virgo|KAGRA))*)[-_ \s]?(S|G|GW)(\d{5,}[a-z]+)/i,
+    /((?:LIGO|Virgo|KAGRA)(?:[/-](?:LIGO|Virgo|KAGRA))*)[-_ \s]?(S|G|GW)(\d{5,6}[a-z]{0,2})/i,
```
Courey, line 48: could there be more than two letters at the end?

Courey: it depends, are these valid subjects?
'LIGO/Virgo GW170817: Chandra X-ray brightening of the counterpart 108 days since merger'
''LIGO/Virgo G211117: Further Swift-XRT sources'
If so, the PR is still necessary because it's not matching the letters as optional, it requires 5 or more digits followed by letters.

Courey: I updated it to make it so that it's 0 or more letters at the end.

lpsinger (approved):

PR #1732, opened 2023-12-08
adding IBAS alert to subject matcher

Courey:


lpsinger (commented):
I'm going to need some help on this one. I'm not too familiar with INTEGRAL or IBAS. Eric would be a good reviewer for this, though I don't know his GitHub handle.

Courey:
@eburnsastro Would you have any input on these circulars? I looked through the subjects of the existing circulars and it looks like for ibas we get 4 patterns that are specific to them:
'IBAS Alert n. 3422: not a GRB'
'IBAS Alert n.3422: not a GRB'
'IBAS Alert No. 3422: not a GRB'
'IBAS Alert 3422: not a GRB'

is it appropriate to give these type of circulars an event id, when most IBAS circulars come in with subjects like 'Bursting activity from SGR 1806-20 detected with IBAS' and would then have the event id like SGR 1806-20? The "not a GRB' ones would have a different format. I am not sure this is correct. would it be better to not have an event id at all on these?

eburnsastro:
My suggestion would be to either avoid giving these items an event ID or give them some filler that makes them obvious we are not classifying them as GRBs, GWs, etc.

The reason for this is because there are many strange edge cases and it is not obvious how to easily handle them. Fermi-GBM, Swift-BAT, and INTEGRAL (effectively the instruments with fully automated GRB reporting) all have circulars which are 'this trigger is not a GRB'. In some cases the trigger has a different source class (magnetar (soft gamma-ray repeater (SGR)) flare, solar flare, particle activity). Some of these we could nominally give an event name, but even this is extremely difficult. SGR flare naming conventions do not exist (i.e. SGR 1806-20 is one of the objects these flares can come from, but the event is a specific flare, which doesn't have an agreed upon designation (IPN will sort this out next year)). Solar flares probably have a naming convention, but GCN isn't an appropriate venue for them. There are also cases where the trigger does not have an astrophysical origin, or even has an unknown origin. 

What may be useful to do is to use these to remove a given event ID from a previous trigger. That is, in the above example, presumably we had a notice /alert that listed IBAS Alert n. 3422 as a GRB, and the noted marking as 'not a GRB' could break a given trigger ID to event ID matching. This is probably not a priority as these can be manually fixed given the fairly low rate of them

lpsinger:
What's up with this PR? No one has touched it in months.

Courey:
closing as @eburnsastro said "avoid giving these items an event ID" was a valid path.

PR #1743, opened 2023-12-15
synonym overview with edits and delete

Courey:
If you have data that needs to be saved, the save button is blue
Screenshot 2023-12-15 at 12 54 44 PM
if you don't have any data that needs to be saved, the button is disabled
Screenshot 2023-12-15 at 12 54 30 PM

### DIFF
--- a/app/routes/_gcn.synonyms.$synonymId._index.tsx
+++ b/app/routes/_gcn.synonyms.$synonymId._index.tsx
@@ -1,0 +1,210 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { type DataFunctionArgs, redirect } from '@remix-run/node'
+import { Form, Link, useLoaderData } from '@remix-run/react'
+import {
+  Button,
+  ButtonGroup,
+  Card,
+  CardBody,
+  CardGroup,
+  CardHeader,
+  FormGroup,
+  GridContainer,
+  Icon,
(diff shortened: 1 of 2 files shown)


lpsinger:
I see that you've requested a review on this PR, but I wanted to review the backend implementation independently from the front end design first. Which PR was that?

Courey:
> I see that you've requested a review on this PR, but I wanted to review the backend implementation independently from the front end design first. Which PR was that?

It's been available here https://github.com/nasa-gcn/gcn.nasa.gov/pull/1505/files#diff-311cd20aeb3b79775a19354235b7e6334b38b6496ab2b3eaa90fd9c5c8a6119bL17 
I cut and pasted it into this branch
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1803/files

jracusin:
The kinds of events that are synonyms are different names for the same events, so I'm a bit confused by your example screenshots.  Those are all separate GRBs, so therefore they are different events and would never be synonyms.  How do we also know what event IDs are available?

Courey:
> The kinds of events that are synonyms are different names for the same events, so I'm a bit confused by your example screenshots. Those are all separate GRBs, so therefore they are different events and would never be synonyms. How do we also know what event IDs are available?

I randomly generated groupings of 2-4 events. So they are there just as functional placeholders unless someone wants to tell me which events in our seed file are related to each other. Apologies if that is confusing, I don't have any concept of what events are related in our seed file locally to make it real relationships.

What would moderators want to be able to tell what eventIds there are? I currently have a search box that you enter the eventId into. What typically are the moderators going to know vs what are they going to need to look up?

Courey:
Closing this as it got split into two, the front and back end.

PR #1769, opened 2023-12-21
Circulars - move date selector component outside of route.tsx to its own file

tylerbarna:
Similar to what was done with the DateSelectorButton component or the CircularsIndex, this PR separates the date selector menu into a separate file outside of route.tsx. This should make it somewhat easier to parse the route file by moving a bit over 150 lines to this separate file. 

Only the start and end date parameters are passed as arguments to the component; I wasn't sure if searchParams itself should be passed.

### DIFF
--- a/app/routes/_gcn.circulars._archive._index/DateSelectorMenu.tsx
+++ b/app/routes/_gcn.circulars._archive._index/DateSelectorMenu.tsx
@@ -1,0 +1,162 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { useSearchParams, useSubmit } from '@remix-run/react'
+import {
+  Button,
+  CardBody,
+  CardFooter,
+  DateRangePicker,
+  Grid,
+  Icon,
+  Radio,
+} from '@trussworks/react-uswds'
+import { useState } from 'react'
+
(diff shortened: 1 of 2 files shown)


app/routes/_gcn.circulars._archive._index/DateSelectorMenu.tsx
```diff
@@ -0,0 +1,162 @@
...
+  Grid,
+  Icon,
+  Radio,
+} from '@trussworks/react-uswds'
+import { useState } from 'react'
+
+import DateSelectorButton from './DateSelectorButton'
+import DetailsDropdownContent from '~/components/DetailsDropdownContent'
+
+export default function ({
```
lpsinger, line 23: We tend not to have default exports in library files.

```suggestion
export function DateSelector ({
```

lpsinger (changes requested):

app/routes/_gcn.circulars._archive._index/DateSelectorMenu.tsx
tylerbarna: Ah, I was taking a cue from the way the DateSelectorButton file was written

lpsinger:
Prettier is complaining about something. What is it upset about?

tylerbarna:
> Prettier is complaining about something. What is it upset about?

I'm trying to figure out what's causing it, Prettier ran fine locally when I made the commit, and the original commit also passed.

tylerbarna:
@lpsinger there was a space in the export function line, should be fixed now

lpsinger (approved):

PR #1780, opened 2023-12-25
Adjust z-order of Circulars search bar

lpsinger:
Improper z-order caused it to appear on top of the main nav on screen sizes smaller than desktop.

# Before

Screenshot 2023-12-25 at 14 03 47

# After

Screenshot 2023-12-25 at 14 00 03


tylerbarna (approved):
I believe we set the z-order as `z-top` since the footer was obscuring the date selector menu in certain cases, but `z-300` seems to fix that without also drawing it over the menu:
image

PR #1783, opened 2023-12-27
Refactor DateSelectorButton, move inside DateSelectorMenu

lpsinger:
We will never need to use `DateSelectorButton` outside of `DateSelectorMenu`.

This refactoring allows us to remove duplicate definitions of `dateSelectorLabels`.


tylerbarna (approved):
> We will never need to use `DateSelectorButton` outside of `DateSelectorMenu`.

Good point, this removes a redundancy left over from #1769

PR #1802, opened 2024-01-03
Add optional revisedOn field to circular and update approveChangeRequest method

dakota002:


lpsinger (changes requested):
I like this, but could we rename the `editedBy` field to `revisedBy` for consistency? Would that require any database migration?

Alternatively, we could rename `revisedOn` to `editedOn`.

lpsinger:
> > I like this, but could we rename the `editedBy` field to `revisedBy` for consistency? Would that require any database migration?
> > Alternatively, we could rename `revisedOn` to `editedOn`.
> 
> I think the latter will be easier, thought I also don't think there would be any migration required with the former

We don't have any existing database entries that would contain that field?

dakota002:
> > > I like this, but could we rename the `editedBy` field to `revisedBy` for consistency? Would that require any database migration?
> > > Alternatively, we could rename `revisedOn` to `editedOn`.
> > 
> > 
> > I think the latter will be easier, thought I also don't think there would be any migration required with the former
> 
> We don't have any existing database entries that would contain that field?

We shouldn't since this is in support of the versioning feature which isn't out yet

lpsinger:
> > > > I like this, but could we rename the `editedBy` field to `revisedBy` for consistency? Would that require any database migration?
> > > > Alternatively, we could rename `revisedOn` to `editedOn`.
> > > 
> > > 
> > > I think the latter will be easier, thought I also don't think there would be any migration required with the former
> > 
> > 
> > We don't have any existing database entries that would contain that field?
> 
> We shouldn't since this is in support of the versioning feature which isn't out yet

Alright. As for the naming of these two fields, which names (or what names) would be most similar to terminology used in the [DataCite REST API](https://support.datacite.org/docs/api)?

dakota002:
> > > > > I like this, but could we rename the `editedBy` field to `revisedBy` for consistency? Would that require any database migration?
> > > > > Alternatively, we could rename `revisedOn` to `editedOn`.
> > > > 
> > > > 
> > > > I think the latter will be easier, thought I also don't think there would be any migration required with the former
> > > 
> > > 
> > > We don't have any existing database entries that would contain that field?
> > 
> > 
> > We shouldn't since this is in support of the versioning feature which isn't out yet
> 
> Alright. As for the naming of these two fields, which names (or what names) would be most similar to terminology used in the [DataCite REST API](https://support.datacite.org/docs/api)?

I don't see any attributes that are really similar. They have a `date` and a `version` property, but i don't see anything about editors, revisions, or dates relating to those ideas

lpsinger (approved):

PR #1478, opened 2023-10-04
adding functionality to allow moderators to add eventId and synonyms

Courey:
This is PR 1 of 3. Once both PRs are in, this body of work must be connected to the synonyms table/index. I have noted where and what change needs to be made.

PR #1803, opened 2024-01-04
server only implementation of synonyms

Courey:


lpsinger:
There are several related draft PRs: #1505, #1479, #1478. Which ones of these are still in play?

Courey:
All drafts are no longer relevant. Those were previous designs that got nixed or were broken up into smaller PRs. I kept the drafts around because it's much easier to see clearly what I did in which branch if I have a comparison handy. 

[1505](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1505/files) is the feature branch that has everything in it and would be what you would look at to see how the functions get used. I have been stealing functionality from that branch for the current open smaller PRs. Those smaller PRs have their server code with them though, so it is clear how they are being used.

Courey:
The missing part that you would only find in 1505 is the front end part of the actual grouping of the circulars. I haven't made a smaller branch for that because I needed to get in the making of the synonym before I put in the using of the synonym. I do not have an open branch with front end for the grouping separated out of 1505 yet as it would share a lot of the base code that the other branches use and I didn't want to duplicate or have complex merging orders etc.

Courey:
this is stale. will update and try again in smaller PRs

PR #1807, opened 2024-01-05
Change deploy workflow to build and deploy in arm64 docker

jak574:
Pull request makes a first pass at changing the `deploy.yml` file, changing it so that `npm ci`, `npm run build` and `npm run deploy` are executed inside a Docker container running arm64 linux under QEMU. This fixes (maybe) the issue where the current deployment can only use binary packages from pip, making it break if no binary is available for the `manylinux_aarch64` architecture.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,21 +24,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install Node.js
-        uses: actions/setup-node@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
         with:
-          node-version: '18'
-      - uses: actions/setup-python@v5
-        with:
-          python-version: '3.11'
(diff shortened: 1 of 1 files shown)


jak574:
Quick cautionary comment. It looks like environment variables get passed to the docket container via the command line, so if there's any credentials or anything else you don't want made public stored in env, beware.

jak574:
Update on the env issue. I guess it's not passing *every* environoment variable, but a select few. So if credentials are passed via env variable (I think they are), we need to pass them to the docker, using `env:`, e.g. like here: https://github.com/addnab/docker-run-action/issues/21#issuecomment-873058697

lpsinger:
I'm going to see if we can run just the hydrate step under emulation, and the rest natively. That way we would avoid passing most environment variables to containers.

lpsinger:
https://github.com/architect/hydrate/pull/184

lpsinger:
This is now fixed on the main branch. Thanks for looking into this!

PR #1818, opened 2024-01-10
ACROSS API: Add Ephem class and endpoints

jak574:
# Add Ephem to ACROSS API

This pull request adds TLE propagation, and adds `Ephem` endpoints for BurstCube and Swift. The code introduces one new class:

### `EphemBase`

This base class, inherited by `SwiftEphem` and `BurstCubeEphem` calculates a position and velocity vectors for the spacecraft, and calculates the relative positions of the Earth, Sun and Moon as seen by the spacecraft. In order to derive these, we create an `astroplan` observer at the location of the spacecraft, as derived from the TLE.

Derived properties include the `beta` angle (angle between the Sun and the orbit plane), and a unit vector that points to the orbit pole (handy for calculating the location of continuous viewing zones or for spacecraft with large earth limb avoidance constraints like Swift, the location of the pole constraint).

Ephemerides are calculated between the given `begin` and `end` dates, with a time resolution defaulting to 60 seconds. Begin/End dates are rounded to the resolution of the calculation, so a stepsize of 60 seconds will always have ephemerides that start at then start of a minute, in UTC.

The Ephemeris also calculates the radius of the Earth as seen from the spacecraft, although there is an option to override this calculation with a fixed value, if `earth_radius` is set.


lpsinger (commented):
As usual, the overall code style and quality is great. I have slightly more general versions of the observer class and the Earth limb constraint that we could use:

- https://github.com/nasa/dorado-scheduling/tree/main/dorado/scheduling/orbit
- https://github.com/nasa/dorado-scheduling/tree/main/dorado/scheduling/constraints

jak574:
> As usual, the overall code style and quality is great. I have slightly more general versions of the observer class and the Earth limb constraint that we could use:
> 
> * https://github.com/nasa/dorado-scheduling/tree/main/dorado/scheduling/orbit
> * https://github.com/nasa/dorado-scheduling/tree/main/dorado/scheduling/constraints

OK. Never seen this stuff before BTW, I was looking at your m4opt repo. 

Playing with it a bit I think it'd be pretty simple to adapt my code to use this. Question is, how would you want me to integrate it into the GCN / ACROSS repo as I can't just pip install it?

jak574:
@lpsinger in 3e1d629 I modified the Ephem generation code minimally to use the code from your dorado-scheduling repo. I just copied over just what I needed into a directory called scheduling.

python/across_api/base/api.py
```diff
@@ -34,3 +35,38 @@ async def epoch(
 
 
 EpochDep = Annotated[datetime, Depends(epoch)]
+
+
+# Depends functions for FastAPI calls.
+async def daterange(
```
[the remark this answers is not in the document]
jak574: This is really just from using examples from the FastAPI documentation. 

https://fastapi.tiangolo.com/tutorial/dependencies/

However, I think as the library calls we're making in the API endpoints don't themselves support async operation, it's probably redundant to have async in all cases in the API endpoint code, so we could just remove all instances of it.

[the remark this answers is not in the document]
lpsinger: The docs have some guidance on this matter: https://fastapi.tiangolo.com/tutorial/dependencies/#to-async-or-not-to-async

I suggest that as a general rule we only make functions `async` if they actually contain asynchronous code (e.g., contain `await` statements). That way they are callable from ordinary Python functions (even functions that are _not_ FastAPI route handlers) when necessary.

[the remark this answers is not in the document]
jak574: Some interesting discussion on using `async def` in FastAPI here: https://github.com/tiangolo/fastapi/discussions/8313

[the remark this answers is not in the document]
jak574: None of the code with async is used for anything except API endpoints. Those Depends functions are specifically to reduce repetitive code in API endpoints, and aren't used elsewhere.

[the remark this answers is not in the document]
jak574: 9cb8ac0 strips out all the async. We can readdress the use of async later on.

python/across_api/base/api.py
```diff
@@ -34,3 +35,38 @@ async def epoch(
 
 
 EpochDep = Annotated[datetime, Depends(epoch)]
+
+
+# Depends functions for FastAPI calls.
```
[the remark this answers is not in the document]
jak574: cc7b533

jak574:
I updated the description of the pull, as I remove the `EarthSatelliteLocation` class, and replaced it with `astroplan` `Observer`.

python/across_api/base/common.py
```diff
@@ -16,6 +19,28 @@
...
+    step
+        The time step to round to.
+
+    Returns
+    -------
+        The rounded Time.
+    """
+    return Time(
+        np.round(t.unix / step.to_value(u.s)) * step.to_value(u.s), format="unix"
+    )
```
lpsinger, line 41: See https://docs.astropy.org/en/stable/time/#precision

lpsinger (changes requested):
My impression was that for BurstCube we have to add CRUD endpoints for targets. Do we actually need API endpoints for the ephemeris?

python/across_api/base/common.py
jak574: Precision is about decimal places when printing, right? This is about rounding the actual time value to whole seconds and whole minutes.

jak574:
> My impression was that for BurstCube we have to add CRUD endpoints for targets. Do we actually need API endpoints for the ephemeris?

The question as to whether we actually need endpoints for Ephemeris calculations is a fair one. I would argue that we don't loose anything by providing it, but will admit it's fair to question it's usefulness overall. My feeling is that someone might want to use the information in the endpoint for their own calculations, so by providing the information we allow that type of usage.

jak574:
> > > My impression was that for BurstCube we have to add CRUD endpoints for targets. Do we actually need API endpoints for the ephemeris?
> > 
> > 
> > The question as to whether we actually need endpoints for Ephemeris calculations is a fair one. I would argue that we don't loose anything by providing it, but will admit it's fair to question it's usefulness overall. My feeling is that someone might want to use the information in the endpoint for their own calculations, so by providing the information we allow that type of usage.
> 
> If we expose it as an API endpoint, then we have to make sure that we preserve compatibility when we make changes in the future. Given that we are probably going to do a massive refactoring shortly after BurstCube's launch, I suggest exposing only the API endpoints that BurstCube really needs and will use.

I talked with the BurstCube team about their needs, and they don't need the Ephemeris generation, so I took it out (also for Swift) in 631ba8a.

python/across_api/base/ephem.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Calculate the Earth radius in degrees
+        if self.earth_radius is not None:
+            self.earthsize = self.earth_radius * np.ones(len(self))
+        else:
+            self.earthsize = np.arcsin(R_earth / dist)
+
+        # Calculate orbit pole vector
+        polevec = self.posvec.cross(self.velvec)
+        self.pole = SkyCoord(polevec / polevec.norm())
```
lpsinger, line 191: What do you need this for?

jak574: For BurstCube, we don't. For other missions it's useful to know where the orbit poles are so you can calculate if a target is inside an orbit pole constraint (Swift has these) or for other missions if targets are inside a continuous viewing zone.

For BurstCube only, it can be removed.

lpsinger: If we needed an orbit pole constraint, then I would suggest implementing that as a reusable constraint class.

lpsinger: Never mind, this has to do with where code goes, which we agreed was off limits for review for now.

lpsinger (changes requested):
Please fix CI failures.

python/across_api/base/ephem.py
jak574: Yeah, it's additional CPU time that's not needed for every mission, so makes sense to move it out of Ephem, especially for missions that don't need it. In the original code I only calculated it for missions that would use it, but in the spirit of code simplification I took that option out. I'll just remove it, no biggy.

lpsinger (approved):

lpsinger (changes requested):
Darn. The Python CI workflow is still failing.

jak574:
> Darn. The Python CI workflow is still failing.

I'm investigating. Not sure why it would fail at the black step, I'm running black locally before submitting requests.

jak574:
> Darn. The Python CI workflow is still failing.

OK fixed in 8124c0c.

jak574:
OK I found a problem with this tonight while setting up unit tests. The velocity values are wrong. 

I was getting `EarthLocation` using the `earth_location` method of the spacecraft in `ITRS` coordinates, then using `get_gcrs_posvel` method of `EarthLocation` to get position and velocity. Problem is that despite giving the correct position for the spacecraft, the velocity is wrong, like an order of magnitude off. So spacecraft velocity is not preserved.

So I'm reverting to calculating position and velocity by doing a transform from ITRS to GCRS. I'm also removing the `astroplan` dependency as it's only providing `earth_location` which can be obtained from `ITRS` (albeit with the wrong velocity).

jak574:
I think my original solution of having a replacement for `EarthLocation` called `EarthSatelliteLocation` which reports back correct values for velocity might be a better solution in the long term. `EarthLocation` basically seems to take the GCRS position and then use a fixed-to-earth body rotation for velocity, which of course is only going to be right for satellites in GEO!

jak574:
BTW I added back endpoints, but they appear in the /testing route, and only when `ARC_ENV == 'testing`. It's really frustrating trying to debug stuff without at least the endpoints, so this seems like a good solution.

lpsinger:
> BTW I added back endpoints, but they appear in the `/testing` route, and only when `ARC_ENV == 'testing'`. It's really frustrating trying to debug stuff without at least the endpoints, so this seems like a good solution.

Alright, let's add a pytest test suite instead of these testing endpoints. Do you have a branch that has a start on that?

jak574:
> I feel strongly about this. 

8fdeeba

jak574:
> Alright, let's add a pytest test suite instead of these testing endpoints. Do you have a branch that has a start on that?

https://github.com/nasa-gcn/gcn.nasa.gov/pull/1878 

When this gets merged, I'll mark it as ready for review.

python/across_api/base/ephem.py
```diff
@@ -0,0 +1,186 @@
...
+)
+from astropy.time import Time  # type: ignore
+from fastapi import HTTPException
+from sgp4.api import Satrec  # type: ignore
+
+from ..base.schema import EphemGetSchema, EphemSchema, TLEEntry
+from .common import ACROSSAPIBase, round_time
+
+
+class EphemBase(ACROSSAPIBase):
```
lpsinger, line 26: I thought that we were removing the ephemeris endpoint. Why does this still inherit from `ACROSSAPIBase`?

jak574: The code breaks if I remove it, and it does no harm, so let's leave that in.

python/across_api/base/schema.py
```diff
@@ -36,6 +122,68 @@ class BaseSchema(BaseModel):
...
+
+    Parameters
+    ----------
+    stepsize
+        The step size in seconds
+
+    """
+
+    stepsize: AstropySeconds
+    ...
```
lpsinger, line 184: We could safely remove these since they aren't exposed by the REST API.

jak574: Causes the same problem as above, breaking working code leading to a required refactor, but OK whatever, I removed a bunch of API code in 1ce600b and b9d08d2.

python/across_api/base/api.py
```diff
@@ -34,3 +39,37 @@ async def epoch(
...
+            ge=1,
+            title="Step Size",
+            description="Time resolution in which to calculate result in seconds.",
+        ),
+    ] = 60,
+) -> int:
+    return stepsize * u.s
+
+
+StepSizeDep = Annotated[int, Depends(stepsize)]
```
swyatt7, line 75: I'm confused on why these belong in `api.py`? Should we have a separate file for dependencies? 

Only routed endpoints should exist in our `api.py` files. Maybe even move the `app` instantiation to a separate file as well.

python/across_api/base/ephem.py
```diff
@@ -0,0 +1,161 @@
...
+    begin: Time
+    end: Time
+    stepsize: u.Quantity
+    earth_radius: Optional[u.Quantity] = None
+    tle: Optional[TLEEntry] = None
+
+    def __len__(self) -> int:
+        return len(self.timestamp)
+
+    def ephindex(self, t: Time) -> int:
```
swyatt7, line 74: Is this where we should use our `EpochDep` definition that is currently in `api.py`?

python/across_api/base/ephem.py
```diff
@@ -0,0 +1,161 @@
...
+        t
+            The time to find the nearest index for.
+
+        Returns
+        -------
+            The index of the nearest time in the ephemeris.
+        """
+        return int(np.argmin(np.abs((self.timestamp.datetime - t.datetime))))
+
+    def __init__(self, begin: Time, end: Time, stepsize: u.Quantity = 60 * u.s):
```
swyatt7, line 91: Is this where we should use our `DateRangeDep` and `StepSizeDep` definitions that are currently in `api.py`?

swyatt7 (changes requested):
Can we also have our file structure be something like:
```
across_api/
   across/  -> maybe we won't need this, seems to be example tutorial copy paste mostly?
   auth/
   base/
   missions/ -> or maybe 'observatories/' to preserve generalized future implementation of ground based observatories 
      burstcube/
      swift/
      [subsequent missions]/
```

It will make it easier to follow where missions are and how we can separate base dependencies from their independent implementations on the missions.

python/across_api/base/api.py
jak574: I'm only using dependency injection in the API endpoints, not in the main code. This is why they live in `api.py`, it just seemed like a natural fit for anything API related. I'm not against moving them elsewhere, perhaps to `depends.py`?

FYI: Originally this code had an Ephem API endpoint, but I took it out. This `StepSizeDep` is a remnant of that API endpoint, so I just removed it in 35fc4c9.

python/across_api/base/ephem.py
jak574: This is why I should read all comments before taking action. `EpochDep` is also a orphaned depency. I removed it in 46df87a.

python/across_api/base/ephem.py
jak574: As above, those are only used in API endpoints, not in the ephem code.... and yes, `DateRangeDep` is another orphaned Depend from the removed Ephemeris API endpoint, so I removed that also in c68918e.

jak574:
> Can we also have our file structure be something like:
> 
> ```
> across_api/
>    across/  -> maybe we won't need this, seems to be example tutorial copy paste mostly?
>    auth/
>    base/
>    missions/ -> or maybe 'observatories/' to preserve generalized future implementation of ground based observatories 
>       burstcube/
>       swift/
>       [subsequent missions]/
> ```
> 
> It will make it easier to follow where missions are and how we can separate base dependencies from their independent implementations on the missions.

My idea was that the across directory would be where non-mission or observatory specific functions associated with ACROSS itself would go. Right now we have the name resolver Endpoint in there, but I plan in future to have other endpoints that provide tools that are not mission specific.

jak574:
I'm generally fine with the idea of putting missions in their own sub directory, should that follow for endpoints do you think? My general structure here is to have codebase match endpoints. 

Then a more general question, what should the categories be?

I can see "missions" and "groundbased", or maybe you want to divide up missions by main agency, so like `/nasa/fermi`. Instead of missions, should we say "space" and "ground"?

swyatt7 (approved):
Everything looks good to me, let's raincheck the file - restructuring like we discussed. It was just my serious first time looking at the codebase and I may have been a bit ambitious.

PR #1829, opened 2024-01-15
Circulars Sort Results by Relevance

tylerbarna:
Currently, searching for circulars sorts by circular ID (ie newest to oldest), but it's also possible to sort by relevance; I'm unsure if it would be better to sort by relevance once we implement the NLP search functionality

Searching 'GRB' and sorting by ID:
image

Searching 'GRB' and sorting by relevance:
image

There's some balance to be made between showing the most relevant results and also weighting the most recent results, since they're arguably the circulars users are more likely to search for.

Another option would be to default to relevance and have the option to sort by circular ID if a toggle is set; this could be accomplished using the sortable tables, which I've been experimenting with in #1816


lpsinger (changes requested):
This is a nice idea. Would you please add a dropdown menu that switches between sort by circular ID and relevance?

tylerbarna:
I tried to do something similar to the date range filtering button, and the component is pretty similar as a result. I wanted to try to have it so there was a button to change the order of the results to be ascending instead of descending, but that isn't fully functional at the moment.
image

Also need to figure out a better looking menu; I used the DetailsDropdownContent component, but I'm not entirely sure if that's the best option here; would be nice to try to figure out a way to use the Select component while still being able to have it appear as a standard button.

image

lpsinger (changes requested):
- [ ] Remove commented out code
- [ ] Use radio buttons instead of links
- [ ] Don't manually construct the query string; the values of the radio buttons will automatically be included in the form request on submission
- [ ] Let's _not_ add support for selecting sort order right now; there's no reason to support sorting by _decreasing_ relevance.

tylerbarna:
switched to radio buttons
image

tylerbarna:
It occurs to me that it would be useful to have some sort of listener that automatically collapses any divs if the user clicks away from them to prevent menus from drawing over each other like this

image

app/routes/_gcn.circulars._archive._index/route.tsx
```diff
@@ -152,6 +156,7 @@ export default function () {
         {featureCircularsFilterByDate && (
           <DateSelector startDate={startDate} endDate={endDate} />
         )}
+        {featureCircularsSort && <SortSelector />}
```
lpsinger, line 159: No need for a feature flag here.

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <>
+      {sortOptions.map(({ id, value, label }) => (
+        <Radio
+          key={id}
+          id={id}
+          name="sort"
+          value={value}
+          label={label}
+          form="searchForm"
+          checked={inputSort === value}
```
lpsinger, line 63: Don't set `checked`; set `defaultChecked`.

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+        Sort by...
+      </Button>
+      <Button type="button" className={`${slimClasses} padding-x-2`}>
+        {<Icon.FilterList role="presentation" />}
+      </Button>
+    </ButtonGroup>
+  )
+}
+
+export function SortSelector({ sort = 'circularId' }: { sort?: string }) {
```
lpsinger, line 41:
```suggestion
export function SortSelector({ sort = 'circularId' }: { sort?: string }) {
```

How about `defaultValue` instead of `sort`?

And define the type as:

```ts
type SortOrder = 'circularId' | 'relevance'

...

function SortSelector({value}: {value?: SortOrder})`
```

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const SortRadioButtons = () => (
+    <>
+      {sortOptions.map(({ id, value, label }) => (
+        <Radio
+          key={id}
+          id={id}
+          name="sort"
+          value={value}
+          label={label}
+          form="searchForm"
```
lpsinger, line 62: `SortSelector` should take this as a prop.

lpsinger (changes requested):

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+    if (sortInputRef.current) {
+      sortInputRef.current.value = value
+    }
+  }
+
+  function radioOnChange({ target: { value } }: ChangeEvent<HTMLInputElement>) {
+    setSortOrder(value)
+    setShowContent(false)
+    const form = sortInputRef.current?.form
+    if (form) console.log(form)
```
lpsinger, line 78:
```suggestion
```

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+  ]
+
+  function setSortOrder(value: string) {
+    if (sortInputRef.current) {
+      sortInputRef.current.value = value
+    }
+  }
+
+  function radioOnChange({ target: { value } }: ChangeEvent<HTMLInputElement>) {
+    setSortOrder(value)
```
lpsinger, line 75:
```suggestion
  function radioOnChange({ target: { value } }: ChangeEvent<HTMLInputElement>) {
```

lpsinger (changes requested):

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [showContent, setShowContent] = useState(false)
+
+  const sortInputRef = useRef<HTMLInputElement>(null)
+
+  const submit = useSubmit()
+
+  const sortOptions = [
+    { id: 'radio-sort-circularId', value: 'circularID', label: 'Circular' },
+    { id: 'radio-sort-relevance', value: 'relevance', label: 'Relevance' },
+  ]
```
lpsinger, line 66: Lift this data up to module scope so that both this component and `SortButton` can use it.

lpsinger (changes requested):

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,113 @@
...
+} & Omit<Parameters<typeof ButtonGroup>[0], 'segmented' | 'children'>) {
+  const slimClasses = 'height-4 padding-y-0'
+
+  return (
+    <ButtonGroup type="segmented" {...props}>
+      <Button type="button" className={`${slimClasses} padding-x-2`}>
+        Sorted By{' '}
+        {sortOptions.find((option) => option.value === sort)?.label ||
+          'Circular'}
+      </Button>
```
lpsinger, line 44: Make this an object keyed by value so that you do not have to do a linear search to find the label:

```suggestion
const sortOptions = {
  circularID: 'Circular',
  relevance: 'Relevance',
}

function SortButton({
  sort,
  expanded,
  ...props
}: {
  sort?: string
  expanded?: boolean
} & Omit<Parameters<typeof ButtonGroup>[0], 'segmented' | 'children'>) {
  const slimClasses = 'height-4 padding-y-0'

  return (
    <ButtonGroup type="segmented" {...props}>
      <Button type="button" className={`${slimClasses} padding-x-2`}>
        Sorted By {sortOptions[sort ?? 'circular']}
      </Button>
```

Also I think it will be more clear if we call this Sort by Date rather than Sort by Circular.

lpsinger (changes requested):

tylerbarna:
the failing check seems to be a network error?

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+    setShowContent(false)
+    if (target.form) submit(target.form)
+  }
+
+  const SortRadioButtons = () => (
+    <>
+      {Object.entries(sortOptions).map(([value, label]) => (
+        <Radio
+          key={`radio-sort-${value}`}
+          id={`radio-sort-${value}`}
```
lpsinger, line 74:
```suggestion
          key={value}
          id={value}
```

app/routes/_gcn.circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  sort?: string
+  expanded?: boolean
+} & Omit<Parameters<typeof ButtonGroup>[0], 'segmented' | 'children'>) {
+  const slimClasses = 'height-4 padding-y-0'
+
+  return (
+    <ButtonGroup type="segmented" {...props}>
+      <Button type="button" className={`${slimClasses} padding-x-2`}>
+        Sorted By{' '}
+        {sortOptions[(sort as keyof typeof sortOptions) || 'circularID']}
```
lpsinger, line 39: `sort` is unsanitized input from the user (it's from the URL query string), so you cannot assume that `sort` is one of the keys of `sortOptions`.

lpsinger (changes requested):

lpsinger (approved):

PR #1834, opened 2024-01-16
Sign out other sessions of user on password change

lpsinger:
Note: do not merge. This does not work as intended. It signs out _all_ of the user's sessions, including the current session. There is no obvious way to sign out only _specific_ OIDC/OAuth2 sessions of the user.

Also, this change wouldn't address the similar cases that the user changes the password through the Cognito hosted "forgot password" form.

### DIFF
--- a/app/routes/_gcn.user.password/password.server.ts
+++ b/app/routes/_gcn.user.password/password.server.ts
@@ -5,7 +5,12 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import { ChangePasswordCommand } from '@aws-sdk/client-cognito-identity-provider'
+import { tables } from '@architect/functions'
+import {
+  ChangePasswordCommand,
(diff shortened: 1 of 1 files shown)


lpsinger:
@dakota002: I was thinking about how, after we migrate from client credentials to refresh tokens, we are going to implement deletion of refresh tokens. That reminded me of this issue, which I stumbled on because I couldn't figure out how to expire a single refresh token programmatically. But I don't see why we can't just make a request to the Cognito-hosted `logout` endpoint like we do in [app/routes/_auth.logout.ts](https://github.com/nasa-gcn/gcn.nasa.gov/blob/8a5e3249d1a7904b00447e403e5de0c0b505b84f/app/routes/_auth.logout.ts).

Would you please try to revive this and complete it, as a proof of concept of how to programmatically expire refresh tokens?

PR #1805, opened 2024-01-05
adding mission timeline

Courey:
image

PR #1840, opened 2024-01-18
Test build and Python dep installation in PR workflow

lpsinger:
This will catch errors such as missing imported assets. For example: https://github.com/nasa-gcn/gcn.nasa.gov/pull/1805


Kirill-Vorobyev (approved):
Confirmed `npm run build` ran as RP CI check
image

PR #1858, opened 2024-01-23
New announcements

### LINKED ISSUE #773: Ability for GCN Team to send Announcements and users opt in/out
- announcements email list managed like circulars list
- toggle button to subscribe/unsubscribe
- admin & moderator roles for website
- user interface for admins and moderators to send announcements

dakota002:
resolves #773 

Updated user email notification page:
image

Announcement email form:
image

### DIFF
--- a/app/routes/_gcn.news.email.tsx
+++ b/app/routes/_gcn.news.email.tsx
@@ -26,14 +26,13 @@ import { getFormDataString } from '~/lib/utils'
 
 export async function loader({ request }: LoaderFunctionArgs) {
   const user = await getUser(request)
-  if (!user || !user.groups.includes(moderatorGroup))
+  if (!user?.groups.includes(moderatorGroup))
     throw new Response(null, { status: 403 })
   return null
 }
 
 export async function action({ request }: ActionFunctionArgs) {
   const user = await getUser(request)
-  if (!user) throw new Response(null, { status: 403 })
   const data = await request.formData()
   const subject = getFormDataString(data, 'subject')
   const body = getFormDataString(data, 'body')
--- a/app/routes/_gcn.news.email.tsx
+++ b/app/routes/_gcn.news.email.tsx
@@ -41,7 +41,7 @@ export async function action({ request }: ActionFunctionArgs) {
   if (!subject || !body)
     throw new Response('subject and body are required', { status: 400 })
   await sendNewsAnnouncementEmail(subject, body, user)
-  return 'OK' //redirect('/news')
+  return null
 }
 
 export default function () {
--- a/app/routes/_gcn.news.email.tsx
+++ b/app/routes/_gcn.news.email.tsx
@@ -6,11 +6,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import type { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node'
-import { Link, useFetcher } from '@remix-run/react'
+import { Form, Link, useActionData } from '@remix-run/react'
 import {
   Button,
   ButtonGroup,
-  Icon,
+  FormGroup,
   InputGroup,
   InputPrefix,
   TextInput,
@@ -22,7 +22,6 @@ import { useState } from 'react'
 import { getUser } from './_gcn._auth/user.server'
 import { moderatorGroup } from './_gcn.circulars/circulars.server'
 import { sendNewsAnnouncementEmail } from './_gcn.user.email/email_notices.server'
-import Spinner from '~/components/Spinner'
 import { getFormDataString } from '~/lib/utils'
 
 export async function loader({ request }: LoaderFunctionArgs) {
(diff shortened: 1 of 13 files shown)


app/routes/_gcn.user.email/email_notices.server.ts
```diff
@@ -216,3 +218,25 @@ export async function sendTestEmail(to: string) {
...
+) {
+  if (!user || !user.groups.includes(moderatorGroup))
+    throw new Response(null, { status: 403 })
+
+  const db = await tables()
+  const to = (
+    await db.announcement_subscriptions.scan({
+      ProjectionExpression: 'email',
+    })
+  ).Items.map((item) => item.email)
```
lpsinger, line 235: You'll definitely have to paginate this. We have enough users that they won't all be returned in a single scan call. See, for example, https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/table-streams/circulars/index.ts#L57-L65.

lpsinger (changes requested):
Remember to configure the action for the unsubscribe link in https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/_gcn.unsubscribe.%24jwt/actions.server.ts#L56-L95.

dakota002:
@lpsinger where do you think it makes sense to put links to these pages? I don't think there is a direct link to /news, and I don't have one in there yet for the form

app/routes/_gcn.user.email/email_notices.server.ts
```diff
@@ -216,3 +218,25 @@ export async function sendTestEmail(to: string) {
...
+  await sendEmail({
```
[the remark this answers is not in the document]
dakota002: I'm getting this error when I test this, 
```
TypeError: Cannot read properties of undefined (reading 'template')
    at sendEmailBulk (C:\Repos\gcn.nasa.gov\app\lib\email.server.ts:101:39)
```
From the email server:
image

Are you familiar with how this is defined?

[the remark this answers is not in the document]
lpsinger: Yes, it's defined here:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/src/plugins/emailOutgoing.js#L50-L57

This won't work in sandbox mode.

app/routes/_gcn.news.email.tsx
```diff
@@ -0,0 +1,114 @@
...
+export async function action({ request }: ActionFunctionArgs) {
+  const user = await getUser(request)
+  if (!user) throw new Response(null, { status: 403 })
+  const data = await request.formData()
+  const subject = getFormDataString(data, 'subject')
+  const body = getFormDataString(data, 'body')
+  if (!subject || !body)
+    throw new Response('subject and body are required', { status: 400 })
+  await sendNewsAnnouncementEmail(subject, body, user)
+  return 'OK' //redirect('/news')
```
lpsinger, line 44: How about:

```suggestion
  return null
```

lpsinger (changes requested):

lpsinger (changes requested):
A better UX for after successful submission may be to show a message saying that the announcement was sent, similar to what we do in the help desk contact form:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/_gcn.contact.tsx#L98-L117

Sorry for flip-flopping on the pending UI.

dakota002:
No worries, I can change that

dakota002:
Copied a lot of the formatting from the contact page:

image

dakota002:
@lpsinger @jracusin Is this worth including in this PR, a "News" header link:
image

dakota002:
Also a button to link to the announcement page:

image

Is it possible to use a loader or get the user's moderator status in markdown? If not, this button is viewable to anyone, but will take them to an error page if they click it while not signed in

lpsinger:
> Also a button to link to the announcement page:
> 
> image
> 
> Is it possible to use a loader or get the user's moderator status in markdown? If not, this button is viewable to anyone, but will take them to an error page if they click it while not signed in

That's a problem. Yes, you can have a loader in an MDX file.

lpsinger (changes requested):

dakota002:
> > Is it possible to use a loader or get the user's moderator status in markdown? If not, this button is viewable to anyone, but will take them to an error page if they click it while not signed in
> 
> That's a problem. Yes, you can have a loader in an MDX file.

Does it support useLoaderData as well? That would make things a lot easier. How do I implement that?

lpsinger:
Yup. I don't see why not.

dakota002:
Ah okay mdx doesn't really play well with the react hooks, but I was able to break out the button as a component and just make a new hook to use the moderator status, and it works well.

app/routes/_gcn/header/Header.tsx
```diff
@@ -135,6 +135,14 @@ export function Header() {
...
                 Documentation
               </NavLink>,
+              <NavLink
+                className="usa-nav__link"
+                to="/news"
+                key="/news"
+                onClick={hideMobileNav}
+              >
+                News
+              </NavLink>,
```
lpsinger, line 145: I think that I omitted News from the nav bar because it was too wide. What does this look like on desktop?

lpsinger (changes requested):

lpsinger (approved):

PR #1871, opened 2024-01-29
Port getEnvOrDie{,inProduction} from TypeScript to Python

lpsinger:
These are Python ports of utility functions in this TypeScript file: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/lib/env.server.ts

### DIFF
--- a/python/env.py
+++ b/python/env.py
@@ -3,7 +3,9 @@
 # All Rights Reserved.
 
 
+import logging
 import os
+import typing
 
 
 def get_features() -> set[str]:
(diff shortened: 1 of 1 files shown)


swyatt7:
The code looks good to me. Where will this will be implemented?

swyatt7 (approved):

lpsinger:
Unfortunately I don't think that `NODE_ENV` is defined for the Python Lambdas, so I need to change this to use `ARC_ENV`.

lpsinger:
> Where will this will be implemented?

I don't understand the question.

python/env.py
jak574, line 1: Where do you expect these new functions to be called from? Right now this pull won't result in code that gets run.

lpsinger: Here, for one: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/python/across_api/base/tle.py#L196-L197

jak574: So the intention is to replace `os.environ.get` with these calls?

lpsinger: Yes.

swyatt7 (approved):

swyatt7:
Looks good

Issue #1862, opened 2024-01-25
Add support for Bearer Tokens in PoC ACROSS API for BurstCube


Kirill-Vorobyev:
## Description
We need to support OIDC bearer access token auth for the ACROSS API as a first step to develop a Proof of Concept (PoC) that delivers functionality for the BurstCube team.

## Motivation
Protect ACROSS API and allow programmatic access to the API on behalf of a client user.
Save time with re-use of gcn.nasa.gov client credential flow to generate `client_id` and `client_secret`.

## Stakeholders
@lpsinger @swyatt7 @jak574 

## Functional Requirements
TBD

## Acceptance Criteria

1. A user can submit an API request with a header that contains `Authorization: Bearer <token>` where `<token>` is the access token that is returned by the Identity Provider (COGNITO)

Kirill-Vorobyev:
Duplicate of #1833 which may have more context

Issue #1833, opened 2024-01-16
Add OIDC bearer token authentication support to /labs/api


lpsinger:
Add support for OAuth2/OIDC bearer token authentication to /labs/api. The API is the "resource server" in the terminology of [RFC 6750 § 3.1](https://datatracker.ietf.org/doc/html/rfc6750#section-1.3).

- [x] Research whether there are OIDC implementations that are designed for use or that integrate well with FastAPI.
- [x] Devise a design pattern for restricting access to FastAPI endpoints based on the OIDC scope(s) of the bearer token - likely a decorator
- [ ] Choose naming conventions for OIDC scopes. [SciTokens](https://scitokens.org) might be relevant for scope naming.
- [x] Add an example "Hello world"-like endpoint with access limited to an example scope.

# Resources

- [RFC 6750 § 2.1](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) - specification for HTTP request headers
- [Authlib](https://docs.authlib.org/en/latest/) - an easy to use, well documented, and well reputed OAuth2/OIDC implementation in Python

swyatt7:
I haven't implemented anything in the /labs/api/ yet, but I've designed a very small working example and pushed it to my local environment. Should be cloneable and testable in < 3 minutes.

I have it pushed to my local GH repository: https://github.com/swyatt7/fastapi_security_example.

I'd like to have a design discussion on this implementation before diving into the /labs/api/ code.

lpsinger:
You have the right idea here! The main differences for the ACROSS application are:

- Cognito will generate the access tokens, not the API.
- We'll [auto-discover the IdP's configuration and signing public keys](https://openid.net/specs/openid-connect-discovery-1_0.html) from Cognito rather than storing the key in the code.
- _Most_ of the time, API requests will be authenticated using the `Authorization: Bearer` header, rather than the query string.

jak574:
https://github.com/nasa-gcn/gcn.nasa.gov/pull/1861 addresses bullet 1 of this.

PR #1874, opened 2024-01-29
Adding security scope validation to api endpoints

swyatt7:
Addresses: https://github.com/nasa-gcn/gcn.nasa.gov/issues/1833 and https://github.com/nasa-gcn/gcn.nasa.gov/issues/1862

Piggy backed off of @jak574 initial Auth implementation.

Created a second layer of dependency that implements specific route scopes that are verified in the `access_token`. This second layer is implemented by decorating functions with the FastAPI Security function, allowing for a list of scopes to be further passed into a method that has access to the access token.

Functionality is testable on the `/labs/api/v1/secure_hello` endpoint that accepts an access token that has been created the gcn client with scope `gcn.nasa.gov/kafka-public-consumer`.


python/across_api/auth/api.py
```diff
@@ -56,17 +57,20 @@ async def __call__(self, request: Request) -> None:
...
-                if signing_key["kid"] == header["kid"]:
-                    break
-            else:
+
+            signing_key = None
+            for key in jwks_data["keys"]:
+                if key["kid"] == header["kid"]:
+                    signing_key = key
+
+            if signing_key is None:
```
jak574, line 66: Any reason to replace the `for/else` version of this?

python/across_api/auth/api.py
```diff
@@ -33,10 +33,11 @@ def __init__(self, **kwargs):
 
         super().__init__(**kwargs)
 
-    async def __call__(self, request: Request) -> None:
+    async def __call__(self, request: Request):
         credentials: Optional[HTTPAuthorizationCredentials] = await super().__call__(
             request
         )
```
jak574, line 39: add `-> Optional[dict]` type decoration to `__call__`.

python/across_api/auth/api.py
```diff
@@ -56,17 +57,20 @@ async def __call__(self, request: Request) -> None:
...
+
+            if signing_key is None:
                 raise HTTPException(
                     status_code=401, detail="Authentication error: Invalid key."
                 )
 
             # Validate the credentials
             try:
-                jwt.decode(
+                access_token = jwt.decode(
```
lpsinger, line 73: How about just:

```python
try:
    return jwt.decode(...)
except ...:
    ...
```

python/across_api/auth/api.py
```diff
@@ -75,18 +79,47 @@ async def __call__(self, request: Request) -> None:
...
+    # assuming the jwt scopes will be comma separated
+    token_scopes = scopes.split(",")
+
+    # check if user scopes is in endpoint scope
+    if len(security_scopes.scopes):
+        for role in token_scopes:
+            if role in security_scopes.scopes:
+                valid = True
+    else:  # endpoint has no scopes
+        valid = True
```
lpsinger, line 111:
```suggestion
    valid = all(scope in security_scopes.scopes for scope in token_scopes)
```

I take it that `token_scopes` are the scopes that the decorator states that the token must have, and `security_scopes.scopes` are the scopes that are actually present in the token?

lpsinger (changes requested):

python/across_api/auth/api.py
swyatt7: Honestly, I'd never used a `for/if-break/else` loop before, and I was experiencing a bug in testing. I changed it to this just in case.

For me it was more readable, and generally I find that instantiating a variable that is used later in the code as the for-loop key is hard to follow. 

I'll change it back to the for-else with this variation:
```python
signing_key = None
for key in jwks_data["keys"]:
    if key["kid"] == header["kid"]:
        signing_key=key
        break
else:
```

python/across_api/auth/api.py
swyatt7: I changed it to `Optional[dict]` as the type decoration, and mypy is still throwing a fit about it. I believe it is from the  inheritance of `HTTPBearer` and how its `__call__` might have a return type that we aren't meeting in the decorator. I just threw a ` # type: ignore` on it. This should be sufficient since a return type shouldn't be expected how we are implementing our nested dependencies for this case.

python/across_api/auth/api.py
```diff
@@ -56,17 +57,21 @@ async def __call__(self, request: Request) -> None:
...
-            for signing_key in jwks_data["keys"]:
-                if signing_key["kid"] == header["kid"]:
-                    break
+
+            signing_key = None
+            for key in jwks_data["keys"]:
+                if "kid" in header.keys():
+                    if key["kid"] == header["kid"]:
+                        signing_key = key
+                        break
```
lpsinger, line 66: Still an unnecessary transformation of this loop.

python/across_api/auth/api.py
```diff
@@ -78,15 +83,41 @@ async def __call__(self, request: Request) -> None:
...
+    scopes = access_token.get("scope", "")
+
+    # assuming the jwt scopes will be comma separated
+    token_scopes = scopes.split(",")
+
+    # check if user scopes is in endpoint scope
+    if len(security_scopes.scopes):
+        valid = any(scope in security_scopes.scopes for scope in token_scopes)
+    else:  # endpoint has no scopes
+        valid = True
```
lpsinger, line 109:
```suggestion
    valid = all(scope in token_scopes for scope in security_scopes.scopes)
```

python/across_api/auth/api.py
```diff
@@ -78,15 +83,41 @@ async def __call__(self, request: Request) -> None:
...
-security = JWTBearer(
+JWTBearerSecurity = JWTBearer(
     scheme_name="ACROSS API Authorization",
     description="Enter your access token.",
 )
-JWTBearerDep = [Depends(security)]
 
 
-@app.get("/auth/verify", dependencies=JWTBearerDep)
+async def ScopeAuthenticate(
```
lpsinger, line 95: This is an authorization decision. The user is already authenticated.

```suggestion
async def ScopeAuthorize(
```

lpsinger (changes requested):

python/across_api/auth/api.py
lpsinger: Oops, sorry for repeating myself. I am looking at your comment https://github.com/nasa-gcn/gcn.nasa.gov/pull/1874#discussion_r1478502298 again.

lpsinger: Oh, right. I think that this version addresses your comment.

python/across_api/auth/api.py
lpsinger: I agree with your example, but I think that the correct implementation is https://github.com/nasa-gcn/gcn.nasa.gov/pull/1874#discussion_r1479072676.

python/across_api/auth/api.py
jak574: I think this is fair enough, the for/else version was compact, but is more esoteric, and therefore less readable.

python/across_api/auth/api.py
jak574: Can you change it to `# type: ignore[override]` so we're telling `mypy` only to ignore this one specific issue, rather than ignoring all issues that might crop up.

python/across_api/auth/api.py
```diff
@@ -78,15 +83,41 @@ async def __call__(self, request: Request) -> None:
         else:
             raise AssertionError("No credentials passed.")
 
+        return {}
+
```
lpsinger, line 87: Remove unreachable code:

```suggestion
```

lpsinger (changes requested):

python/across_api/auth/api.py
```diff
@@ -56,17 +57,21 @@ async def __call__(self, request: Request) -> None:
...
-                if signing_key["kid"] == header["kid"]:
-                    break
+
+            signing_key = None
+            for key in jwks_data["keys"]:
+                if "kid" in header.keys():
+                    if key["kid"] == header["kid"]:
+                        signing_key = key
+                        break
             else:
```
lpsinger, line 67: It's a valid criticism that the `for: else` construct esoteric, but this version is now redundant. Try this instead:

```suggestion
            if signing_key is None:
```

lpsinger (changes requested):

python/across_api/auth/api.py
jak574: OK, sorry I just stared at this code for 5 more minutes. The root cause here is that you're forcing this function to return a dictionary of the access token, whereas it should just pass through the value of `credentials`. If you change the return to Optional[HTTPAuthorizationCredentials] and just return `credentials` at the end, and make no other returns, it works fine. 

Then you can change `ScopeAuthenticate` to start like this:
```
async def ScopeAuthenticate(
    security_scopes: SecurityScopes,
    credentials: Annotated[HTTPAuthorizationCredentials, Depends(JWTBearerSecurity)],
):
    valid = False
    access_token = jwt.get_unverified_claims(credentials.credentials)
    scopes = access_token.get("scope", "")
    ...
```

lpsinger: I'm fairly nervous about calling `jwt.get_unverified_claims` in a function that is separate from the one that checks the signature. It makes it less obvious that we have ever checked the signature.

Could you instead have `JWTBearerSecurity` depend upon `HTTPBearer` rather than inheriting from it?

jak574: I can understand the appearance in insecurity, however `credentials` doesn't get passed to `ScopeAuthenticate` unless its signature verification is passed. If you treat `get_unverified_claim` for what it is, a way of extracting a key from a JWT without doing another signature check, it's fine. If we're going to rework the whole code to allay his fear, I'd just favor leaving as is.

lpsinger: I actually don't know whether this is secure or not. We used `jwt.decode` in the other context, not `get_unverified_claims`. I'd have to consult the implementation of `get_unverified_claims` to determine if it only harvests claims from the signed section of the JWT (though it probably does). However this code _is_ insecure in isolation, and it creates the risk that we might carelessly copy-paste this somewhere else without understanding the consequences.

But this code has _two_ problems: (1) it appears insecure, though it might actually be secure and (2) it doesn't pass the type checking.

I think that changing the relationship between JWTBearerSecurity from inheritance to dependency would be a minor change, and it would be manifestly correct.

lpsinger: There is actually a third problem that it decodes the JWT twice, which is inefficient.

jak574: My feeling here is that I fundamentally handled this wrong in my original pull request by having the JWT decode inside `JWTBearer`. `HTTPBearer` is supposed to pass through the Bearer JWT when `__call__` is executed. This is the root cause of the mypy issue, changing this behavior so we can pass decoded JWT to `ScopeAuthenticate`.

So this would suggest that the JWT decoding/signature checking should occur elsewhere, because yes, it doesn't make sense to have to decode a JWT twice. So it should happen in `ScopeAuthenticate`, which becomes a dependency for all authenticated endpoints. So the same code checking the scope is also checking the bearer token signature.

lpsinger: See #1906.

jak574: Yeah, like that.

jak574: Yep, I reviewed the code last night but wanted to have a chance to run it, which I'm doing now...

python/across_api/auth/api.py
swyatt7: This will work if our endpoints only have one scope. The `any()` will work if the endpoint's `security_scopes.scopes` contains more than one.

I'm fine with changing it, but if we envision our endpoints having more than one scope, the `all()` will return false.

lpsinger: Huh? This is supposed to check that all of the required scopes are present in the access token. Why would that fail if the endpoints require more than one scope?

swyatt7: Our endpoints can be decorated with a list of scopes. For example `/read_foo` api endpoint can be decorated with `scopes = ["read:/", "read:/foo"]`. There might be an instance where a user only has a JWT with scope privileges that only contain ["read:/foo"].

This leads to:
```python
security_scopes.scopes =  ["read:/", "read:/foo"] #api endpoint scopes
token_scopes = ["read:/foo"]

valid = all(scope in token_scopes for scope in security_scopes.scopes) #returns False
``` 

`valid` returns `False` with this implementation, event though the user scopes are within the endpoint scopes.

whereas if we implement this validation with ` any(scope in token_scopes for scope in security_scopes.scopes)` it will return `valid = True`, which is the desired outcome. The only issue is that if `security_scopes.scopes` is empty, then we will need to check that since the `any(...)` evaluation will return `False`.

lpsinger: I see. However, the `all(...)` expression I gave is consistent with the behavior of the FastAPI example code in https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/#global-view.

lpsinger: Is there some documentation about SciTokens that says that `read:/` is supposed to imply `read:/foo` as well? If this is supposed to model POSIX permissions, they don't work that way.

Also, if we intend a token to have both `read:/` and `read:/foo` permissions, couldn't we just issue a token with both scopes?

python/across_api/auth/api.py
```diff
@@ -79,14 +84,38 @@ async def __call__(self, request: Request) -> None:
...
     description="Enter your access token.",
 )
-JWTBearerDep = [Depends(security)]
 
 
-@app.get("/auth/verify", dependencies=JWTBearerDep)
+async def ScopeAuthorize(
+    security_scopes: SecurityScopes,
+    access_token: Annotated[dict, Depends(JWTBearerSecurity)],
+):
```
lpsinger, line 96: Why are all of these functions CamelCase rather than snake_case? That's usually reserved for classes.

python/across_api/auth/api.py
swyatt7: I see what you mean in their tutorial. 

Here is the documentation on [SciToken Claims](https://scitokens.org/technical_docs/Claims). Go down to the section on "SciToken Claim Semantics". 

I assume the `read:/` is an administrator role to _read everything_ where `read:/foo" is a restricted read, where the user only has the ability to _read what is in foo_. I 

Your vision is that when a user is assigned a scope that includes `read:/foo` that it will automatically give them the scopes of `["read:/", "read:/foo"]`? Which would work with the `all(...)` implementation

lpsinger: > Your vision is that when a user is assigned a scope that includes `read:/foo` that it will automatically give them the scopes of `["read:/", "read:/foo"]`? Which would work with the `all(...)` implementation

No, following the example in the "SciTokens Scopes" section, it should be the opposite: an assigned scope of `read:/` should be equivalent to `["read:/", "read:/foo"]`.

Anyway, I don't want to get into path parsing in this PR, though we could in a future one.

swyatt7: Ok. I've implemented the `all(...)`, I still had to check if the endpoint scopes had a length because the line of code would return `False` if the `security_scopes.scopes` list was empty.

swyatt7:
rebased with the changes above and re-implemented the `scopeauthorize()` dependency on the endpoints with the new codebase.

python/across_api/auth/api.py
swyatt7: I agree, this has been fixed to follow the snake_case for variable names.

lpsinger (approved):

PR #1876, opened 2024-01-30
Ignore python/vendor directory

lpsinger:
This permits test discovery to work while running a sandbox server.


jak574 (approved):

lpsinger:
@jak574, you mentioned that you couldn't get pytest working. Was this the issue that you were having?

PR #1878, opened 2024-01-30
ACROSS API: Add unit tests for Ephemeris calculations and others

jak574:
This pull adds unit tests for `BurstCubeEphem` and `SwiftEphem` classes. There are some modifications included to the TLE code to allow passing a TLE directly, so that the test is passed a fixed TLE, and a database or space-track.org query is not required to run the test. 

This change also necessitated having to add the ability to hash the `TLEEntry` model, so I just implemented hashing for all Pydantic models globally. This allows for classes that get Pydantic models passed as arguments to be cacheable.

The tests added check that the Ephemeris generation validates against pre-calculated values determined using a different method utilizing the  [skyfield](https://rhodesmill.org/skyfield/]) python module. 

The tests don't introduce a `skyfield` dependency, but the code to generate the tests is included (commented out) in the tests, only the outputs of this code are used. This way we protect ourselves from both dependency creep and any changes in `skyfield` that might break our tests.

For Swift I also validated the values against on-hand values generated by the Swift Flight Operations Team using STK, which agree with the `skyfield` values.

This code also introduces unit tests for `/across/resolve` endpoint.


swyatt7 (changes requested):
All of your hardcoded data needs to be put into a `conftest.py` file as a [pytest fixture](https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files).

Then the defined fixtures can be injected into the tests as dependencies.

python/across_api/base/schema.py
```diff
@@ -39,6 +39,9 @@ class BaseSchema(BaseModel):
 
     model_config = ConfigDict(from_attributes=True, arbitrary_types_allowed=True)
 
+    def __hash__(self):
+        return hash((type(self),) + tuple(self.__dict__.values()))
```
lpsinger, line 43: Why does this need a hash function now?

swyatt7 (approved):
I pushed the `pytest.fixture` changes to the branch directly.

python/across_api/base/schema.py
jak574: Quoting above "This change also necessitated having to add the ability to hash the TLEEntry model, so I just implemented hashing for all Pydantic models globally. This allows for classes that get Pydantic models passed as arguments to be cacheable."

So I want to be able to pass a `TLEEntry` as an argument to `BurstCubeEphem`, but that breaks the caching because cashing works by creating a hash based on the arguments, which means all arguments have to be hashable. This will likely come up again at some point where we want to pass a Pydantic model based object as an argument, so it seemed best to globally solve the issue.

lpsinger:
@jak574, would you please start rebasing your branches? That was 51 commits onto main; it makes it harder to do troubleshooting should we ever need to resort to `git bisect`.

PR #1905, opened 2024-02-06
Run flake8 in CI workflow and pre-commit hook

lpsinger:
This is a Python static analysis tool that can check simple mistakes. We use it on all of our other Python projects.


jak574 (approved):
In my own workflow I've moved to using ruff to replace both flake8 and black. It's compatible but way faster.

lpsinger:
Alright, it looks like that's a replacement for both `black` and `flake8`.

lpsinger:
See #1907.

PR #1906, opened 2024-02-06
Decode access token in a type-safe, low boilerplate way

lpsinger:
- Instead of inheriting from HTTPBearer, use it as a dependency.
- This allows us to return something other than the credential, namely: the verified and decoded claims.
- Factor out the issuer and JWKS discovery as dependency functions that have no other dependencies so that these can be shared across multiple requests.

### DIFF
--- a/python/across_api/auth/api.py
+++ b/python/across_api/auth/api.py
@@ -5,7 +5,6 @@
 
 import os
 from typing import Annotated, Any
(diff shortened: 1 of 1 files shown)


jak574 (approved):

jak574:
Looks good. I removed an unused import (actually meant to "suggest the change" but apparently I just committed it).

PR #1908, opened 2024-02-08
ACROSS API: Add SAA and EarthLimb constraint classes

jak574:
# Constraint Classes

This PR adds the following constraint classes. 

## `SAAPolygonConstraint`  

Calculate for a given spacecraft SAA polygon, if the spacecraft is in the SAA at a given time. The constraint class is instantiated with the polygon as an argument, and the object uses `__call__` to calculate a bool or array of bools indicating if the spacecraft is inside (`True`) or outside (`False`) the SAA. Arguments for are `time`, an astropy`Time` object (scalar or array-like) and a pre-calculated ephemeris object (e.g. `BurstCubeEphem`). `time` must be a subset of the times for which the ephemeris was calculated.

Constraint classes for BurstCube and Swift are instantiated with the correct SAA polygons for those missions.

## `EarthLimbConstraint` 

Calculates if a target is inside an Earth Limb constraint. When instantiating the class, `min_angle` determines how close to the Earth limb is considered a constraint, setting this to zero means you are calculating for a pure Earth occultation. Arguments for the object are the same as for `SAAPolygonConstraint` except it adds a required `skycoord` parameter which calculates the Earth limb constraint for a given `SkyCoord` object, this can be scalar or array-like.

## Tests

The PR also adds `pytest` tests for each constraint class, validating against pre-calculated values using third party methods, including using `skyfield` derived numbers and the [Swift TOO API](https://github.com/Swift-SOT/swifttools).

### DIFF
--- a/python/across_api/base/constraints.py
+++ b/python/across_api/base/constraints.py
@@ -14,7 +14,7 @@ from shapely import Polygon, points  # type: ignore[import]
 from .ephem import EphemBase
 
 
-def getslice(time: Time, ephem: EphemBase) -> slice:
+def get_slice(time: Time, ephem: EphemBase) -> slice:
     """
     Return a slice for what the part of the ephemeris that we're using.
 
@@ -101,13 +101,13 @@ class SAAPolygonConstraint(Constraint):
         """
 
         # Find a slice what the part of the ephemeris that we're using
-        i = getslice(time, ephem)
+        i = get_slice(time, ephem)
 
-        inconstraint = self.polygon.contains(
+        in_constraint = self.polygon.contains(
             points(ephem.longitude[i], ephem.latitude[i])
         )
         # Return the result as True or False, or an array of True/False
-        return inconstraint[0] if time.isscalar else inconstraint
+        return in_constraint[0] if time.isscalar else in_constraint
 
 
 class EarthLimbConstraint(Constraint):
@@ -166,16 +166,18 @@ class EarthLimbConstraint(Constraint):
         assert skycoord is not None, "SkyCoord object must be passed"
 
         # Find a slice what the part of the ephemeris that we're using
-        i = getslice(time, ephem)
+        i = get_slice(time, ephem)
 
         # Calculate the angular distance between the center of the Earth and
         # the object. Note that creating the SkyCoord here from ra/dec stored
         # in the ephemeris `earth` is 3x faster than just doing the separation
         # directly with `earth`.
-        inconstraint = (
+        in_constraint = (
             SkyCoord(ephem.earth[i].ra, ephem.earth[i].dec).separation(skycoord)
             < ephem.earthsize[i] + self.min_angle
         )
 
         # Return the result as True or False, or an array of True/False
-        return inconstraint[0] if time.isscalar and skycoord.isscalar else inconstraint
+        return (
+            in_constraint[0] if time.isscalar and skycoord.isscalar else in_constraint
+        )
--- a/python/across_api/base/constraints.py
+++ b/python/across_api/base/constraints.py
@@ -2,8 +2,7 @@
 # Administrator of the National Aeronautics and Space Administration.
 # All Rights Reserved.
 
-from abc import ABC, abstractmethod
-from typing import Optional, Union
+from typing import Union
 
 import astropy.units as u  # type: ignore[import]
 import numpy as np
@@ -50,17 +49,7 @@ def get_slice(time: Time, ephem: EphemBase) -> slice:
         return slice(ephem.ephindex(time[0]), ephem.ephindex(time[-1]) + 1)
 
 
-class Constraint(ABC):
-    """Define the basic structure of a Constraint class."""
-
-    @abstractmethod
-    def __call__(
-        self, time: Time, ephem: EphemBase, skycoord: Optional[SkyCoord] = None
-    ) -> Union[bool, np.ndarray]:
-        pass
-
-
-class SAAPolygonConstraint(Constraint):
+class SAAPolygonConstraint:
     """
     Polygon based SAA constraint. The SAA is defined by a Shapely Polygon, and
     this constraint will calculate for a given set of times and a given
@@ -77,9 +66,7 @@ class SAAPolygonConstraint(Constraint):
     def __init__(self, polygon: list):
         self.polygon = Polygon(polygon)
 
-    def __call__(
-        self, time: Time, ephem: EphemBase, skycoord: Optional[SkyCoord] = None
-    ) -> Union[bool, np.ndarray]:
+    def __call__(self, time: Time, ephem: EphemBase) -> Union[bool, np.ndarray]:
         """
         Return a bool array indicating whether the spacecraft is in constraint
         for a given ephemeris.
@@ -110,7 +97,7 @@ class SAAPolygonConstraint(Constraint):
         return in_constraint[0] if time.isscalar else in_constraint
 
 
-class EarthLimbConstraint(Constraint):
+class EarthLimbConstraint:
     """
     For a given Earth limb avoidance angle, is a given coordinate inside this
     constraint?
@@ -136,7 +123,7 @@ class EarthLimbConstraint(Constraint):
         self,
         time: Time,
         ephem: EphemBase,
-        skycoord: Optional[SkyCoord] = None,
+        skycoord: SkyCoord,
     ) -> Union[bool, np.ndarray]:
         """
         Check for a given time, ephemeris and coordinate if positions given are
@@ -162,9 +149,6 @@ class EarthLimbConstraint(Constraint):
             otherwise.
 
         """
-        # This class requires a SkyCoord object to be passed
-        assert skycoord is not None, "SkyCoord object must be passed"
-
         # Find a slice what the part of the ephemeris that we're using
         i = get_slice(time, ephem)
 
--- a/python/across_api/base/constraints.py
+++ b/python/across_api/base/constraints.py
@@ -1,0 +1,181 @@
+# Copyright © 2023 United States Government as represented by the
+# Administrator of the National Aeronautics and Space Administration.
+# All Rights Reserved.
+
+from abc import ABC, abstractmethod
+from typing import Optional, Union
+
+import astropy.units as u  # type: ignore[import]
+import numpy as np
+from astropy.coordinates import Angle, SkyCoord  # type: ignore[import]
+from astropy.time import Time  # type: ignore[import]
+from shapely import Polygon, points  # type: ignore[import]
+
+from .ephem import EphemBase
+
+
+def getslice(time: Time, ephem: EphemBase) -> slice:
+    """
+    Return a slice for what the part of the ephemeris that we're using.
+
+    Arguments
+    ---------
+    time
+        The time to calculate the slice for
+    ephem
+        The spacecraft ephemeris
+
+    Returns
+    -------
+        The slice for the ephemeris
+    """
+    # If we're just passing a single time, we can just find the index for that
+    if time.isscalar:
+        # Check that the time is within the ephemeris range, this should never
+        # happen as the ephemeris is generated based on this range.
+        assert (
+            time >= ephem.begin and time <= ephem.end
+        ), "Time outside of ephemeris of range"
+        # Find the index for the time and return a slice for that index
+        index = ephem.ephindex(time)
+        return slice(index, index + 1)
+    else:
+        # Check that the time range is within the ephemeris range, as above.
+        assert (
+            time[0] >= ephem.begin and time[-1] <= ephem.end
+        ), "Time outside of ephemeris of range"
+
+        # Find the indices for the start and end of the time range and return a
+        # slice for that range
+        return slice(ephem.ephindex(time[0]), ephem.ephindex(time[-1]) + 1)
+
+
+class Constraint(ABC):
+    """Define the basic structure of a Constraint class."""
+
+    @abstractmethod
+    def __call__(
+        self, time: Time, ephem: EphemBase, skycoord: Optional[SkyCoord] = None
+    ) -> Union[bool, np.ndarray]:
+        pass
+
+
+class SAAPolygonConstraint(Constraint):
+    """
+    Polygon based SAA constraint. The SAA is defined by a Shapely Polygon, and
+    this constraint will calculate for a given set of times and a given
+    ephemeris whether the spacecraft is in that SAA polygon.
+
+    Attributes
+    ----------
+    polygon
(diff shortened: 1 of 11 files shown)


lpsinger (changes requested):
Shapely does 2D Euclidean geometry. Is this code correct for a polygon on the sphere?

python/across_api/base/constraints.py
```diff
@@ -0,0 +1,181 @@
...
+    this constraint will calculate for a given set of times and a given
+    ephemeris whether the spacecraft is in that SAA polygon.
+
+    Attributes
+    ----------
+    polygon
+        Shapely Polygon object defining the SAA polygon.
+    """
+
+    polygon: Polygon
```
swyatt7, line 75: My only concern is also that this is shapely.Polygon. Should consider re-implementing with spherical reference calculations. 

Maybe starting with an arbitrarily high order healpix map (order=10) and finding the pixels contained by the SAA, and testing whether there is an intersection between the ephem lon lat pixels (in the same order map).

swyatt7 (changes requested):

python/across_api/base/constraints.py
lpsinger: I would use Astropy regions. https://astropy-regions.readthedocs.io/en/stable/contains.html

jak574: I will look into astropy regions, if I can get rid of the shapely dependency I'd be fine with that. 

Regarding the accuracy, I've compared the predicted SAA windows with those predicted by the Swift Flight Operations Team using Satellite ToolKit (STK). They match within seconds, and we're calculating with a 60s time resolution. 

I also happen to have on hand the algorithm that Swift's Flight Software uses to calculate if it's inside the SAA polygon, and it's extremely basic, more so that shapely even, and is certainly not correcting for spherical geometry. 

It's not important to be mathematically perfect, it's important to match what the spacecraft would do. For Swift shapely does. For BurstCube, we'll I can talk with Ava about how they calculate SAA passages so we can best match theirs.

lpsinger: It turns out that astropy-regions does _not_ use Shapely to implement geometric predicates, but its implementation of predicates for sky regions _does_ rely on projecting into a WCS and then doing Euclidean geometry. So it would have the same general problems as shapely.

lpsinger: I would instead use https://github.com/spacetelescope/spherical_geometry which I have used in the past.

Yes, you'll get the same results in this case because the SAA polygon does not cross the dateline or enclose either pole, but in the likely event that we have to do any _other_ computational geometry on the sphere, we're likely to want to use implementations of spatial predicates that give correct results in general on the sphere.

jak574: Yes, I just came to the same conclusion that regions isn't fit for this purpose. I'll look into that STScI package.

jak574: Here is the result of my investigation. I plugged in the `spherical-geometry` library and it works. For Swift side, the results validate using pytest against the Swift API, but then again so does the `shapely` driven calculation also. 

The big issue is that `spherical-geometry` is slow. For single points it's 17x slower than `shapely` and even worse, it does not support vectorization (https://github.com/spacetelescope/spherical_geometry/issues/259), so you have to feed it the points one by one, slowing things down even more. For 86400 data points, calculating SAA using `spherical_geometry` takes 21.4s, for `shapely`, 384ms (M1 MacBook Pro). 

I compared the results against actual SAA passages from Swift using a high resolution (1s) ephemeris, so these are *real numbers* vs prediction. For both the `shapely` and `spherical_geometry` versions of the code, the difference in the calculated SAA entry/exit times vs Swift actual is <<60s, so when we're using 60s time resolution, it's a wash.   

The `spherical_geometry` code is slightly more accurate, but the difference is on the ~2-3s level.

This all leads to my previous conclusion that `shapely` is *good enough* for this purpose, insofar that it matches the spacecraft values fot SAA when calculated at a 60s resolution.

lpsinger: I agree, that's not a worthwhile tradeoff. Shapely is an excellent pick for now.

In case we need something like this but more general... I skimmed spherical_geometry and mapped out what it would take to make it vectorize. It's clear that the code was written with that in mind at some level (lots of `np.broadcast_arrays`, the Numpy C code has lots of `PyUFunc_FromFuncAndDataAndSignature` for [generalized ufuncs](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html)). It wouldn't be a huge lift, but I would guess that it would be a week or two of solid work.

There is also https://github.com/lsst/sphgeom which looks more complete but it would be _mad difficult_ to vectorize effectively because its Python wrapper sits on top of a C++ class library.

jak574: I will add, that the difference between the calculated SAA using the `shapely` and `spherical-geometry` is more pronounced at low latitudes, so the inaccuracies of the `shapely` method may be more critical for high-inclination orbit missions. 

Attached image shows the difference between the two methods.
Screenshot 2024-02-08 at 20 56 40

lpsinger: You could minimize those errors by working in a projection (i.e., a WCS) that is centered on the SAA.

python/across_api/base/constraints.py
```diff
@@ -0,0 +1,181 @@
...
+import astropy.units as u  # type: ignore[import]
+import numpy as np
+from astropy.coordinates import Angle, SkyCoord  # type: ignore[import]
+from astropy.time import Time  # type: ignore[import]
+from shapely import Polygon, points  # type: ignore[import]
+
+from .ephem import EphemBase
+
+
+def getslice(time: Time, ephem: EphemBase) -> slice:
```
lpsinger, line 17:
```suggestion
def get_slice(time: Time, ephem: EphemBase) -> slice:
```

python/across_api/base/constraints.py
```diff
@@ -0,0 +1,181 @@
...
+        assert skycoord is not None, "SkyCoord object must be passed"
+
+        # Find a slice what the part of the ephemeris that we're using
+        i = getslice(time, ephem)
+
+        # Calculate the angular distance between the center of the Earth and
+        # the object. Note that creating the SkyCoord here from ra/dec stored
+        # in the ephemeris `earth` is 3x faster than just doing the separation
+        # directly with `earth`.
+        inconstraint = (
```
lpsinger, line 175:
```suggestion
        in_constraint = (
```

python/across_api/base/constraints.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Returns
+        -------
+        bool
+            `True` if the coordinate is inside the constraint, `False`
+            otherwise.
+
+        """
+        # This class requires a SkyCoord object to be passed
+        assert skycoord is not None, "SkyCoord object must be passed"
```
lpsinger, line 166: If it must not be None, then why does the type signature permit it to be None?

lpsinger (changes requested):
Please remove all commented out code.

python/across_api/base/constraints.py
jak574: I'm universally defining all constraint classes through an ABC as taking skycoord as an argument. Some (e.g. SAA) can ignore it and safely have it set to `None`, others where like `EarthConstraint` require it. In the ABC you have to set it to something that will work for every class, so it's `Optional[SkyCoord]`. This assert both makes sure that you pass coordinates when it's needed, plus it lets `mypy` know that you're not passing a `None`.

Another way of handling this would just be to remove the ABC, removing the restrictions on Constraint Class structure. This would break the concept behind these classes, but not the BurstCube MVP.

python/across_api/base/constraints.py
jak574: Fixed by ecc8a80.

python/across_api/base/constraints.py
jak574: Fixed by ecc8a80.

python/across_api/base/constraints.py
jak574: WCS is a rabbit hole I don't want to go down right now. I will say that just made a HEALPix based SAA constraint, and it's super fast and accurate, faster than `shapely`. I created the HEALPix map using `spherical_geometry` (took 15 mins for a NSIDE=512 map). The created map bzipped is ~3.4kb and is only read in at instantiation of the constraint. This could be a good alternative if we don't just go with the current `shapely` implementation. Basically a lookup table for the `spherical_geometry` version of this constraint.

jak574:
> Please remove all commented out code.

You mean in the tests? I thought it was nice to show how the values we tested against were actually created. I've seen it done elsewhere.

lpsinger:
> > Please remove all commented out code.
> 
> You mean in the tests? I thought it was nice to show how the values we tested against were actually created. I've seen it done elsewhere.

Yes, it is, but if it's commented out, then there's no way to make sure that the commented-out code continues to work. Why not just add a test dependency on skyfield?

jak574:
> Yes, it is, but if it's commented out, then there's no way to make sure that the commented-out code continues to work. Why not just add a test dependency on skyfield and uncomment that code?

Sure OK.

swyatt7:
I have a mild concern. How does the web-api handle all of the `assert` statements? If it returns a 500 internal service error (which I suspect it might), then all of these should be changed to raises with a valid message to the user. I do see that you do use `raise exceptions` in the `base/ephem` class. I think we should (as a standard) only use `raise exceptions` in web-api type validation.

lpsinger:
> I have a mild concern. How does the web-api handle all of the `assert` statements? If it returns a 500 internal service error (which I suspect it might), then all of these should be changed to raises with a valid message to the user. I do see that you do use `raise exceptions` in the `base/ephem` class. I think we should (as a standard) only use `raise exceptions` in web-api type validation.

One usually use assertions to check for preconditions or postconditions that must be true if the code is functioning correctly. As such a failed assertion _should_ result in an opaque 500 internal service error. However we should _not_ use assertions to check for foreseeable and correctable error conditions.

python/across_api/base/constraints.py
lpsinger: I thought that inherited methods are allowed to narrow types?

swyatt7:
> > I have a mild concern. How does the web-api handle all of the `assert` statements? If it returns a 500 internal service error (which I suspect it might), then all of these should be changed to raises with a valid message to the user. I do see that you do use `raise exceptions` in the `base/ephem` class. I think we should (as a standard) only use `raise exceptions` in web-api type validation.
> 
> One usually use assertions to check for preconditions or postconditions that must be true if the code is functioning correctly. As such a failed assertion _should_ result in an opaque 500 internal service error. However we should _not_ use assertions to check for foreseeable and correctable error conditions.

From a user experience perspective, opaque 500 ISE's are never a good time. And will result in more "why is my code not working emails". I don't think there is any harm in raising an exception letting the user know why their parameters were entered incorrectly from the api.

jak574:
> I have a mild concern. How does the web-api handle all of the `assert` statements? If it returns a 500 internal service error (which I suspect it might), then all of these should be changed to raises with a valid message to the user. I do see that you do use `raise exceptions` in the `base/ephem` class. I think we should (as a standard) only use `raise exceptions` in web-api type validation.

I will never put in an assertion that I expect to ever be triggered except by a coding error. Where I use them, I'm using them because they prove to mypy that something is what it says it is. They should never be able to be triggered by user inputs. If they can be, then we should remove them and replace with `raise HTTPException`.

jak574:
> > I have a mild concern. How does the web-api handle all of the `assert` statements? If it returns a 500 internal service error (which I suspect it might), then all of these should be changed to raises with a valid message to the user. I do see that you do use `raise exceptions` in the `base/ephem` class. I think we should (as a standard) only use `raise exceptions` in web-api type validation.
> 
> One usually use assertions to check for preconditions or postconditions that must be true if the code is functioning correctly. As such a failed assertion _should_ result in an opaque 500 internal service error. However we should _not_ use assertions to check for foreseeable and correctable error conditions.
> 
> Take the following example from this PR:
> 
> ```python
>     if time.isscalar:
>         # Check that the time is within the ephemeris range, this should never
>         # happen as the ephemeris is generated based on this range.
>         assert (
>             time >= ephem.begin and time <= ephem.end
>         ), "Time outside of ephemeris of range"
> ```
> 
> If this condition must hold because it is a precondition and the code that calls it ensures that it is so, then an assertion is appropriate. If on the other hand this is checking that input provided by the user is in the required range, then no, an assertion is not appropriate --- 400 Bad Request would be a better fit.

This is an assertion because it should never be true. The user provides the input date rate, this generates an ephemeris that covers that range, therefore the ephemeris should *always* be valid, unless a bug has been introduced to the code.

python/across_api/base/constraints.py
jak574: `mypy` thinks differently.
```
base/constraints.py:135: error: Signature of "__call__" incompatible with supertype "Constraint"  [override]
base/constraints.py:135: note:      Superclass:
base/constraints.py:135: note:          def __call__(self, time: Any, ephem: EphemBase, skycoord: Any | None = ...) -> bool | ndarray[Any, Any]
base/constraints.py:135: note:      Subclass:
base/constraints.py:135: note:          def __call__(self, time: Any, ephem: EphemBase, skycoord: Any) -> bool | ndarray[Any, Any]
Found 1 error in 1 file (checked 25 source files)
```

jak574:
> Please remove all commented out code.

Fixed in  [0ddbd3f](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1908/commits/0ddbd3f10a5a8cbef4f9df8a598bffe2652fef22).

jak574:
> From a user experience perspective, opaque 500 ISE's are never a good time. And will result in more "why is my code not working emails". I don't think there is any harm in raising an exception letting the user know why their parameters were entered incorrectly from the api.

I agree, 500s should only ever be raised due to the code not functioning properly (i.e. something is fundamentally broken) and never as a result of wrong user input. 

I believe I have tried to raise 400 range exceptions wherever there's an issue with inputs. If you find anywhere where this is not the case, it needs to be fixed.

python/across_api/base/constraints.py
jak574: Anyway, [ed41cbb](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1908/commits/ed41cbb8f645194fb73d2179b73bb5a289d271e1) removes the ABC stuff. Arguments for the two constraints are now appropriate for the calls. This is good enough for the MVP we're building.

swyatt7 (approved):
Alright my concerns are alleviated. We are OK with the shapely implementation. I'll approve.

PR #1909, opened 2024-02-08
Adds SSM document to update Kafka ACLs and methods to trigger it

dakota002:


lpsinger (changes requested):
I don't think that we modify ACLs by logging into the brokers and executing commands. Here's why:

- We have the ability to log in to them now because they are EC2 instances, but after we migrate to Kubernetes we probably won't. We definitely won't after the migration to Confluent Cloud.
- Passing ACL names on the command line opens up a host of input sanitization issues.

Instead, we should use either the Kafka admin API or the Confluent REST API.

PR #1910, opened 2024-02-09
adjusting synonyms table and index

Courey:


app.arc
```diff
@@ -83,8 +83,8 @@ circulars
   PointInTimeRecovery true
 
 synonyms
-  eventId *String
-  synonymId **String
+  synonymId *String
+  eventId **String
```
lpsinger, line 87: Could we use the field names `term` and `uuid` instead? The `synonymId` is just the word itself.

lpsinger (changes requested):

app.arc
Courey: > The synonymId is just the word itself.
what word?

I'm fine with changing the names, but I want to make sure we have the same concept in mind. The eventId would be the term? is this because each id isn't necessarily an event? That is how we link to the circular.

The way I was looking at it is that the synonymId (or uuid) doesn't really matter aside from giving a unique grouping identifier. I can delete all synonyms in a synonym group by that uuid, I can find a synonym group by it, and I can get sibling synonyms using it.

lpsinger: What I have in mind in #745 is that the partition key is `term` which is the name of the event, like "GRB 170817". The uuid field is used to match terms together; terms that have the same UUID are synonyms. Note that the uuid field is _not_ a sort key, it is just an ordinary field. So I define the table like this:

```
synonyms
  term *String
```

and the secondary index like this:

```
synonyms
  uuid *String
  term **String
  name synonymsByUuid
```

Courey: I would strongly advise against the use of "term". In a relational database we would use a foreign key here which would enforce the connection between the two fields. Here we are pretending this is relational data. The data that links the two tables together is the eventId on the circular and the eventId (or what you want to call term). If someone looked at the two tables and didn't know the history, they would have absolutely no idea that term on a synonym was the same thing as an eventId on a circular. We try to avoid that in databases, especially in cases like this where you can't have some kind of foreign key constraint to make it explicit that the two are linked.

I am happy to change synonymId to uuid and use the secondary index like you suggested.

lpsinger: Oh, I see. Yes, we should call it eventId to be consistent with the circulars table. But that should be the partition key.

lpsinger (commented):
Won't this change break the existing code on main? Can we really safely change the table schema without updating the code?

Courey:
> Won't this change break the existing code on main? Can we really safely change the table schema without updating the code?
I don't have access to the database, would you verify that the synonyms table is empty?

lpsinger:
> > Won't this change break the existing code on main? Can we really safely change the table schema without updating the code?
> 
> I don't have access to the database, would you verify that the synonyms table is empty?

It is empty, but what I was concerned about was whether this would break existing code on main. But the primary area of concern is https://github.com/nasa-gcn/gcn.nasa.gov/blob/83aaf8dd0b727333d774b48d0794d9c36d98b3c0/app/table-streams/synonyms/index.ts, which isn't impacted by this change.

lpsinger (approved):

PR #1914, opened 2024-02-10
ACROSS API: Remove commented code from Ephem unit tests

jak574:
# Reason

This updates the unit tests for Ephemeris code to perform the calculations with `skyfield` and `swifttools` to validate the output from ACROSS `BurstCubeEphem` and `SwiftEphem`, rather than just comparing with precomputed arrays.


requirements.txt
```diff
@@ -6,4 +6,6 @@ pytest
 types-requests
 types-cachetools
 types-python-jose
+skyfield==1.48
+swifttools==3.0.21
```
lpsinger, line 10: Why are these pinned?

lpsinger (changes requested):

requirements.txt
jak574: I figured as they were used for tests I should pin them incase a future update of either breaks their output.

jak574:
> everything looks good. All I'd recommend is to add the `de421.bsp` downloaded file to the `.gitignore`. Don't want to accidentally add and commit a 16M file to the repo.

I added this [e6e9195](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1914/commits/e6e9195d2ed11334b80e97297d88bdaf2f319ef3)

requirements.txt
lpsinger: I'd unpin them; if they break in the future then we can start restricting them.

swyatt7 (approved):
Alright. looks good

PR #1923, opened 2024-02-13
Add functionality for Corrections to be requested by users

dakota002:
Will include screenshots shortly


app/routes/_gcn.user.tsx
```diff
@@ -40,6 +40,9 @@ export default function () {
                     </NavLink>,
                   ]
                 : []),
+              <NavLink key="circulars" to="circulars">
+                Circulars
+              </NavLink>,
```
lpsinger, line 45: Why have we added this to the user menu?

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
```diff
@@ -48,20 +50,24 @@ export async function loader({
...
   )
 
-  return json(result, {
-    headers: {
-      ...publicStaticShortTermCacheControlHeaders,
-      ...getCanonicalUrlHeaders(new URL(`/circulars/${circularId}`, origin)),
-    },
-  })
+  return json(
+    { result },
```
lpsinger, line 54: Why did you wrap the result in an otherwise empty object?

lpsinger (changes requested):

app/routes/_gcn.user.tsx
dakota002: Oops, leftover from an idea i was trying. Will remove

app/routes/_gcn.circulars.$circularId.($version)/route.tsx
dakota002: Probably another holdover from something I tried and didn't end up going with. Will also remove

app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {
+  approveChangeRequest,
+  deleteChangeRequest,
+  get,
+  getChangeRequest,
+  moderatorGroup,
+} from './_gcn.circulars/circulars.server'
+import { getFormDataString } from '~/lib/utils'
+
+const DiffViewer = require('react-diff-viewer-continued').default
```
lpsinger, line 27:
```suggestion
import DiffViewer from 'react-diff-viewer-continued'
```

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -133,10 +135,30 @@ export function CircularEditForm({
...
+      saveButtonText = 'Submit'
+      break
+    case 'edit':
+      headerText = 'Edit'
+      saveButtonText = 'Update'
+      break
+    case 'new':
+      headerText = 'New'
+      saveButtonText = 'Send'
+
```
lpsinger, line 153:
```suggestion
```

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -133,10 +135,30 @@ export function CircularEditForm({
   const [showPreview, setShowPreview] = useState(false)
   const sending = Boolean(useNavigation().formData)
   const valid = subjectValid && bodyValid
+  let headerText,
+    saveButtonText = ''
```
lpsinger, line 139:
```suggestion
  let headerText, saveButtonText
```

app/routes/_gcn.circulars.correction.$circularId.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function SignInButton() {
+  const url = useUrl()
+  return (
+    <Link to={`/login?redirect=${encodeURIComponent(url)}`}>
+      <Button type="button">Sign in</Button>
+    </Link>
+  )
+}
```
lpsinger, line 98: These modal dialog boxes in the new Circular form are supposed to facilitate onboarding of new users. I don't think that we need them here. The loader can just throw a 403 to redirect them to the login.

lpsinger (changes requested):

app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
dakota002: This is intentional. When the `DiffViewer` is imported, it throws this error:
image

I can look into making a PR in their repo to fix this, but for now this is the solution

lpsinger: Yuck. Are there any other React diff viewers out there we could choose from?

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
lpsinger: It's automatically initializing `saveButtonText` to an empty string when it is defined by all paths through the switch statement? Why? Prettier normally doesn't make semantic changes like that.

app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
lpsinger: @Courey, you had poked around at some packages that we had difficulty using with ESM-style imports. Does this error message suggest anything to you?

app/routes/_gcn.circulars.edit.$circularId/CircularEditForm.tsx
dakota002: oh i thought this was a formatting change, nevermind

app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
dakota002: This seems to be the main one, there was a prior iteration that this was built off of. Corey and I went down this rabbit hole a while ago, though I don't remember where we landed with this other than this style of import

lpsinger: There is an open issue about this already: https://github.com/Aeolun/react-diff-viewer-continued/issues/40

lpsinger: There's yet another fork here that might have proper ESM support: https://www.npmjs.com/package/react-diff-viewer-esm

To be honest I am a bit worried about this dependency given the large number of forks. Also the code is quite dated (it's not even written as a React functional component, for example).

dakota002: What do you suggest? I think its definitely useful to show the changes, but not married to the idea of using this package specifically. The esm one might be good, but I don't see a link on their page to their code

dakota002: Played around a bit, and it looks like under the hood that package is just doing some fancy stuff around https://www.npmjs.com/package/diff . Here is what I got by just using uswds colors and the jsdiff package:

image

lpsinger: That looks pretty nice too.

app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
lpsinger, line 1: - Add title breadcrumb
- Exclude from sitemap

app/routes/_gcn.circulars.correction.$circularId.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const circular = await get(parseFloat(circularId))
+  const data = {
+    formattedContributor: user ? formatAuthor(user) : '',
+    defaultBody: circular.body,
+    defaultSubject: circular.subject,
+    circularId: circular.circularId,
+    submitter: circular.submitter,
+    searchString: '',
+  }
+  return { isAuthenticated, data }
```
lpsinger, line 41: We don't need to return `isAuthenticated`, do we? If the user is not authenticated, then the loader will throw a 403 response.

lpsinger (changes requested):

app/routes/_gcn.circulars.correction.$circularId.tsx
dakota002: Nope! another hold over from some previous iteration. I will remove it

app/routes/_gcn.circulars.correction.$circularId.tsx
```diff
@@ -0,0 +1,46 @@
...
+    defaultSubject: circular.subject,
+    circularId: circular.circularId,
+    submitter: circular.submitter,
+    searchString: '',
+  }
+  return { data }
+}
+
+export default function () {
+  const { data } = useLoaderData<typeof loader>()
```
lpsinger, line 44:
```suggestion
  return {
    formattedContributor: user ? formatAuthor(user) : '',
    defaultBody: circular.body,
    defaultSubject: circular.subject,
    circularId: circular.circularId,
    submitter: circular.submitter,
    searchString: '',
  }
}

export default function () {
  const data = useLoaderData<typeof loader>()
```

lpsinger (changes requested):

lpsinger (approved):

PR #1927, opened 2024-02-15
Across api add fov

swyatt7:
Adds the FOV checks for ally-sky and constrained FOV instruments.

Added classes:
In `/base/fov.py`:
- `FOVBase`: Base class for the FOV checks. Main difference from @jak574's branch is that it doesn't assume all-sky. Instead relies on computed `visible_pixels` list that is calculated from a generic order 9 healpix map. This class contains the functionality to calculate the probability contained within the `visible_pixels` based off of a single `astropy.coordinates.SkyCoord` -> `in_fov()`, for a circular error region based off of `astropy.coordinates.SkyCoord` and an `error_radius`, and finally based off of a skymap (healpix probability map).
- `AllSkyFOV`: Inherites `FOVBase`. Upon instantiation it calculates the `visible_pixels` based on the complement of the earth's position. The earth's position is calculated from a passed in `ephemeris` and `time` utilizing `healpy.query_disc()`. 
- `FootprintFOV`: Inherits `FOVBase`. Upon instantiation it calculates the `visibble_pixels` based on the instrument footprint. Utilizies `healpy.query_polygon()`.

in `/base/footprint.py`:
- `Footprint`: class that defines the polygon for instrument footprints. Contains functionality to project the footprint on a sphere based on pointing observation

in `/base/pointing.py`:
- `PointingBase`: class that loosely defines a pointing object. Very minimal, only to support proof of concept of footprint projection during BurstCube MVP. Leaving it a baseclass for future functionality scope.

Added an `/burstcube/fov.py` that defines the `BurstCubeFOV` class based off the `EarthLimbConstraint`.
Added an `/swift/fov.py` that defines the `SwiftFOV` class that is based off the `SwiftFootprint` which instantiates a `Footprint` object based on the polygon defined at [TreasureMap](https://treasuremap.space/).

Also added `/tests/fov/` for functionality testing and example usage.


jak574:
Add `astropy_healpix` to `requirements.in` and do a `pip-compile`.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+# All Rights Reserved.
+
+from typing import Optional, Union
+import astropy_healpix as ah  # type: ignore
+import astropy.units as u  # type: ignore
+from astropy.io.fits import FITS_rec  # type: ignore
+import numpy as np
+from astropy.coordinates import SkyCoord, angular_separation, spherical_to_cartesian  # type: ignore
+from astropy.time import Time  # type: ignore
+import healpy as hp
```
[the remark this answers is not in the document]
swyatt7: Yeah, I'm aware of the dependencies. But they don't have the functionality to query a polygon, which is used for the `FootprintFOV.visible_pixel` calculation.

swyatt7:
> Add `astropy-healpix` to `requirements.in` and do a `pip-compile`.

should be fixed in recent push

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,100 @@
...
+    y /= r
+    z /= r
+    theta = np.arctan2(y, x)
+    phi = np.arccos(z)
+    dec = 90 - np.rad2deg(phi)
+    if theta < 0:
+        ra = 360 + np.rad2deg(theta)
+    else:
+        ra = np.rad2deg(theta)
+    return ra, dec
```
jak574, line 29: Are these sorts of functions necessary if we're storing coordinates as `SkyCoord`? You can just do use the `cartesian` attribute of a coordinate `SkyCoord` to get a unit vector representation, and convert back also.

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,100 @@
...
+
+def z_rot(theta_deg):
+    theta = np.deg2rad(theta_deg)
+    return np.array(
+        [
+            [np.cos(theta), -np.sin(theta), 0],
+            [np.sin(theta), np.cos(theta), 0],
+            [0, 0, 1],
+        ]
+    )
```
jak574, line 62: `astropy` has a handy `rotation_matrix` function that could most of this. 
See details here: https://docs.astropy.org/en/stable/api/astropy.coordinates.CartesianRepresentation.html

Example code:
```
from astropy.coordinates.matrix_utilities import rotation_matrix
rotation = rotation_matrix(30 * u.deg, axis='z')
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+    hp = ah.HEALPix(nside=nside, order="nested")
+
+    # Find RA/Dec for each pixel in HEALPix map
+    hpra, hpdec = hp.healpix_to_lonlat(range(ah.nside_to_npix(nside)))
+
+    # Find angular distance of each HEALPix pixel to the skycoord
+    distance = angular_separation(hpra, hpdec, skycoord.ra, skycoord.dec).to(u.deg)
+
+    # Create the probability density distribution HEALPix map
+    prob = normal_pdf(distance, error_radius).value
```
lpsinger, line 84: Consider instead `scipy.stats.norm`:

```suggestion
    prob = stats.norm(scale=error_radius).pdf(distance)
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+
+    # Normalize it
+    prob = np.divide(prob, np.sum(prob))
```
lpsinger, line 87: More concise:

```suggestion
    prob /= np.sum(prob)
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+        The probability density distribution HEALPix map.
+    """
+    # Create HEALPix map
+    hp = ah.HEALPix(nside=nside, order="nested")
+
+    # Find RA/Dec for each pixel in HEALPix map
+    hpra, hpdec = hp.healpix_to_lonlat(range(ah.nside_to_npix(nside)))
+
+    # Find angular distance of each HEALPix pixel to the skycoord
+    distance = angular_separation(hpra, hpdec, skycoord.ra, skycoord.dec).to(u.deg)
```
lpsinger, line 81:
```suggestion
    hpcoord = hp.healpix_to_skycoord(np.arange(hp.npix))

    # Find angular distance of each HEALPix pixel to the skycoord
    distance = hpcoord.separation(skycoord).to(u.deg)
```

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,100 @@
...
+        for idx in range(footprint_zero_center_x.shape[0]):
+            vec = np.asarray(
+                [
+                    footprint_zero_center_x[idx],
+                    footprint_zero_center_y[idx],
+                    footprint_zero_center_z[idx],
+                ]
+            )
+            new_vec = vec @ x_rot(-pos_angle) @ y_rot(dec) @ z_rot(-ra)
+            new_x, new_y, new_z = new_vec.flat
```
lpsinger, line 96: Rather than doing the rotation manually, use [sky offset frames](https://docs.astropy.org/en/stable/coordinates/matchsep.html#sky-offset-frames).

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def project(self, ra, dec, pos_angle):
+        if pos_angle is None:
+            pos_angle = 0.0
+
+        footprint_zero_center_ra = np.asarray([pt[0] for pt in self.polygon])
+        footprint_zero_center_dec = np.asarray([pt[1] for pt in self.polygon])
+        footprint_zero_center_uvec = ra_dec_to_uvec(
+            footprint_zero_center_ra, footprint_zero_center_dec
+        )
```
lpsinger, line 79: I agree with @jak574; use astropy to [convert between coordinate representations](https://docs.astropy.org/en/stable/coordinates/representations.html) if you need to.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # Check if skycoord pixels in list of visible pixels
+        skycoord_pix = hp.ang2pix(
+            hp.order2nside(HEALPIX_MAP_EVAL_ORDER),
+            skycoord.ra.value,
+            skycoord.dec.value,
+            lonlat=True,
+            nest=True,
+        )
+        in_fov = np.ma.isin(skycoord_pix, self.visible_pixels)
```
lpsinger, line 173: [np.searchsorted](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html) might do this faster because we can take advantage of `self.visible_pixels` being pre-sorted.

python/across_api/base/pointing.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    ra: float
+    dec: float
+    position_angle: float
+
+    def __init__(self, ra, dec, position_angle) -> None:
+        self.ra = ra
+        self.dec = dec
+        self.position_angle = position_angle
```
lpsinger, line 20: I suggest removing this class and replacing it with [SkyOffsetFrame](https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyOffsetFrame.html#astropy.coordinates.SkyOffsetFrame) which consists of an Astropy coordinate origin and a position angle.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+        # Calculate the amount of probability inside the FOV
+        if healpix_order == "NUNIQ":
+            # Calculate probability in FOV by multiplying the probability density by
+            # area of each pixel and summing up
+            pixarea = ah.nside_to_pixel_area(uniq_nside[visible_probability_pixels])
+            return float(
+                round(
+                    np.sum(healpix_loc[visible_probability_pixels] * pixarea.value), 5
+                )
+            )
```
lpsinger, line 294: Why round it?

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+        else:
+            ra, dec = ah.healpix_to_lonlat(
+                nonzero_prob_pixels, nside=nside, order=healpix_order
+            )
+
+        # Convert these coordinates into a SkyCoord
+        skycoord = SkyCoord(ra=ra, dec=dec, unit="deg")
+
+        # Calculate pixel indicies of the all the regions inside of the FOV
+        visible_probability_pixels = nonzero_prob_pixels[self.in_fov(skycoord=skycoord)]
```
lpsinger, line 284: This will be inefficient if the HEALPix map is flat-resolution and matches the resolution used for the FOV intersection test, and it will be incorrect if the HEALPix map is multi-resolution or coarser in resolution than the FOV intersection test.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+    # Create the probability density distribution HEALPix map
+    prob = normal_pdf(distance, error_radius).value
+
+    # Normalize it
+    prob = np.divide(prob, np.sum(prob))
+
+    return prob
+
+
+class FOVBase(ACROSSAPIBase):
```
lpsinger, line 92: What does inheriting from `ACROSSAPIBase` do?

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+        Function does assume that it is a space mission with available ephemeris
+        """
+        n_side = hp.order2nside(HEALPIX_MAP_EVAL_ORDER)
+        n_pix = hp.order2npix(HEALPIX_MAP_EVAL_ORDER)
+        i_slice = get_slice(time=time, ephem=ephem)
+
+        ra = ephem.earth.ra[i_slice].degree
+        dec = ephem.earth.dec[i_slice].degree
+
+        vec = hp.ang2vec(ra, dec, lonlat=True)
```
lpsinger, line 352:
```suggestion
        vec = ephem.earth.cartesian.xyz.value
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,382 @@
...
+        earth_pixels = np.unique(earth_pixels)
+
+        # simulate the whole sky
+        all_sky_pixels = np.zeros(n_pix, dtype=np.float32)
+
+        # set the earth values on the sky
+        all_sky_pixels[earth_pixels] = 1
+
+        # find the complement (where sky != 1)
+        self.visible_pixels = np.where(all_sky_pixels == 0)[0]
```
lpsinger, line 382: You can avoid this set complement by instead calling `query_disc` with an angle that is 180° minus the constraint angle. (Right?)

lpsinger (changes requested):

python/across_api/base/footprint.py
swyatt7: I've re-written the entire footprint projection analysis with purely astropy Representations and Transformation methods.

python/across_api/base/footprint.py
swyatt7: I've re-written the entire footprint projection analysis with purely astropy Representations and Transformation methods.

python/across_api/base/fov.py
swyatt7: I tested this out, and wasn't getting the same results as the masked array implementation in the tests.I also found that it seemed to be around the same speed as the masked array.

python/across_api/base/pointing.py
swyatt7: I want to hold off on abstracting this to other existing classes. I am hoping/foreseeing this class becoming more robust when we want to implement more functionality when it comes to observational analysis. 

Things like `pointing_type` being `photometric` or `spectroscopic`. Other fields like `exposure_time` and `obsdate` being taken into account as well. 

This class won't be just for positional information.

If you have ideas with that scope, then I'd be down to discuss those in respect to this class.

python/across_api/base/fov.py
swyatt7: Was in Jamie's original implementation. I changed the result to be rounded in the tests instead.

python/across_api/base/fov.py
swyatt7: Was in Jamie's original implementation. I've removed it.

python/across_api/base/fov.py
swyatt7: I tried this, and it didn't give the anticipated results in the tests.

python/across_api/base/fov.py
swyatt7: In theory calculating the `hp.query_disc` with the antipodal position and a radius of `pi/2 - radius (radians)` will give you the same result. 

It shouldn't really be all that more performative since how its originally implemented is in numpy, and that is fast. 

I can implement it in a way:

```python
    def __init__(self, ephem: EphemBase, time: Time):
        """
        Finds all healpix pixels within the ephemeris time slot that are not earth
        occulted, and sets them to FOVBase.visible_pixels.

        This is done by finding the antipodal position of the earth and calculating
        all of the pixels within a disc of radius: 180 (deg) - radius_of_earth (deg)

        Antipodal:
            a_ra = ra-180 (deg)
            a_dec = -dec  (deg)
            a_radius = 180-radius (deg)

        Function does assume that it is a space mission with available ephemeris
        """
        n_side = hp.order2nside(HEALPIX_MAP_EVAL_ORDER)
        i_slice = get_slice(time=time, ephem=ephem)

        #antipodal positions
        ra = ephem.earth.ra[i_slice].degree - 180.0
        dec = -1.0 * ephem.earth.dec[i_slice].degree

        vec = hp.ang2vec(ra, dec, lonlat=True)

        # np.array to store visible pixels
        visible_pixels = np.array([]).astype(int)

        # loop over each time-slice vector and concatenate
        # visible pixels calculated with healpy's cone search (query_disc)
        # assume the earth is a disc at ephem.earth position with radius
        # radius = 180 - earthsize + constraint min angle.
        for i, v in enumerate(vec):
            radius = np.radians(180) - np.radians(ephem.earthsize[i].value) + np.radians(
                self.earth_constraint.min_angle.value
            )
            visible_pixels = np.concatenate(
                (
                    visible_pixels,
                    hp.query_disc(n_side, v, radius, nest=True, inclusive=False),
                )
            )

        # only need unique pixels
        self.visible_pixels = np.unique(visible_pixels)
```

swyatt7: That passed the tests, so I'll push it since it will be more efficient.

python/across_api/base/fov.py
lpsinger: If you want to use `isin`, then please use the ndarray version and not the matrix version:

```suggestion
        in_fov = np.isin(skycoord_pix, self.visible_pixels)
```

Here is a version with searchsorted.

```
def is_in_sorted_array(a, v):
    i = np.clip(np.searchsorted(a, v), 0, len(a) - 1)
    return a[i] == v
```

I would think that `np.isin` is O(N log N + M log N) where N and M are the lengths of the two arrays, because it must first sort the array. However if it uses a sufficiently clever sort then it might be only O(N + M log N) if the array is pre-sorted. However, the `is_in_sorted_array` is only O(M log N).

This is definitely a micro-optimization given the small sizes of these arrays.

python/across_api/base/fov.py
lpsinger: This is still incorrect, or at least inaccurate. What you should do is upsample both HEALPix maps to the higher of the two resolutions and then sum. That would be faster and more accurate.

There should be no need to do any sky coordinate to HEALPix conversion here. Upsampling a HEALPix map just means repeating every pixel (nside_new / nside_old)**2 times, whereas these coordinate conversions involve unnecessary trigonometry.

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,43 @@
...
+
+        projected_ra = [
+            round(lon.to(u.deg).value, 3) for lon in projected_footprint.lon
+        ]
+        projected_dec = [
+            round(lat.to(u.deg).value, 3) for lat in projected_footprint.lat
+        ]
+
+        return list(zip(projected_ra, projected_dec))
```
lpsinger, line 43: Again, I suggest using sky offset frames. Roughly like this:

```suggestion
    def project(self, center: SkyCoord, pos_angle: u.Quantity[u.deg] = 0 * u.deg):
        representation = (
            UnitSphericalRepresentation(
                u.Quantity([pt[0] for pt in self.polygon], u.deg),
                u.Quantity([pt[1] for pt in self.polygon], u.deg),
            )
        )
        return SkyCoord(representation, frame=center.skyoffset_frame(pos_angle).icrs
```

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,43 @@
...
+from astropy import units as u  # type: ignore[import]
+from astropy.coordinates.representation import (  # type: ignore[import]
+    CartesianRepresentation,
+    UnitSphericalRepresentation,
+)
+from astropy.coordinates.matrix_utilities import rotation_matrix  # type: ignore[import]
+
+
+class Footprint:
+    polygon: list = []
```
lpsinger, line 15: Store polygon as a frame-free Representation rather than as a list.

```suggestion
    polygon: UnitSphericalRepresentation
```

python/across_api/base/footprint.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, polygon: list) -> None:
+        self.polygon = polygon
```
lpsinger, line 18: You could make this a dataclass and thereby remove the need for an explicit __init__ method.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,351 @@
...
+from .footprint import Footprint
+from .pointing import PointingBase
+from .ephem import EphemBase
+
+HEALPIX_MAP_EVAL_ORDER = 9
+
+
+def healpix_map_from_position_error(
+    skycoord: SkyCoord,
+    error_radius: u.Quantity,
```
lpsinger, line 24:
```suggestion
    error_radius: u.Quantity[u.deg],
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+            visible_pixels = np.concatenate(
+                (
+                    visible_pixels,
+                    hp.query_disc(n_side, v, radius, nest=True, inclusive=False),
+                )
+            )
+
+        # only need unique pixels
+        self.visible_pixels = np.unique(visible_pixels)
```
lpsinger, line 351: You may be able to do this even faster using query_polygon. query_polygon is implemented by calling query_disc for every edge, then taking the intersection. To employ query_disc to take the union, employ de Morgan's Law.

lpsinger (changes requested):
Overall comments:

- Please replace healpy with astropy-healpix. There are several reasons:
   1. Healpy doesn't currently have builds for Windows, so Healpy would break local development for several contributors.
   2. astropy-healpix is an Astropy affiliated package, so it has better integration with Astropy --- for example, better integration with astropy.coordinates.
   3. Healpy has more dependencies than astropy-healpix.
- Replace low-level rotations to compute projected footprints with [sky offset frames](https://docs.astropy.org/en/stable/coordinates/matchsep.html#sky-offset-frames).
- Use Astropy unit conversion rather than `np.radian`. Keep quantities unitful until as late as possible.
- Where you need coordinates in specific representations, use Astropy's representation methods rather than `hp.ang2vec`, `hp.vec2ang`. Delay conversion to a specific representation until as late as possible.
- Where you take astropy.unit.Quantity type arguments, include the compatible unit type in the annotation --- e.g., `u.Quantity[u.deg]`.

python/across_api/base/fov.py
lpsinger: Scratch that; these are not great circles, so `query_polygon` wouldn't work.

python/across_api/base/fov.py
swyatt7: Implemented in recent push

python/across_api/base/footprint.py
swyatt7: I kept the explicit `__init__` and instead of having `polygon` being the variable set, I just instantiated the representation field.

python/across_api/base/footprint.py
swyatt7: Implemented in recent push

python/across_api/base/footprint.py
swyatt7: Implemented in recent push. Makes the `footprint.py` file very concise.

python/across_api/base/fov.py
swyatt7: We discussed that this works _well enough_ at the moment, we can create an issue on this after the BurstCube MVP

python/across_api/base/fov.py
swyatt7: I tested both, and actually found that `np.isin()` is actually faster for our tests. Not by much, but it's less lines of code and more readable, so I went with that

swyatt7:
> Member



> Overall comments:
> 
> * ~Please replace healpy with astropy-healpix. There are several reasons:~
>   
>   1. Healpy doesn't currently have builds for Windows, so Healpy would break local development for several contributors.
>   2. astropy-healpix is an Astropy affiliated package, so it has better integration with Astropy --- for example, better integration with astropy.coordinates.
>   3. Healpy has more dependencies than astropy-healpix.
> * Replace low-level rotations to compute projected footprints with [sky offset frames](https://docs.astropy.org/en/stable/coordinates/matchsep.html#sky-offset-frames).
> * Use Astropy unit conversion rather than `np.radian`. Keep quantities unitful until as late as possible.
> * Where you need coordinates in specific representations, use Astropy's representation methods rather than `hp.ang2vec`, `hp.vec2ang`. Delay conversion to a specific representation until as late as possible.
> * Where you take astropy.unit.Quantity type arguments, include the compatible unit type in the annotation --- e.g., `u.Quantity[u.deg]`.
> 
> EDIT: Don't convert from healpy to astropy-healpix at this time. Its cone search implementation is significantly slower than Astropy's for large radii, and unfortunately for the polygon search we would need the largest possible radii (90°, great circles). I have enlisted @dakota002's help with porting Healpy to Windows.

- I kept the `healpy` dependency and implementations
- Low level rotations for projected footprints are computed with sky offset frames
- Astropy unit conversion implemented for astropy objects
- I still kept the hp.ang2vec in there since we are still using healpy
- `Quantity` unit types implemented

python/across_api/base/fov.py
```diff
@@ -0,0 +1,354 @@
...
+        -------
+        bool
+                True or False
+        """
+
+        # Check if skycoord pixels in list of visible pixels
+        skycoord_pix = hp.ang2pix(
+            hp.order2nside(HEALPIX_MAP_EVAL_ORDER),
+            skycoord.ra.value,
+            skycoord.dec.value,
```
lpsinger, line 141: Don't assume that the units are degrees.

```suggestion
            skycoord.ra.deg,
            skycoord.dec.deg,
```

python/across_api/base/fov.py
```diff
@@ -0,0 +1,354 @@
...
+        )
+
+        ras_radians = projected_footprint.ra.to(u.rad).value[:-1]
+        decs_radians = projected_footprint.dec.to(u.rad).value[:-1]
+
+        cartesian_vertices = spherical_to_cartesian(
+            1,  # radial component
+            decs_radians,  # LAT - DECs
+            ras_radians,  # LON - RAs
+        )
```
lpsinger, line 291:
```suggestion
        cartesian_vertices = projected_footprint.cartesian.xyz.value
```

I'm not sure about the shape; you might need to remove the transpose below.

python/across_api/base/fov.py
```diff
@@ -0,0 +1,354 @@
...
+        # loop over each time-slice vector and concatenate
+        # visible pixels calculated with healpy's cone search (query_disc)
+        # assume the earth is a disc at ephem.earth position with radius
+        # radius = 180 - earthsize + constraint min angle.
+        for i, v in enumerate(vec):
+            radius = (
+                np.radians(180)
+                - ephem.earthsize[i].to(u.rad).value
+                + self.earth_constraint.min_angle.to(u.rad).value
+            )
```
lpsinger, line 345:
```suggestion
            radius = (
                180 * u.deg
                - ephem.earthsize[i]
                + self.earth_constraint.min_angle
            ).to_value(u.rad)
```

lpsinger (changes requested):

python/across_api/base/fov.py
jak574: This works, no other changes needed: 

`cartesian_vertices = projected_footprint[:-1].cartesian.xyz.value`

jak574 (approved):

PR #1933, opened 2024-02-16
Courey/create synonym server only

Courey:
### Title   
 feat: add create function to synonym server

### Description
This function takes an array of eventIds that are synonymous to each other and creates a grouping of synonym records. 

### Reviewers
Leo - because he's already reviewed it before i finished this comment.

### Changes
This adds a function that creates synonyms based off eventIds passed in. It returns an array of synonym records that were created.

### Acceptance Criteria
a user can pass an array of eventIds in and it will create the expected synonym records for each eventId.

### DIFF
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -1,0 +1,25 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { tables } from '@architect/functions'
+import crypto from 'crypto'
+
+export async function createSynonyms({
+  synonymousEventIds,
+}: {
+  synonymousEventIds: string[]
+}) {
+  const uuid = crypto.randomUUID()
+  const db = await tables()
+  await Promise.all(
+    synonymousEventIds.map((eventId) => {
+      if (!eventId) return null
+      return db.synonyms.put({ uuid, eventId })
+    })
+  )
+  return uuid
+}
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -13,7 +13,7 @@ import crypto from 'crypto'
  * eventId from the old synonym and the only remaining link will be to the
  * new synonym.
  */
-export async function createSynonyms(...synonymousEventIds: string[]){
+export async function createSynonyms(...synonymousEventIds: string[]) {
   const uuid = crypto.randomUUID()
   const db = await tables()
   await Promise.all(
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -9,10 +9,10 @@ import { tables } from '@architect/functions'
 import crypto from 'crypto'
 
 /*
-If an eventId already has a synonym and is passed in, it will unlink the
-eventId from the old synonym and the only remaining link will be to the
-new synonym
-*/
+ * If an eventId already has a synonym and is passed in, it will unlink the
+ * eventId from the old synonym and the only remaining link will be to the
+ * new synonym.
+ */
 export async function createSynonyms(...synonymousEventIds: string[]){
   const uuid = crypto.randomUUID()
   const db = await tables()
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -8,11 +8,12 @@
 import { tables } from '@architect/functions'
 import crypto from 'crypto'
 
-export async function createSynonyms({
-  synonymousEventIds,
-}: {
-  synonymousEventIds: string[]
-}) {
+/*
+If an eventId already has a synonym and is passed in, it will unlink the
+eventId from the old synonym and the only remaining link will be to the
+new synonym
+*/
+export async function createSynonyms(...synonymousEventIds: string[]){
   const uuid = crypto.randomUUID()
   const db = await tables()
   await Promise.all(
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -6,21 +6,37 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { tables } from '@architect/functions'
+import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb/dist-types/DynamoDBDocument'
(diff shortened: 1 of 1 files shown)


app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { tables } from '@architect/functions'
+import crypto from 'crypto'
+
+export async function createSynonyms({
+  synonymousEventIds,
+}: {
+  synonymousEventIds: string[]
+}) {
```
lpsinger, line 15:
```suggestion
export async function createSynonyms(...synonymousEventIds: string[]) {
```

lpsinger (changes requested):
Remember that the terms that have been passed to the function may _themselves_ already have synonyms, and so you either need to:

1) Link all of _their_ synonyms with a common UUID, or
2) Decide (and document) that this function will unlink any mutual synonyms of those terms that you do not pass to it.

See, for example, https://github.com/nasa-gcn/gcn.nasa.gov/blob/cadd4e4378244bd37aa4a3aced7cf49c89c252f0/app/lib/synonyms.server.ts#L115-L154.

Courey:
The current behavior:
1. if there is no existing link, it just creates the new synonyms
2. if there is ONE existing link, it turns all the new synonyms into synonyms of the existing one
3. if there are more than one existing link, and they are all the same, it turns the new synonyms into synonyms of these
4. if there are more than one existing link, and they are not the same, it throws a 422 because it won't ever be able to be processed as is, an gives some context of the involved synonyms

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -0,0 +1,79 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { tables } from '@architect/functions'
+import crypto from 'crypto'
+
+import type { Synonym } from './synonyms.lib'
+
+export function filterEmpties({ ids }: { ids: string[] }) {
+  return ids.filter(function (element) {
+    return element != (null || '')
```
lpsinger, line 15: What is this function supposed to do? The expression `(null || '')` always evaluates to `''` due to short circuit evaluation, so this is equivalent to:

```ts
return ids.filter(function (element) {
  return element != ''
}
```

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -0,0 +1,79 @@
...
+  uuid,
+}: {
+  eventIds: string[]
+  uuid: string
+}) {
+  const db = await tables()
+  await Promise.all(
+    eventIds.map((eventId) => {
+      return db.synonyms.put({ uuid, eventId })
+    })
```
lpsinger, line 40:
```suggestion
    eventIds.map((eventId) => db.synonyms.put({ uuid, eventId }))
```

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -0,0 +1,79 @@
...
+  const db = await tables()
+  const existingSynonyms = [] as Synonym[]
+  eventIds.forEach(async (eventId) => {
+    const hasExisting = (await db.synonyms.get({ eventId })) as Synonym
+    if (hasExisting) existingSynonyms.push(hasExisting)
+  })
+  return existingSynonyms
+}
+
+export async function insertSynonyms({
```
lpsinger, line 29:
```suggestion
async function insertSynonyms({
```

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import type { Synonym } from './synonyms.lib'
+
+export function filterEmpties({ ids }: { ids: string[] }) {
+  return ids.filter(function (element) {
+    return element != (null || '')
+  })
+}
+
+export async function checkForExisting({ eventIds }: { eventIds: string[] }) {
```
lpsinger, line 19:
```suggestion
async function checkForExisting({ eventIds }: { eventIds: string[] }) {
```

lpsinger (commented):
I question whether all of this validation is necessary. You could just as easily document that the function expects, as a precondition, that none of the terms that you pass have existing synonyms.

app/routes/_gcn.synonyms/synonyms.server.ts
Courey: yup. if there is a blank element passed in, it takes it out.

Courey: I don't have a front end, I don't know what validations are in place, I just know that if you pass in an empty element, it will break, so we protect for that.

Courey:
> I question whether all of this validation is necessary. You could just as easily document that the function expects, as a precondition, that none of the terms that you pass have existing synonyms.

Where, where do I document this?

Courey:
> Remember that the terms that have been passed to the function may themselves already have synonyms, and so you either need to:
> 
>     Link all of their synonyms with a common UUID, or

is the why behind:

> I question whether all of this validation is necessary. You could just as easily document that the function expects, as a precondition, that none of the terms that you pass have existing synonyms.

if you wanted me to do it the second way, why didn't you just tell me to document it?

app/routes/_gcn.synonyms/synonyms.server.ts
lpsinger: Yes, but why do you have `(null || '')`?

```
$ node
Welcome to Node.js v20.11.1.
Type ".help" for more information.
> (null || '')
''
```

Did you mean this?


```
return element !== undefined && element !== ''
```

lpsinger:
> > Remember that the terms that have been passed to the function may themselves already have synonyms, and so you either need to:
> > ```
> > Link all of their synonyms with a common UUID, or
> > ```
> 
> is the why behind:
> 
> > I question whether all of this validation is necessary. You could just as easily document that the function expects, as a precondition, that none of the terms that you pass have existing synonyms.
> 
> if you wanted me to do it the second way, why didn't you just tell me to document it?

That's what I was trying to say with the second option I proposed:

> Decide (and document) that this function will unlink any mutual synonyms of those terms that you do not pass to it.

Note that if you do go this route, then you don't need to add a function to unlink synonyms, because that's a degenerate case of the same behavior.

Courey:
So one concern I have with the just linking it to another synonym is that is can leave synonym groups of 1 in its wake. Is that acceptable or should there then be additional complexity to remove the synonym entirely if it only has one remaining synonym?

lpsinger:
> So one concern I have with the just linking it to another synonym is that is can leave synonym groups of 1 in its wake. Is that acceptable or should there then be additional complexity to remove the synonym entirely if it only has one remaining synonym?

An isolated synonym is fine.

lpsinger (changes requested):

lpsinger (approved):

PR #1957, opened 2024-02-20
Adds a confirmation alert

### LINKED ISSUE #1930: Add confirmation messages when submitting a correction request or making a change
Show some confirmation to the user that the action was completed successfully. This could be as simple as a toast message on the Circulars archive page.

dakota002:
Closes #1930

### DIFF
--- a/app/routes/_gcn.circulars._archive._index/route.tsx
+++ b/app/routes/_gcn.circulars._archive._index/route.tsx
@@ -15,6 +15,7 @@ import {
   useSubmit,
 } from '@remix-run/react'
 import {
+  Alert,
   Button,
   ButtonGroup,
   Icon,
@@ -93,7 +94,8 @@ export async function action({ request }: ActionFunctionArgs) {
   const user = await getUser(request)
   const circularId = getFormDataString(data, 'circularId')
 
-  let result
+  let newItem
+  let showSuccess = false
   const props = { body, subject, ...(format ? { format } : {}) }
   switch (intent) {
     case 'correction':
@@ -103,7 +105,8 @@ export async function action({ request }: ActionFunctionArgs) {
         { circularId: parseFloat(circularId), ...props },
         user
       )
-      result = null
+      showSuccess = true
+      newItem = null
       break
     case 'edit':
       if (!circularId)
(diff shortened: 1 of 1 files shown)


Courey (approved):

lpsinger (changes requested):
Rather than returning a new boolean, how about returning `intent` so that you can, if need be, display a different confirmation message depending on the operation?

app/routes/_gcn.circulars._archive._index/route.tsx
```diff
@@ -115,24 +118,27 @@ export async function action({ request }: ActionFunctionArgs) {
...
       break
     case 'new':
-      result = await put({ ...props, submittedHow: 'web' }, user)
+      newItem = await put({ ...props, submittedHow: 'web' }, user)
       break
     default:
       break
   }
-  return result
+  return { newItem, showSuccess }
```
lpsinger, line 129: Rather than call it `result`, how about `newCircular`?

lpsinger (changes requested):

app/routes/_gcn.circulars._archive._index/route.tsx
dakota002: In the action? Or in the component function? In the latter that will lead to it looking like 

```ts
const newCircular= useActionData<typeof action>() 
// const {newItem, intent} = useActionData<typeof action>() throws 
// errors about newItem and intent not being defined
...
const allItems = [
    ...(newCircular?.newItem ? [newCircular.newItem] : []),
    ...(items || []),
  ]
```

lpsinger:
```suggestion
  return { newCircular, intent }
```

lpsinger (changes requested):

lpsinger (approved):

dakota002:
@lpsinger What are your thoughts on this? Went with your wording on the sentence and moved the 'Request Submitted' text to the heading of the alert

image

lpsinger:
Yup, looks good!

dakota002:
Nice, new PR out for it ^

PR #1958, opened 2024-02-20
Announcement template

dakota002:


lpsinger (approved):
Looks good, but @jracusin should also review.

jracusin:
Content looks fine, but could I see a screenshot of what it'll look like from the editor's perspective?

dakota002:
@jracusin 
image

jracusin:
Please make it clear that the footer material is included, so the person drafting doesn't duplicate it.

lpsinger:
> Please make it clear that the footer material is included, so the person drafting doesn't duplicate it.

You could add a Preview button. I could perhaps refactor some of the Markdown editor stuff so that we have a consistent look and feel.

lpsinger:
> > Please make it clear that the footer material is included, so the person drafting doesn't duplicate it.
> 
> You could add a Preview button. I could perhaps refactor some of the Markdown editor stuff so that we have a consistent look and feel.

Nah, ignore me. That's a lot of extra work. It's a possible future enhancement, but not something we need right now.

dakota002:
@lpsinger  @jracusin 

I added some descriptive text to both text boxes. The button is a toggle to show/hide the footer text, same way the Known Keywords section behaves on the circulars page

image

app/routes/_gcn.news.email.tsx
```diff
@@ -21,14 +22,17 @@ import { useState } from 'react'
...
+  sendAnnouncementEmail,
+} from './_gcn.user.email/email_announcements.server'
 import { getFormDataString } from '~/lib/utils'
 
 export async function loader({ request }: LoaderFunctionArgs) {
   const user = await getUser(request)
   if (!user?.groups.includes(moderatorGroup))
     throw new Response(null, { status: 403 })
-  return null
+  return announcementAppendedText
```
lpsinger, line 35: Rather than pass this through the loader, just import it. Move the definition of this symbol to a file whose extension is not `.server.ts`.

lpsinger (changes requested):

app/routes/_gcn.user.email/email_announcements.server.ts
```diff
@@ -12,18 +12,9 @@ import { dedent } from 'ts-dedent'
 
 import type { User } from '../_gcn._auth/user.server'
 import { moderatorGroup } from '../_gcn.circulars/circulars.server'
+import { announcementAppendedText } from '../_gcn.news.email'
```
lpsinger, line 15: I find it a bit strange that a library module imports a route module. Would you please instead put this definition inside a new file, `email_announcements.ts`?

lpsinger (changes requested):

app/routes/_gcn.user.email/email_announcements.server.ts
lpsinger: I don't think so. That's the only part that you need in both client and server bundles, right?

dakota002: Yup, just to keep the text consistent so if it changes we only have to update it in one place. I'll push that out shortly

lpsinger (approved):

lpsinger:
Looks good. Please merge when the pipeline finishes.

Note, we now use this expand/collapse widget in a few places. Would you please open a follow-up issue to refactor and consolidate those into a common component?

PR #1973, opened 2024-02-22
ACROSS API: Add parameter to indicate ordering for HEALPix maps.

jak574:
The `probability_in_fov` method of `BaseFOV` did not support passing through the ordering of the HEALPix map. This meant that only nested maps were supported. Adding this parameter as a pass through to `in_fov_healpix_map` enables ingestion HEALPix maps with other ordering, most importantly `NUNIQ` AKA Multi-Order Coverage maps.

### DIFF
--- a/python/across_api/base/fov.py
+++ b/python/across_api/base/fov.py
@@ -69,6 +69,7 @@ class FOVBase:
         skycoord: Optional[SkyCoord] = None,
         error_radius: Optional[u.Quantity[u.deg]] = None,
         healpix_loc: Optional[FITS_rec] = None,
+        healpix_order: str = "NESTED",
     ) -> float:
         """
         For a given sky position and error radius, calculate the probability of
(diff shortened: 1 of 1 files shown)


swyatt7 (changes requested):
This all looks good. One thing:

Can we change the variable name of `healpix_order` to `healpix_scheme` or `healpix_ordering`, since it could be confused with the actual order (corresponding to nside) of the healpix map.

jak574:
> This all looks good. One thing:
> 
> Can we change the variable name of `healpix_order` to `healpix_scheme` or `healpix_ordering`, since it could be confused with the actual order (corresponding to nside) of the healpix map.

Sure thing. Fixed in [e13ae0e](https://github.com/nasa-gcn/gcn.nasa.gov/pull/1973/commits/e13ae0ed2242747f58c0489290ec3ae480fc9520).

swyatt7 (approved):

PR #1986, opened 2024-02-26
adding update synonym function

### LINKED ISSUE #1985: Add server only implementation of updating synonyms
### Issue

* [ ] This is a bug.
* [x] This is a feature request.
* [x] Did you check for similar existing issues?
* [ ] If this is a bug, can you reproduce the problem?

### Description

Add a server only function that is currently unused. It must take a uuid (synonym id), a list of eventIds to add, and a list of eventIds to delete. The function must add records appropriately for the eventIds to add, and delete appropriately the records with eventIds in the delete list. It should return the updated synonym.

### Acceptance Criteria

A function exists which (once connected) will allow a moderator to add and/or remove events from synonym groups.

Courey:
### Title
feat: adds update to synonyms server

### Description
This adds functionality to update existing synonyms. It allows you to pass in additions and subtractions (eventIds to add or remove from the synonym) to update the existing synonym members.

### Reviewers
@dakota002 - this is something either one can review
@lpsinger  - this is something either one can review

### Changes
This adds one function that is currently unused. It is intended to be used to update synonyms.
This also changes one index. We don't typically have the eventId in cases where we want to get synonyms by uuid, so I removed the eventId constraint so it would just be uuid. This index is currently unused.

### Acceptance Criteria
The code is unused currently, so review for syntax, format, and general performance.

### Related Issue(s)
Resolves #1985 

### Testing
This code is unused.

### DIFF
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -61,34 +61,21 @@ export async function putSynonyms({
   const db = await tables()
   const client = db._doc as unknown as DynamoDBDocument
   const writes = []
-  if (!subtractions && !additions) {
-    const { Items } = await db.synonym.query({
-      IndexName: 'synonymsByUuid',
-      KeyConditionExpression: 'uuid = :uuid',
-      ExpressionAttributeValues: {
-        ':uuid': uuid,
-      },
-    })
-    return Items
-  }
+  if (!subtractions && !additions) return
   if (subtractions) {
-    const subtraction_writes = subtractions
-      .filter((eventId) => Boolean(eventId))
-      .map((eventId) => ({
-        DeleteRequest: {
-          Key: { uuid, eventId },
-        },
-      }))
+    const subtraction_writes = subtractions.map((eventId) => ({
+      DeleteRequest: {
+        Key: { uuid, eventId },
+      },
+    }))
     writes.push(subtraction_writes)
   }
   if (additions) {
-    const addition_writes = additions
-      .filter((eventId) => Boolean(eventId))
-      .map((eventId) => ({
-        PutRequest: {
-          Item: { uuid, eventId },
-        },
-      }))
+    const addition_writes = additions.map((eventId) => ({
+      PutRequest: {
+        Item: { uuid, eventId },
+      },
+    }))
     writes.push(addition_writes)
   }
   const params = {
--- a/app/routes/_gcn.synonyms/synonyms.server.ts
+++ b/app/routes/_gcn.synonyms/synonyms.server.ts
@@ -40,3 +40,52 @@ export async function createSynonyms(...synonymousEventIds: string[]) {
 
   return uuid
 }
+
+/*
+ * If an eventId already has a synonym and is passed in, it will unlink the
+ * eventId from the old synonym and the only remaining link will be to the
+ * new synonym.
+ *
+ * BatchWriteItem has a limit of 25 items, so the user may not add and/or remove
+ *  more than 25 synonyms at a time.
+ */
+export async function putSynonyms({
+  uuid,
+  additions,
+  subtractions,
+}: {
+  uuid: string
+  additions?: string[]
+  subtractions?: string[]
+}) {
+  const db = await tables()
+  const client = db._doc as unknown as DynamoDBDocument
+  const writes = []
+  if (!subtractions && !additions) return await db.synonyms.get({ uuid })
+  if (subtractions) {
+    const subtraction_writes = subtractions
+      .filter((eventId) => Boolean(eventId))
+      .map((eventId) => ({
+        DeleteRequest: {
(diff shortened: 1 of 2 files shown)


lpsinger (commented):
I thought that we decided in #1933 that the way that we use the `createSynonyms` function is that we pass the full, desired list of synonyms to it, and that it is alright and even intentional that omitting some terms that are synonyms will cause them to be unlinked.

What are the intended use cases of `createSynonyms` vs. `putSynonyms`?

Courey:
If you have an existing synonym that has existing eventIds, and you remove one, how does the original function delete the record that isn't being replaced? 
For example: 
I have synonym abc. 
it has events 123, 456, 789. 
I remove 789.
in the original function we only have the put request, not any deletes. 
So there would be no way of deleting the event from the synonym. 

We could combine the create with the put and call it put, and add additional functionality to delete when necessary. But as is I do not believe the create function updates things as expected.

lpsinger:
Suppose that you have terms `foo`, `bar`, `bat`, and `baz`, which are all synonyms because they have the same UUID 1234. If you call `putSynonyms('foo', 'bar')` and it generates the UUID 4567. Then `foo` and `bar` are synonyms and `bat` and `baz` are synonyms, but they are no longer all synonyms with each other.

This is starting to get into how you will be using this backend calls once we move onto reviewing the front end. I'll just review this PR with the assumption that it meets your needs for that.

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -40,3 +40,52 @@ export async function createSynonyms(...synonymousEventIds: string[]) {
...
+  subtractions,
+}: {
+  uuid: string
+  additions?: string[]
+  subtractions?: string[]
+}) {
+  const db = await tables()
+  const client = db._doc as unknown as DynamoDBDocument
+  const writes = []
+  if (!subtractions && !additions) return await db.synonyms.get({ uuid })
```
lpsinger, line 64: The partition key is `eventId`, not `uuid`. You can only query by eventId if you use the secondary index.

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -40,3 +40,52 @@ export async function createSynonyms(...synonymousEventIds: string[]) {
...
+ *
+ * BatchWriteItem has a limit of 25 items, so the user may not add and/or remove
+ *  more than 25 synonyms at a time.
+ */
+export async function putSynonyms({
+  uuid,
+  additions,
+  subtractions,
+}: {
+  uuid: string
```
lpsinger, line 57: I had envisioned the uuid as an implementation detail that is not exposed in the API.

lpsinger (changes requested):

app/routes/_gcn.synonyms/synonyms.server.ts
lpsinger: No, it's not a problem. My original design in #745 did not use it. I was just remarking on a difference in design; that is all.

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -40,3 +40,61 @@ export async function createSynonyms(...synonymousEventIds: string[]) {
...
+  const writes = []
+  if (!subtractions && !additions) {
+    const { Items } = await db.synonym.query({
+      IndexName: 'synonymsByUuid',
+      KeyConditionExpression: 'uuid = :uuid',
+      ExpressionAttributeValues: {
+        ':uuid': uuid,
+      },
+    })
+    return Items
```
lpsinger, line 72: You return the list of items if there are no subtractions or additions. However, you return nothing if there are additions or subtractions. Please use one of of those two behaviors consistently in both cases.

app/routes/_gcn.synonyms/synonyms.server.ts
```diff
@@ -40,3 +40,61 @@ export async function createSynonyms(...synonymousEventIds: string[]) {
...
+  }
+  if (subtractions) {
+    const subtraction_writes = subtractions
+      .filter((eventId) => Boolean(eventId))
```
lpsinger, line 76: This checks that `eventId` is not an empty string. While it would be unusual if it was an empty string, nothing would actually break. Is this validation necessary, and is it this function's responsibility?

```suggestion
```

lpsinger (changes requested):

lpsinger (changes requested):
There are two bugs here:
- [x] The condition `if (!subtractions && !additions)` does not catch the case that these are both defined but as zero-length arrays, which I think would also cause `client.batchWrite` to raise an error.
- [x] The name of the table is _not_ `synonyms`. That is the logical name used by Architect; you need to look up the actual table name. Refer to other instances of batch or transactional writes in our code for examples.

lpsinger (approved):

PR #1989, opened 2024-02-26
Hide dropdown content when clicking away 

tylerbarna:
This adds basic functionality to the DetailsDropdownContent component for listening for clicking away from a dropdown, which will cause the dropdown to be hidden from the user unless the dropdown button is pressed again. This will prevent dropdown menus drawing over one another in the circulars archive like so:

image

and users won't have to manually hide dropdown menus should they choose to not impose any of the additional search filters.

### DIFF
--- a/app/routes/circulars._archive._index/DateSelectorMenu.tsx
+++ b/app/routes/circulars._archive._index/DateSelectorMenu.tsx
@@ -17,7 +17,7 @@ import {
   Radio,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
-import { type ChangeEvent, forwardRef, useRef, useState } from 'react'
+import { type ChangeEvent, useRef, useState } from 'react'
 import { useOnClickOutside } from 'usehooks-ts'
 
 import DetailsDropdownContent from '~/components/DetailsDropdownContent'
@@ -69,17 +69,15 @@ function DateSelectorButton({
   )
 }
 
-export const DateSelector = forwardRef<
-  HTMLDivElement,
-  {
-    form?: string
-    defaultStartDate?: string
-    defaultEndDate?: string
-  }
->(function DateSelector(
-  { form, defaultStartDate, defaultEndDate },
-  ref: React.Ref<HTMLDivElement>
-) {
+export function DateSelector({
+  form,
+  defaultStartDate,
+  defaultEndDate,
+}: {
+  form?: string
+  defaultStartDate?: string
+  defaultEndDate?: string
+}) {
   const dateSelectorRef = useRef<HTMLDivElement>(null)
   const [showDateSelector, setShowDateSelector] = useState(false)
   useOnClickOutside(dateSelectorRef, () => {
@@ -227,4 +225,4 @@ export const DateSelector = forwardRef<
       </DetailsDropdownContent>
     </div>
   )
-})
+}
--- a/app/routes/circulars._archive._index/DateSelectorMenu.tsx
+++ b/app/routes/circulars._archive._index/DateSelectorMenu.tsx
@@ -69,163 +69,167 @@ function DateSelectorButton({
 }
 
 interface DateSelectorProps {
-  form?: string;
-  defaultStartDate?: string;
-  defaultEndDate?: string;
-  showDateSelector: boolean;
-  setShowDateSelector: React.Dispatch<React.SetStateAction<boolean>>;
+  form?: string
+  defaultStartDate?: string
+  defaultEndDate?: string
+  showContent: boolean
+  setShowContent: React.Dispatch<React.SetStateAction<boolean>>
 }
 
-export const DateSelector = forwardRef<HTMLDivElement, DateSelectorProps>(function DateSelector(
-  {
-    form,
-    defaultStartDate,
-    defaultEndDate,
-    showDateSelector,
-    setShowDateSelector,
-  },
-  ref: React.Ref<HTMLDivElement>
-) {
-  const defaultShowDateRange = Boolean(
-    (defaultStartDate && !dateSelectorLabels[defaultStartDate]) ||
-      defaultEndDate
-  )
-  const [showDateRange, setShowDateRange] = useState(defaultShowDateRange)
-  const startDateInputRef = useRef<HTMLInputElement>(null)
-  const endDateInputRef = useRef<HTMLInputElement>(null)
-  const submit = useSubmit()
+export const DateSelector = forwardRef<HTMLDivElement, DateSelectorProps>(
+  function DateSelector(
+    {
+      form,
+      defaultStartDate,
+      defaultEndDate,
+      showContent: showDateSelector,
+      setShowContent,
+    },
+    ref: React.Ref<HTMLDivElement>
+  ) {
+    const defaultShowDateRange = Boolean(
+      (defaultStartDate && !dateSelectorLabels[defaultStartDate]) ||
+        defaultEndDate
+    )
+    const [showDateRange, setShowDateRange] = useState(defaultShowDateRange)
+    const startDateInputRef = useRef<HTMLInputElement>(null)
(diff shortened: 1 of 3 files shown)


tylerbarna:
The current commit seems to interfere with the dropdown button interaction slightly; if the menu is expanded, pressing the button again does not collapse the menu, though clicking anywhere else on the page will collapse the dropdown. This is not the case in the currently deployed commit, so it seems to be due to the implementation made in this PR.

tylerbarna:
@dakota002 would this affect the use of [DetailsDropdownContent in the schema browser docs](https://github.com/nasa-gcn/gcn.nasa.gov/blob/f657b161c35f3cd92ec2ecfb03f804c2be122b76/app/routes/_gcn.docs_._schema-browser.schema.(%24version).%24/route.tsx#L141)?

dakota002:
> @dakota002 would this affect the use of [DetailsDropdownContent in the schema browser docs](https://github.com/nasa-gcn/gcn.nasa.gov/blob/f657b161c35f3cd92ec2ecfb03f804c2be122b76/app/routes/_gcn.docs_._schema-browser.schema.(%24version).%24/route.tsx#L141)?

Without checking, they should work well together, assuming that the property is not required. You should be able to test it out locally though. If you are having issues with the github content on your local let me know

lpsinger (changes requested):
Please rebase.

Please use the existing hooks [useOnClickOutside](https://usehooks-ts.com/react-hook/use-on-click-outside) or [useClickAnywhere](https://usehooks-ts.com/react-hook/use-click-any-where).

tylerbarna:
Change to useOnClickOutside changes issue from button not minimizing at all to it minimizing before immediately being expanded again; this is because the DetailsDropdownContent change now detects a click anywhere outside of the actual dropdown content, which then sets the expanded boolean to false, but clicking the button toggles the boolean, resulting in it being set to true again.

app/components/DetailsDropdownContent.tsx
```diff
@@ -6,14 +6,25 @@
...
 export default function DetailsDropdownContent({
   children,
+  onClose,
   ...props
-}: Parameters<typeof Card>[0]) {
+}: Parameters<typeof Card>[0] & { onClose?: () => void }) {
+  const contentRef = useRef<HTMLDivElement>(null)
+
+  useOnClickOutside(contentRef, () => {
+    if (onClose) onClose()
```
lpsinger, line 20:
```suggestion
    onClose?.()
```

lpsinger (changes requested):

tylerbarna:
Having some difficulty coming up with an adequately general way of detecting when a user is clicking an element we don't want to trigger the onClose with (eg detecting when the buttons in the buttongroup associated with the popup are clicked). Attempted to use both refs and classnames for this purpose with a check inside DetailsDropdownContent for one of those properties on the event.target, but there's variability in what the identified event.target is depending on where on the segmented button the user clicks

tylerbarna:
this is not the most eloquent method, but the general idea is present. May be slightly better to use a ref rather than a classname, also some expression that checks all parents/grandparents/etc in the DetailsDropdownContent would be a bit better

lpsinger:
Did you know that `useOnClickOutside` accepts a list of refs?

```ts
useOnClickOutside([dropdownContentRef, dropdownButtonRef], () => {...})`
```

tylerbarna:
> Did you know that `useOnClickOutside` accepts a list of refs?
> 
> ```ts
> useOnClickOutside([dropdownContentRef, dropdownButtonRef], () => {...})
> ```

Looking back at our discussion on Monday, did you intend to have the useOnClickOutside defined outside of the DetailsDropdownContent component/have the refs defined in the route.tsx file instead of the DateSelectorMenu and SortSelectorMenu component files?

lpsinger:
Yes, that would be the simplest.

app/routes/circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -51,63 +51,69 @@ function SortButton({
...
+  setShowContent: React.Dispatch<React.SetStateAction<boolean>>
+}
 
-  const submit = useSubmit()
+export const SortSelector = forwardRef<HTMLDivElement, SortSelectorProps>(
+  function SortSelector(
+    { form, defaultValue, showContent, setShowContent },
+    ref: React.Ref<HTMLDivElement>
+  ) {
+    // const [showContent, setShowContent] = useState(false)
```
Courey, line 66: did you mean to delete this?

tylerbarna: yes i did

app/routes/circulars._archive._index/DateSelectorMenu.tsx
```diff
@@ -68,156 +68,168 @@ function DateSelectorButton({
...
+              <Grid col={4} key="radio-alltime">
                 <Radio
                   form=""
-                  id={`radio-${value}`}
+                  id="radio-alltime"
                   name="radio-date"
-                  value={value}
-                  label={label}
-                  defaultChecked={value === defaultStartDate}
+                  value=""
```
Courey, line 154: is this intentional?

Courey: oh I get it, nm.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -170,6 +171,18 @@ export default function () {
...
+  const [showDateSelector, setShowDateSelector] = useState(false)
+  const SortSelectorRef = useRef<HTMLDivElement>(null)
+  const [showSortSelector, setShowSortSelector] = useState(false)
+
+  useOnClickOutside(dateSelectorRef, () => {
+    setShowDateSelector(false)
+  })
+  useOnClickOutside(SortSelectorRef, () => {
+    setShowSortSelector(false)
+  })
```
lpsinger, line 184: These refs and states are only used inside `SortSelector` and `DateSelector`. Please move them inside those two respective components. They need not be exposed as props.

lpsinger (changes requested):

app/routes/circulars._archive._index/DateSelectorMenu.tsx
```diff
@@ -77,7 +78,11 @@ export function DateSelector({
   defaultStartDate?: string
   defaultEndDate?: string
 }) {
-  const [showContent, setShowContent] = useState(false)
+  const dateSelectorRef = useRef<HTMLDivElement>(null)
+  const [showDateSelector, setShowDateSelector] = useState(false)
```
lpsinger, line 82: Renaming of the state from `showContent` to `showDateSelector` makes this a larger diff than necessary. Please keep the old name.

lpsinger (changes requested):

app/routes/circulars._archive._index/DateSelectorMenu.tsx
```diff
@@ -77,7 +78,11 @@ export function DateSelector({
   defaultStartDate?: string
   defaultEndDate?: string
 }) {
+  const dateSelectorRef = useRef<HTMLDivElement>(null)
```
lpsinger, line 81: This is the only ref, so we can just call this variable `ref` without introducing ambiguity.

```suggestion
  const ref = useRef<HTMLDivElement>(null)
```

app/routes/circulars._archive._index/SortSelectorButton.tsx
```diff
@@ -58,9 +59,12 @@ export function SortSelector({
   form?: string
   defaultValue?: string
 }) {
-  const [showContent, setShowContent] = useState(false)
-
   const submit = useSubmit()
+  const SortSelectorRef = useRef<HTMLDivElement>(null)
```
lpsinger, line 63:
```suggestion
  const ref = useRef<HTMLDivElement>(null)
```

lpsinger (approved):

PR #1998, opened 2024-02-28
Forced refresh method

### LINKED ISSUE #1996: Bug: Latest circular version does not display after edit, cached version is returned
### Description

When a moderator makes an update to a circular the browser is still serving the cached version, rather than the latest. 

### Steps to Reproduce

1. As a moderator, sign in to dev.gcn.nasa.gov
2. Go to the circulars page
3. Select any circular
4. Click the request edit button, make a small change, and submit the change **IMPORTANT**: Make a note of your change for later steps
5. Go back to /circulars
6. Click the review change requests button
7. Review and approve your change
8. Navigate to the circular you changed, the change is missing

**Expected behavior:** The change should be present immediately after saving the changes

**Actual behavior:** No change is present in browser, unless the user clears their cache and reloads

dakota002:
Resolves #1996 

Still testing, the removal of the caching headers may be unnecessary

### DIFF
--- a/app/lib/headers.server.ts
+++ b/app/lib/headers.server.ts
@@ -21,13 +21,6 @@ export const publicStaticShortTermCacheControlHeaders = {
 }
 
 /**
- * HTTP headers for 0 life.
- */
-export const publicStaticZeroLifeCacheControlHeaders = {
-  'Cache-Control': 'no-cache, no-store, must-revalidate',
-}
-
-/**
  * Get HTTP headers for declaring the canonical URL to search engines.
  *
  * This adds HTTP headers that are equivalent to the HTML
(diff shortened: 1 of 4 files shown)


lpsinger (changes requested):
That's not what `reloadDocument` does. It tells React Router to bypass client side routing. Probably all you need to do is remove the cache control headers. But don't remove the canonical URL headers!

dakota002:
> That's not what `reloadDocument` does. It tells React Router to bypass client side routing. Probably all you need to do is remove the cache control headers. But don't remove the canonical URL headers!

I'm currently trying that but it continues to load the old version. Wouldn't we want to avoid the client side routing in this case? My thought being that we need to make the request to the server to get the new data

lpsinger:
> I'm currently trying that but it continues to load the old version. Wouldn't we want to avoid the client side routing in this case? My thought being that we need to make the request to the server to get the new data

No. Try this: https://remix.run/docs/en/main/route/should-revalidate

dakota002:
This now works as expected with that function added

lpsinger (approved):

PR #2001, opened 2024-02-29
Removes the feature flag for versions

dakota002:
Draft until  #1998 is merged


lpsinger (approved):

lpsinger:
This looks great! I've approved it but moved it to draft status so that we don't merge it prematurely.

jracusin (approved):

jracusin:
#1998 is merged

lpsinger:
> #1998 is merged

Yes, but we still should not merge this yet. For one, deployments are still broken. For another, we should do some final user testing.

PR #2014, opened 2024-03-01
Use ruff instead of black for formatting

### LINKED ISSUE #2013: Update VSCode default Python formatter to Ruff
Ruff is used by GitHub workflows and pre-commit to format code, but the default VSCode formatter is still set to Black. 

I'll put in a PR to fix...

jak574:
Switch VSCode to use Ruff for Python formatting instead of Black, to be consistent with the rest of the codebase.

Fixes #2013

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "[python]": {
-    "editor.defaultFormatter": "ms-python.black-formatter",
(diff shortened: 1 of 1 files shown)


lpsinger (approved):
I checked at https://docs.astral.sh/ruff/integrations/#vs-code-official that this is the first-party VS Code extension.

PR #2047, opened 2024-03-07
Fix for import breaking ci

dakota002:


Courey (approved):

lpsinger:
This was my fault --- I rebased https://github.com/trussworks/react-uswds/pull/2726 and pushed ad2ea234991c2bd3c631ffe00fc1e6942f08d190 to update our package-lock.json file. I should have made a PR. I'll revert that.

Meanwhile, could this be a regression in react-uswds? Do you know why this is no longer working?

dakota002:
Not sure. From what I can tell the ButtonProps are not included as an export in their index.d.ts. I just deleted the import and did a ctrl+. to see where vs code would find it. I will see if they made a recent change that affects this

PR #2054, opened 2024-03-07
ACROSS API: Add norm_pdf function to remove scipy dependency.

### LINKED ISSUE #2053: Remove dependency on Scipy
python/across_api/base/fov.py imports scipy. I believe that we are only using scipy right now for its normal distribution PDF. Unfortunately, scipy is currently too large for our deployment methodology (though we are working on some mitigations for that).

Would you please remove the dependency on Scipy?

jak574:
# Description

SciPy is a large dependency, and the only function we currently use from it is the `stats.norm.pdf` function. This PR removes the usage of that function, and replaces it with an inline implementation of the same function. Fixes #2053.

# Testing

Did the following test to find that the output values of both functions are the same:
```python
x = np.arange(-5,5,0.1) * u.deg
std = 2.234 * u.deg
assert ((stats.norm(scale=std).pdf(x) - normal_pdf(x, std)) < 1e-17).all() == True
```

Also existing pytest tests.

### DIFF
--- a/python/across_api/base/fov.py
+++ b/python/across_api/base/fov.py
@@ -6,7 +6,6 @@ from typing import Optional, Union
 import astropy_healpix as ah  # type: ignore
 import astropy.units as u  # type: ignore
 from astropy.io.fits import FITS_rec  # type: ignore
-from scipy import stats  # type: ignore[import]
 import numpy as np
 from astropy.coordinates import SkyCoord  # type: ignore
 from astropy.time import Time  # type: ignore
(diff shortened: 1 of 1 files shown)


jak574:
Moved back to draft while I fix pytest issue.

lpsinger (changes requested):
I don't think that you need to do any explicit handling of Astropy units or quantities here; everything should just work as long as `std` and `x` have compatible unites, and fail otherwise.

jak574:
> Moved back to draft while I fix pytest issue.

OK, fixed, and confirmed to work.

python/across_api/base/fov.py
```diff
@@ -19,6 +18,36 @@
...
+    pdf
+        The probability density at the given x.
+    """
+
+    # Convert the standard deviation to degrees
+    std = std.to_value(u.deg)
+    x = x.to_value(u.deg)
+
+    # Calculate the exponent term
+    exponent = -(x**2) / (2 * std**2)
```
lpsinger, line 43:
```suggestion
    exponent = -0.5 * np.square(x / std)
```

lpsinger (changes requested):

jak574:
> I don't think that you need to do any explicit handling of Astropy units or quantities here; everything should just work as long as `std` and `x` have compatible unites, and fail otherwise.

Because `std` ends in a expononent, it needs to be a float otherwise the following error gets raised:

```python
UnitTypeError: Can only apply 'exp' function to dimensionless quantities
```

lpsinger:
> > I don't think that you need to do any explicit handling of Astropy units or quantities here; everything should just work as long as `std` and `x` have compatible unites, and fail otherwise.
> 
> Because `std` ends up in a exponent, it needs to be a float otherwise the following error gets raised:
> 
> ```python
> UnitTypeError: Can only apply 'exp' function to dimensionless quantities
> ```

That's a sign that you are passing it values of `std` and `x` that are not both of the same physical units. Notice that this works:

```
In [4]: np.exp((1 * u.deg) / (2 * u.deg))
Out[4]: <Quantity 1.64872127>
```

jak574:
> That's a sign that you are passing it values of `std` and `x` that are not both of the same physical units. Notice that this works:
> 
> ```
> In [4]: np.exp((1 * u.deg) / (2 * u.deg))
> Out[4]: <Quantity 1.64872127>
> ```

Right, I figured out what was going on and removed those lines.

lpsinger (approved):

PR #2064, opened 2024-03-08
Burst cube ToO group

dakota002:


app/routes/_gcn.user.credentials/client_credentials.server.ts
```diff
@@ -48,7 +48,11 @@ export class ClientCredentialVendingMachine {
     if (!user) throw new Response('not signed in', { status: 403 })
     return new this(
       user.sub,
-      user.groups.filter((x) => x.startsWith('gcn.nasa.gov/kafka'))
+      user.groups.filter(
+        (x) =>
+          x.startsWith('gcn.nasa.gov/kafka') ||
+          x == 'gcn.nasa.gov/burstcube-too-submitter'
+      )
```
lpsinger, line 55: Would there be any security or correctness implications of expanding this to:

```suggestion
      user.groups.filter((x) => x.startsWith('gcn.nasa.gov/'))
```

dakota002: This would make the moderator and submitter groups available as well

lpsinger: Ah, right. Good point.

Issue #2005, opened 2024-03-01
Python code should make AWS API calls asynchronously


lpsinger:
> * [ ]  I had thought that we had agreed to make AWS API calls asynchronously using aioboto3 rather than synchronously using boto3.

I'm using `architect-functions` to access AWS, not boto3. This is what I used for the TLE database, so I'm using it here to be consistent.

I'm down with using aioboto3, but it makes the code inconsistent and I don't want to make changes to code that isn't covered by this PR unless absolutely necessary.

I'll do a PR to change over to aioboto3 after this PR is merged.

_Originally posted by @jak574 in https://github.com/nasa-gcn/gcn.nasa.gov/issues/1975#issuecomment-1971902826_

PR #2073, opened 2024-03-11
ACROSS API: Move to using aioboto3 and make things as async as possible

jak574:
### Description

In response to Issue #2005, this moves the ACROSS API to use `aioboto3`, therefore making DynamoDB calls asynchronously. In addition it updates various components of the ACROSS API to be as async as possible. This involves solving several issues, caused by the fact that async functions can only be called by other async functions.

1. We currently use the `architect-functions` module to handle database connections. This does not support aioboto3, so we use aioboto3 direct, however we want to be able to run in local and testing modes. Therefore in order to support this, we use several internal calls from `architect-functions` to determine the table name with any prefixes, and also if we are running in a testing mode or on AWS.
2. Currently many modules can run automatically from instantiation if the correct variables are provide, where others cannot. We are moving all `get()`, `del()`, `put()` and `post()` functions to async, which cannot be called from `__init__()`, so therefore we removed this automatic running in all cases.
3. The `BurstCubeEphem` and `SwiftEphem` classes are slightly rewritten to have custom `get()` functions rather than `__init__()` in order to support async.
4. `pytest` tests updated to support `async` using the `pytest-asyncio` module.

In addition we rewrote the `TLE` class to use async fetch from spacetrack (supported by async support in the `spacetrack` module), and also fetch `async` from the ANTARES API in `Resolve`. Generally an effort has been made to make API endpoints async by default. 

### Reviewers
Leo Singer 
Samuel Wyatt

### Functional Requirements
* Use `aioboto3` library instead of `architect-functions` to perform database actions.
* Support the architect provided dev environment.

### Non-functional Requirements
* Move as many calls to be async as possible.

### Acceptance Criteria
* Pass code review.
* Pass testing.

### DIFF
--- a/python/across_api/across/hello.py
+++ b/python/across_api/across/hello.py
@@ -33,10 +33,8 @@ class Hello(ACROSSAPIBase):
             Your name
         """
         self.name = name
-        if self.validate_get():
-            self.get()
 
-    def get(self) -> bool:
+    async def get(self) -> bool:
         """
         GET method for ACROSS API Hello class.
 
--- a/python/across_api/base/database.py
+++ b/python/across_api/base/database.py
@@ -1,0 +1,16 @@
+import os
+
+import aioboto3  # type: ignore[import]
+from arc._lib import get_ports, use_aws  # type: ignore[import]
+
+
+def dynamodb_resource():
+    dynamodb_session = aioboto3.Session()
+    if use_aws():
+        return dynamodb_session.resource("dynamodb")
+    else:
+        port = get_ports()["tables"]
+        region_name = os.environ.get("AWS_REGION", "us-east-1")
+        return dynamodb_session.resource(
+            "dynamodb", endpoint_url=f"http://localhost:{port}", region_name=region_name
+        )
--- a/python/across_api/across/api.py
+++ b/python/across_api/across/api.py
@@ -6,8 +6,8 @@ from typing import Annotated, Optional
 
 from fastapi import Depends, Query, Security
 
-from ..base.api import app
 from ..auth.api import scope_authorize
+from ..base.api import app
 from .hello import Hello
 from .resolve import Resolve
 from .schema import HelloSchema, ResolveSchema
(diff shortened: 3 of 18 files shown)


python/across_api/across/hello.py
```diff
@@ -33,10 +33,8 @@ def __init__(self, name: Optional[str] = None):
             Your name
         """
         self.name = name
-        if self.validate_get():
-            self.get()
 
-    def get(self) -> bool:
+    async def get(self) -> bool:
```
lpsinger, line 37: Functions that don't contain any `await` statements need not be `async`.

python/across_api/base/database.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def dynamodb_resource():
+    dynamodb_session = aioboto3.Session()
+    if use_aws():
+        return dynamodb_session.resource("dynamodb")
+    else:
+        port = get_ports()["tables"]
+        return dynamodb_session.resource(
+            "dynamodb", endpoint_url=f"http://localhost:{port}", region_name="us-east-1"
```
lpsinger, line 12: There ought to be a way to get the region programmatically and to not hard-code it.

lpsinger (changes requested):

python/across_api/base/database.py
jak574: It doesn't really matter for this because it's accessing the DynamoDB instance running locally here. However, I changed this to fetch from the `AWS_REGION_NAME` environment variable or default to `us-east-1`, which is how `architect-functions` does it (except they default to `us-west-1`).

lpsinger (approved):
I skimmed this and it looks fine. @swyatt7 should do a more careful review than I.

python/across_api/across/hello.py
jak574: FastAPI documentation says it's better to use async on simple compute functions.

https://fastapi.tiangolo.com/async/#path-operation-functions

lpsinger: How very curious. Thanks for explaining!

swyatt7 (approved):
Pretty straightforward. lgtm

PR #2085, opened 2024-03-13
Make strip-python.py --test option respect --exclude

lpsinger:
The `--test` option for `--strip-python.py` will now ignore excluded directories.

It is currently not safe to strip tests from astropy because `astropy/__init__.py` imports `astropy/tests/__init__.py`.


jak574 (approved):
Tested in dev and works!

PR #2094, opened 2024-03-14
Fixes an error being thrown when the format is not defined on a correction

### LINKED ISSUE #2092: Moderation review gives unexpected error
We received our first Circulars revision request.  But when I click on "Review", I get "Unexpected Error".  I tried logging out and back in, and the behavior remains.

dakota002:
Resolves #2092

### DIFF
--- a/app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
+++ b/app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
@@ -20,6 +20,7 @@ import {
   moderatorGroup,
 } from './_gcn.circulars/circulars.server'
 import { getFormDataString } from '~/lib/utils'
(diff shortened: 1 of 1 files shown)


app/routes/_gcn.circulars.moderation.$circularId.$requestor.tsx
```diff
@@ -79,11 +80,16 @@ export default function () {
         oldString={circular.subject}
         newString={correction.subject}
       />
-      <h3>Format</h3>
-      <DiffedContent
-        oldString={circular.format ?? 'text/plain'}
-        newString={correction.format}
-      />
+
```
lpsinger, line 83:
```suggestion
```

lpsinger (changes requested):

PR #2100, opened 2024-03-14
Disable Submit/Update button when the content is unchanged

dakota002:


lpsinger (changes requested):
This should also be conditional on the format having changed.

PR #2102, opened 2024-03-15
ACROSS API: Make blocking IO async and fix missing awaits

jak574:
# Description
This PR covers 2 main changes.

1. There are some missing awaits in the `tle.py` code, this adds them back in. This code would never get run in the current configuration, but is a bug, which this fixes.
2. Makes 2 examples of blocking IO run asynchronously.

For item 2, the two blocking IO cases are the reading in of a FITS HEALPix file, and the using `SkyCoord.from_name` for name resolution. As `astropy` does not have support for `async`, we use the lightweight [Asyncer](https://asyncer.tiangolo.com) library to allow these to run asynchronously. 

Using Asyncer's `syncify` function, you can place blocking IO into its own worker thread, so it will not block the event loop. To the programmer this just turns a non-async function into an async coroutine that can be awaited.

### DIFF
--- a/python/across_api/across/resolve.py
+++ b/python/across_api/across/resolve.py
@@ -8,6 +8,7 @@ from typing import Optional, Tuple
 import httpx
 from astropy.coordinates.name_resolve import NameResolveError  # type: ignore
 from astropy.coordinates.sky_coordinate import SkyCoord  # type: ignore
+from asyncer import asyncify
 
 from ..base.common import ACROSSAPIBase
 from .schema import ResolveGetSchema, ResolveSchema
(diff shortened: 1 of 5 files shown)


lpsinger (commented):
As far as I can tell, this does the same thing as [`asyncio.to_thread`](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread) from the Python standard library.

jak574:
> As far as I can tell, this does the same thing as [`asyncio.to_thread`](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread) from the Python standard library.

Right! Asyncer's `asyncify` is essentially syntactic sugar that makes doing this extremely simple. The only difference is that it calls uses AnyIO, which FastAPI's async is based on.  

So you can use the synchronous command `fits.open(file)` asynchronously by doing `asyncify(fits.open)(file)`. There are additional benefits over the simplicity, also it works better with mypy.

lpsinger (approved):
This looks fine to me, but @swyatt7 should also review.

swyatt7 (approved):
lgtm

PR #2104, opened 2024-03-15
Fix code coverage in Python

jak574:
Upload to codecov.io in the python workflow is currently broken, this fixes it.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -32,8 +32,11 @@ jobs:
       - name: Pip Install
         run: pip install -r requirements.txt
 
-      - name: Pytest
-        run: pytest python
+      - name: Install Package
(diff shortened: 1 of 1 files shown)


swyatt7 (approved):

lpsinger:
This patch isn't correct. This package does not define any dependencies, and definitely defines no setuptools extras, so the `-e .[test]` does nothing.

jak574:
> This patch isn't correct. This package does not define any dependencies, and definitely defines no setuptools extras, so the `-e .[test]` does nothing.

FYI I just copied it from here https://github.com/nasa-gcn/python-dynamodb-autoincrement/blob/cf860e359f50f4273bdea21d772314d08f71d2e8/.github/workflows/pull_request.yml#L35

jak574:
Should have realized the [test] referred to something in pyproject.toml, sorry!

PR #2120, opened 2024-03-18
Adjust button text for tense and capitalization

lpsinger:
Change `Sorted By` to `Sort by` to agree with the tense and capitalization of `Filter by`.


tylerbarna (approved):
Good point, also nice because it reduces the horizontal width of the search components ever so slightly

PR #2135, opened 2024-03-21
Add tooltips for cross references

### LINKED ISSUE #1991: Add tooltips for Astro Flavored Markdown annotations for GCN Circulars
Acceptance criteria:
add tooltips that display the author, date, and title of astroflavored markdown enriched links for arxiv, doi, and gcn circulars. 
- all the data we need to query on the client side, not the server side. it would be too complicated to do this in loaders.
- there is a relatively new feature in react called suspense. you can have react components that depend on the result of awaiting something. this should be used if possible.
- add caching on client side.
- tooltips size is not initially known on the server side, so a patch in uswds is necessary to watch for resize events. 

EDIT: this turns out to be too big a problem to tackle all of the data sources at once, so this issue ONLY covers tooltips for GCN Circulars references.

lpsinger:
Fixes #1991.

### DIFF
--- a/app/routes/circulars.$circularId.($version)/AstroData.components.tsx
+++ b/app/routes/circulars.$circularId.($version)/AstroData.components.tsx
@@ -5,20 +5,73 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import { AstroDataLink } from './AstroDataContext'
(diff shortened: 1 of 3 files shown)


lpsinger:
Note: merge #2140 first.

dakota002 (approved):

PR #2176, opened 2024-04-03
Bulk delete

### LINKED ISSUE #2113: Moderation Bulk Delete
Script to bulk delete spam moderation requests

dakota002:
Resolves #2113

None selected:

image

Some selected:
image

### DIFF
--- a/app/routes/circulars.moderation._index.tsx
+++ b/app/routes/circulars.moderation._index.tsx
@@ -6,13 +6,21 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import type { SEOHandle } from '@nasa-gcn/remix-seo'
-import type { LoaderFunctionArgs } from '@remix-run/node'
-import { Link, useLoaderData } from '@remix-run/react'
-import { ButtonGroup, Grid } from '@trussworks/react-uswds'
+import type { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node'
+import { Form, Link, useLoaderData } from '@remix-run/react'
+import { Button, ButtonGroup, Checkbox, Grid } from '@trussworks/react-uswds'
+import { useState } from 'react'
 
 import { getUser } from './_auth/user.server'
-import type { CircularChangeRequest } from './circulars/circulars.lib'
-import { getChangeRequests, moderatorGroup } from './circulars/circulars.server'
+import type {
+  CircularChangeRequest,
+  CircularChangeRequestKeys,
+} from './circulars/circulars.lib'
+import {
+  bulkDeleteChangeRequests,
+  getChangeRequests,
+  moderatorGroup,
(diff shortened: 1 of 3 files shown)


app/routes/circulars.moderation._index.tsx
```diff
@@ -31,45 +39,100 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
-          />
-        ))}
-      </SegmentedCards>
+      <Form method="POST">
+        <ButtonGroup>
+          <Link to="/circulars" className="usa-button usa-button--outline">
+            Back
+          </Link>
+          <Button type="submit" secondary disabled={disableSubmit}>
+            Bulk Delete
```
lpsinger, line 89:
```suggestion
            Delete Selected
```

app/routes/circulars.moderation._index.tsx
```diff
@@ -31,45 +39,100 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
+    if (checked) {
+      setSelectedList([...selectedList, id])
+    } else {
+      setSelectedList(selectedList.filter((x) => x !== id))
+    }
+  }
+
+  function setBool() {
+    setDisableSubmit(selectedList.length === 0)
+  }
```
lpsinger, line 76: You can probably achieve a simpler state update if you just store the _number_ of selected checkboxes: it's initially zero; you increment it whenever a checkbox goes from unchecked to checked; you decrement it whenever a checkbox goes from checked to unchecked.

lpsinger (changes requested):

app/routes/circulars.moderation._index.tsx
dakota002: Good call, that worked much better and cleaned up the code a bit too!

app/routes/circulars.moderation._index.tsx
```diff
@@ -60,23 +60,16 @@ export async function action({ request }: ActionFunctionArgs) {
...
 
-  function checkboxOnChange(id: string, checked: boolean) {
+  function checkboxOnChange(checked: boolean) {
     if (checked) {
-      setSelectedList([...selectedList, id])
+      setSelectedCount(selectedCount + 1)
     } else {
-      setSelectedList(selectedList.filter((x) => x !== id))
+      setSelectedCount(selectedCount - 1)
     }
```
lpsinger, line 70: I don't know if Typescript will allow, but due to JavaScript's rules for type promotion, you could do

```ts
setSelectedCount(selectedCount + checked)
```

Elegantly concise? Unholy terror from beyond? You decide.

Also, is there any benefit to using the callback version of setState here? I confess that I do not totally understand when you need that and when you don't.

```ts
setSelectedCount((old) => old + 1)
```

lpsinger (approved):

PR #2177, opened 2024-04-03
Adds playwright, beginning tests

dakota002:
I trimmed this down a bit from what I was working on, adding auth states for various tests is more complex than we need to start.

Adds Playwright and a super simple example based on the demo content


.gitignore
```diff
@@ -23,3 +23,13 @@ sam.yaml
 
 .env
 .eslintcache
+
+# Generated by Skyfield
+de421.bsp
```
lpsinger, line 28:
```suggestion
# Generated by Playwright
```

That de421.bsp file is a Solar System ephemeris file. It would have been left over from the ACROSS Python code.

lpsinger (changes requested):
Looks remarkably clean and simple! Only one minor suggestion for the gitignore file.

@Courey, have you used Playwright before? Does this look reasonable?

lpsinger:
Oh, right, the Playwright tests are failing in the CI pipeline. Have you investigated that, @dakota002?

dakota002:
> Oh, right, the Playwright tests are failing in the CI pipeline. Have you investigated that, @dakota002?

Yup! Working on it at the moment, has to do with hitting the endpoint while it spins up the test server. I need to do a bit more investigating on how it runs though github actions

lpsinger (changes requested):
Oops, I approved but the tests are still failing.

lpsinger (approved):

lpsinger:
We ought to collect coverage data and upload to codecov, but we can do that in a separate PR.

Courey (approved):

lpsinger:
Is there an issue that this closes?

PR #2188, opened 2024-04-10
Adds zendesk request function, post request when correction is requested

### LINKED ISSUE #2114: Integrate revision requests into Zendesk workflow
we would like to track edit requests via zendesk.

acceptance criteria:
- integrate revision requests with zendesk
- a user should be able to submit a revision request and we receive a request in zendesk.
- CC the original submitter

dakota002:
Resolves #2114

### DIFF
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -1,0 +1,48 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { getEnvOrDie } from './env.server'
+import { getBasicAuthHeaders } from './headers.server'
+
+interface ZendeskRequest {
+  requester: Requester
+  subject: string
+  comment: RequestComment
+}
+
+interface Requester {
+  email: string
+  name: string
+}
+
+interface RequestComment {
+  body: string
+}
+
+export async function postToZendesk(request: ZendeskRequest) {
+  const response = await fetch(
+    'https://nasa-gcn.zendesk.com/api/v2/requests.json',
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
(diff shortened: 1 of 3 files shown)


dakota002:
@lpsinger should I include a refactor of the contact form zendesk request in this PR?

lpsinger (commented):
Is there an existing TypeScript library for the Zendesk API?

dakota002:
Their website lists 2 nodejs packages. Neither are officially theirs, and one is not maintained anymore. 

This is the one that is currently maintained: https://github.com/blakmatrix/node-zendesk 

Might be worth investigating, but with our current usage it seems a bit bulky. It could be a worthwhile addition in the future though if we come up with some other admin features that want to make use of/interact with  it

app/lib/zendesk.server.ts
```diff
@@ -0,0 +1,48 @@
...
+interface Requester {
+  email: string
+  name: string
+}
+
+interface RequestComment {
+  body: string
+}
+
+export async function postToZendesk(request: ZendeskRequest) {
```
lpsinger, line 26: Suggest more descriptive function name:

```suggestion
export async function postZendeskRequest(request: ZendeskRequest) {
```

app/lib/zendesk.server.ts
```diff
@@ -0,0 +1,48 @@
...
+        request,
+      }),
+    }
+  )
+
+  if (!response.ok) {
+    console.error(response)
+    throw new Error(`Reqeust failed with status ${response.status}`)
+  }
+}
```
lpsinger, line 48: This code is now duplicated in contact.tsx. Please refactor to eliminate multiple copies.

lpsinger (changes requested):
There is now d

lpsinger (changes requested):
There is now duplicated code in contact.tsx. Please refactor to remove duplicated code.

dakota002:
> There is now duplicated code in contact.tsx. Please refactor to remove duplicated code.

I addressed it in the most recent commit

lpsinger (approved):

PR #2197, opened 2024-04-15
Fetch tooltip data on server side

### LINKED ISSUE #2170: Move tooltip REST API calls to backend so that CDN caches them
# Description

In #2135, the browser makes API calls to third-party data sources.

# Acceptance criteria

- Move the API calls to the backend and add cache control settings that permits CloudFront to cache the responses for about a day.
- Update the front end so that it queries our REST API rather than the upstream data source.

lpsinger:
Fixes #2170.

### Testing
Tested interactively in local sandbox.


dakota002 (approved):
Looks great, tested locally too

PR #2201, opened 2024-04-15
Responsive hamburger and navbar menus

### LINKED ISSUE #2037: bug: hamburger menu and navbar menus do not automatically close when clicking away or changing responsive layout
if you have the user nav menu open, and you're in the desktop display, you resize the window to mobile size, which causes it to display the hamburger menu. when you resize the window again the user nav menu is still open. There are bugs with when you click away and or resize the window the menus do not behave as expected.

acceptance criteria:
- research other responsive sites that hav nav bar sub menus and hamburger menus for mobile.
- determine correct behavior
- create an issue ticket documenting the solution

tylerbarna:
# Description
This PR fixes the issue with the rendering of the hamburger menu when switching between mobile and desktop widths for the window by adding an event listener to check the width of the window.

As of 04981c32, the behaviour is such that opening the hamburger menu and resizing the window to be wider than 1025 pixels will cause the hamburger menu and overlay to disappear. Depending on feedback, it may be preferable to make it so going from <1025 pixel window to a >1025 pixel window sets the states for the hamburger menu to false by default, which can be accomplished by adding `if (window.innerWidth >= 1025) hideMobileNav()` to `updateMobile` or something similar.

The hamburger menu and user dropdown have also been altered to use useOnClickOutside to have them be hidden when a user clicks elsewhere on the page.

# Related Issue(s)
Resolves #2037

# Testing
Tested by setting the window width to below 1025 pixels (eg, mobile), opening the hamburger menu, and then resizing the window to be wider than 1025 pixels. The overlay element and hamburger menu should disappear. The click away aspect can be tested by expanding the user menu and clicking elsewhere on the page, or expanding the hamburger menu and clicking outside of the hamburger menu.

### DIFF
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -13,6 +13,7 @@ import {
   Title,
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
+import classNames from 'classnames'
 import { useState } from 'react'
 
 import { Meatball } from '~/components/meatball/Meatball'
@@ -86,9 +87,10 @@ export function Header() {
 
   return (
     <>
-      {expanded && (
-        <div className="usa-overlay is-visible" onClick={hideMobileNav} />
-      )}
+      <div
+        className={classNames('usa-overlay', { 'is-visible': expanded })}
+        onClick={hideMobileNav}
+      />
       <USWDSHeader basic className={`usa-header--dark ${styles.header}`}>
         <div className="usa-nav-container">
           <div className="usa-navbar">
@@ -102,6 +104,7 @@ export function Header() {
           </div>
           <PrimaryNav
             mobileExpanded={expanded}
+            onToggleMobileNav={toggleMobileNav}
             items={[
               <NavLink
                 className="usa-nav__link"
@@ -211,7 +214,6 @@ export function Header() {
                 </Link>
               ),
             ]}
-            onToggleMobileNav={toggleMobileNav}
           />
         </div>
       </USWDSHeader>
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -14,7 +14,7 @@ import {
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
-import { useEffect, useLayoutEffect, useRef, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useOnClickOutside } from 'usehooks-ts'
 
 import { Meatball } from '~/components/meatball/Meatball'
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -14,7 +14,7 @@ import {
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
-import { useLayoutEffect, useRef, useState } from 'react'
+import { useEffect, useLayoutEffect, useRef, useState } from 'react'
 import { useOnClickOutside } from 'usehooks-ts'
 
 import { Meatball } from '~/components/meatball/Meatball'
@@ -79,7 +79,7 @@ export function Header() {
     typeof window !== 'undefined' ? window.innerWidth < 1025 : false
   )
 
-  useLayoutEffect(() => {
+  useEffect(() => {
     if (typeof window !== 'undefined') {
       const updateMobile = () => {
         setMobile(window.innerWidth < 1025)
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -14,7 +14,7 @@ import {
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
-import { useEffect, useRef, useState } from 'react'
+import { useRef, useState } from 'react'
 import { useOnClickOutside, useWindowSize } from 'usehooks-ts'
 
 import { Meatball } from '~/components/meatball/Meatball'
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -234,7 +234,7 @@ export function Header() {
                   </>
                 ) : (
                   <Link
-                    className="usa-nav__link"
+                    className="usa-nav__link text-no-wrap"
                     to="/login"
                     key="/login"
                     onClick={hideMobileNav}
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -14,7 +14,8 @@ import {
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
 import classNames from 'classnames'
-import { useState } from 'react'
+import { useRef, useState } from 'react'
+import { useOnClickOutside } from 'usehooks-ts'
 
 import { Meatball } from '~/components/meatball/Meatball'
 import { useEmail, useUserIdp } from '~/root'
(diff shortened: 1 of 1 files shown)


tylerbarna:
Outstanding issue as of 04981c32 : current implementation results in a warning about useLayoutEffect and differing behaviour on server vs client. This is because window is not defined on the server, so there are conditionals to prevent the window references from causing an error.
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

tylerbarna:
not sure why it's throwing an error for `Test / node / build (20, ubuntu-latest)` , the file in question isn't changed by this PR
```
Run npx prettier -c .
Checking formatting...
[warn] app/routes/missions.einsteinprobe/route.mdx
[warn] Code style issues found in the above file. Run Prettier to fix.
Error: Process completed with exit code 1.
```

lpsinger:
> not sure why it's throwing an error for `Test / node / build (20, ubuntu-latest)` , the file in question isn't changed by this PR
> 
> ```
> Run npx prettier -c .
> Checking formatting...
> [warn] app/routes/missions.einsteinprobe/route.mdx
> [warn] Code style issues found in the above file. Run Prettier to fix.
> Error: Process completed with exit code 1.
> ```

https://github.com/nasa-gcn/gcn.nasa.gov/pull/2203

tylerbarna:
> > not sure why it's throwing an error for `Test / node / build (20, ubuntu-latest)` , the file in question isn't changed by this PR
> > ```
> > Run npx prettier -c .
> > Checking formatting...
> > [warn] app/routes/missions.einsteinprobe/route.mdx
> > [warn] Code style issues found in the above file. Run Prettier to fix.
> > Error: Process completed with exit code 1.
> > ```
> 
> #2203

will rebase once that's merged

app/root/header/Header.tsx
```diff
@@ -100,119 +128,123 @@ export function Header() {
...
-                        onClick={onClickUserMenuItem}
-                      >
-                        Sign Out
-                      </Link>,
-                    ]}
-                  />
-                </>
-              ) : (
-                <Link
+          <div ref={menuRef}>
```
dakota002, line 131: Can this ref be moved to the PrimaryNav component instead of wrapping it in a div? 

The div adds some extra styling that bunches up the sign in button:
image

dakota002: Actually this can just be fixed by adding the `.text-no-wrap` class to the button, you can ignore my previous comment

tylerbarna: Just updated to fix

dakota002:
There is a small bug still. If you open the user menu while signed in it will close and reopen when you click the button again, instead of closing. It is triggering the onClickOutside, setting the open state to false, then triggering the ontoggle setting it back to true. 

You could resolve this by adding an id attribute to the NavDropDownButton component, (ex: `id="userMenuLabel"`) and updating the hook to check that the clicked item is not a child of the menu:

```js
  useOnClickOutside(userMenuRef, (event) => {
    if (
      !document
        .querySelector('#userMenuLabel')
        ?.contains(event.target as Node) &&
      userMenuIsOpen
    )
      setUserMenuIsOpen(false)
  })
```

tylerbarna:
> There is a small bug still. If you open the user menu while signed in it will close and reopen when you click the button again, instead of closing. It is triggering the onClickOutside, setting the open state to false, then triggering the ontoggle setting it back to true.
> 
> You could resolve this by adding an id attribute to the NavDropDownButton component, (ex: `id="userMenuLabel"`) and updating the hook to check that the clicked item is not a child of the menu:
> 
> ```js
>   useOnClickOutside(userMenuRef, (event) => {
>     if (
>       !document
>         .querySelector('#userMenuLabel')
>         ?.contains(event.target as Node) &&
>       userMenuIsOpen
>     )
>       setUserMenuIsOpen(false)
>   })
> ```

oh, good catch, I'll fix that now

tylerbarna:
> There is a small bug still. If you open the user menu while signed in it will close and reopen when you click the button again, instead of closing. It is triggering the onClickOutside, setting the open state to false, then triggering the ontoggle setting it back to true.
> 
> You could resolve this by adding an id attribute to the NavDropDownButton component, (ex: `id="userMenuLabel"`) and updating the hook to check that the clicked item is not a child of the menu:
> 
> ```js
>   useOnClickOutside(userMenuRef, (event) => {
>     if (
>       !document
>         .querySelector('#userMenuLabel')
>         ?.contains(event.target as Node) &&
>       userMenuIsOpen
>     )
>       setUserMenuIsOpen(false)
>   })
> ```

@dakota002 would this also work for #1989? The implementation I currently have for that isn't the best

dakota002:
Yea, it will probably clean it up a bit so you won't have to chain parentElement stuff

app/root/header/Header.tsx
```diff
@@ -70,6 +72,36 @@ export function Header() {
...
+
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      const updateMobile = () => {
+        setMobile(window.innerWidth < 1025)
+      }
+      window.addEventListener('resize', updateMobile)
+      return () => window.removeEventListener('resize', updateMobile)
+    }
+  }, [])
```
lpsinger, line 90: Try https://usehooks-ts.com/react-hook/use-window-size

lpsinger (changes requested):
Please try the React hook `useWindowSize` from usehooks-ts.

dakota002 (approved):

Courey (approved):

app/root/header/Header.tsx
```diff
@@ -66,10 +68,33 @@ function NavDropDownButton({
...
   const [expanded, setExpanded] = useState(false)
   const [userMenuIsOpen, setUserMenuIsOpen] = useState(false)
+  const menuRef = useRef<HTMLDivElement>(null)
+  const userMenuRef = useRef<HTMLDivElement>(null)
+
+  const [mobile, setMobile] = useState(isMobileWidth ?? false)
+
+  useEffect(() => {
+    setMobile(isMobileWidth)
+  }, [isMobileWidth])
```
lpsinger, line 83: This is probably unnecessary. The point of `useWindowSize()` is so that you can have props that depend functionally rather than imperatively on the window size.

app/root/header/Header.tsx
```diff
@@ -66,10 +68,33 @@ function NavDropDownButton({
...
+    if (expanded) setExpanded(false)
+  })
+
+  useOnClickOutside(userMenuRef, (event) => {
+    if (
+      !document
+        .querySelector('#userMenuDropdown')
+        ?.contains(event.target as Node) &&
+      userMenuIsOpen
+    )
```
lpsinger, line 95: What is this supposed to do?

tylerbarna: this is so clicking the user menu dropdown while it's open collapses the menu; the onClickOutside would set the state to false before the onClick toggled the state otherwise

app/root/header/Header.tsx
tylerbarna: Yep, removed in cb419f1

PR #2210, opened 2024-04-19
Add tooltips for DOIs

### LINKED ISSUE #2171: Add tooltips for DOIs
# Description

Display tooltips when user hovers over DOIs. Some complications:

- We don't currently have a production DataCite or CrossRef API key. If we used either of those services as the data source, we would need to follow up with the agency's DataCite POC.
- If we use ADS as our data source, then we should create a separate ADS API key so that rate limiting due to the tooltips does not cause the weekly harvesting of Circulars bibcodes to break.

# Acceptance criteria

- Tooltips with title, author, date, and journal displayed for DOIs.

lpsinger:
Fixes #2171.

# Testing
Tested locally.


dakota002 (approved):

lpsinger:
Thanks. I'll deploy the new env variables before I merge this.

PR #2218, opened 2024-04-22
Fix correction requests for ID equals 0

dakota002:
# Description
@Vidushi-GitHub noticed a 400 error when submitting a correction for circular 0. This is caused by a check in the react component evaluating the ID of 0 as a false

# Testing
Added a circular with ID 0 and submitted a correction request locally, then approved.

### DIFF
--- a/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
@@ -170,7 +170,7 @@ export function CircularEditForm({
       <h1>{headerText} GCN Circular</h1>
       <Form method="POST" action={`/circulars${formSearchString}`}>
         <input type="hidden" name="intent" value={intent} />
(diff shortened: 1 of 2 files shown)


app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -181,7 +181,7 @@ export function CircularEditForm({
         )}
         <InputGroup className="border-0 maxw-full">
           <InputPrefix className="wide-input-prefix">
-            {circularId ? 'Editor' : 'From'}
+            {circularId !== undefined ? 'Editor' : 'From'}
```
lpsinger, line 184:
```suggestion
            {circularId === undefined ? 'From' : 'Editor'}
```

lpsinger (changes requested):

lpsinger (approved):

PR #2227, opened 2024-04-26
Navigate between circulars within specific circulars pages

### LINKED ISSUE #2226: Feature: In-page Navigation Between Circulars
# Description
Add ability to navigate between circulars without returning to the main circulars page

# Acceptance criteria

- [x] Buttons for navigating to previous and next circulars match react-uswds style
- [x] URL parameters are preserved when navigating
- [ ] Navigation handles Circulars with non-integer IDs
- [ ] On most recent and oldest Circular, change relevant button to show that pressing it will return users to main Circulars page
- [ ] (optional) Have navigation between Circulars iterate through user search results rather than all Circulars

tylerbarna:
# Description
Adds navigation buttons inside specific circular pages to navigate to previous and next circulars. If attempting to navigate to a circular that does not exist, users will be returned to the main circulars page.

# Related Issue(s)
Resolves #2226 

# Testing
Testing for basic navigation:
1. Navigate to a specific circular
2. Scroll to bottom of page
3. Click 'Next Circular' or 'Previous Circular' buttons
4. Page for next or previous circular should load

Testing for navigation to a nonexistent circular:
1. Navigate to the most recent circular
2. Click 'Next Circular'
3. Main Circulars page should load
4. Navigate to oldest circular
5. Click 'Previous Circular'
6. Main Circulars page should load

### DIFF
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -148,7 +148,7 @@ export default function () {
       <h1 className="margin-bottom-0">GCN Circular {circularId}</h1>
       <FrontMatter {...frontMatter} />
       <Body className="margin-y-2">{body}</Body>
-      <ButtonGroup className="display-flex flex-row flex-align-center flex-justify-center">
+      <ButtonGroup className="display-flex flex-justify-center">
         <Link to={`${previousCircularLinkString}`} className="usa-button">
           <div className="position-relative">
             <Icon.ArrowBack
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -73,6 +73,12 @@ export default function () {
   const linkString = `/circulars/${circularId}${
     !latest && version ? `/${version}` : ''
   }`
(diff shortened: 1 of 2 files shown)


tylerbarna:
[8033408](https://github.com/nasa-gcn/gcn.nasa.gov/pull/2227/commits/80334085322e894af5650cc86f39646135391ed4) does not handle instances of non-standard circular IDs, hence the draft status. Also does not indicate when user is navigating back to the main Circulars page when on the oldest or most recent Circular

lpsinger:
There are also occasions when Circular IDs increment by more than 1 --- when we have deleted duplicate Circulars, for example.

tylerbarna:
they really should add a confirmation for the close PR button...

tylerbarna:
I was thinking a bit more about this, and if this were to use some way of maintaining the search context as we had discussed, it would work when navigating from the root archive page to a specific circular, but in cases where there isn't a search context (ie when someone clicks on a direct link to a specific circular from outside GCN), it wouldn't work. It might be more robust to use the URL search parameters to initiate a search when a specific circular page is loaded, though I'm not sure how much overhead it would add to implement the feature.

PR #2228, opened 2024-04-26
Permit requests that come via the CDN, reject all others

lpsinger:
Allow only requests that come via our content delivery network (CDN), AWS CloudFront. For all other requests, return a 502 error. This prevents direct requests to our origin.

By intent or mistake, if an actor managed to make requests directly to our origin, it could cause unwanted cloud expenses due to unnecessary processing by our origin.

Our Lambdas now have an environment variable called `CDN_SECRET` which should be set to a long, random string. We will configure CloudFront to add the header `X-CDN-Secret` to all requests to our origin. If the environment variable is set and the header does not match, then we reject the request and return an HTTP 502 response.

The AWS CloudFormation documentation [suggests such custom headers as a method to identify requests that come from CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/add-origin-custom-headers.html#add-origin-custom-headers-use-cases). We chose this solution as a sensible balance of security and simplicity. Here are some other possible approaches:

- Instead of adding the check here in the Remix server entry point, we could put the check in a new Lambda that is configured as an [API Gateway authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html). This would have the advantage that the request would be blocked even before making it to Remix. However, it would have the disadvantage of adding latency due to the additional Lambda's startup time.

- We could have turned on AWS IAM authorization for our API Gateway and written a Lamdba@Edge function for CloudFormation to sign requests to our origin using the configured IAM role, [as described in this blog post](https://medium.com/@dario_26152/restrict-access-to-lambda-functionurl-to-cloudfront-using-aws-iam-988583834705). However this would also add some Lambda startup time, not to mention complexity.

### DIFF
--- a/app/routes/docs.contributing.configuration/route.mdx
+++ b/app/routes/docs.contributing.configuration/route.mdx
@@ -81,6 +81,19 @@ account for GCN production by registering the email address
       <Default>Hard-coded constant</Default>
     </tr>
     <tr>
+      <Key>`CDN_SECRET`</Key>
+      <Description>
+        Secret key that must be present in the `X-CDN-Secret` request header of
+        all requests to prove that the request came via the content delivery
+        network rather than directly to the origin. Should be a long, random
+        string
+      </Description>
+      <Default>
+        Requests are permitted regardless of the value of the `X-CDN-Secret`
+        header
+      </Default>
+    </tr>
+    <tr>
       <Key>`COGNITO_USER_POOL_ID`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`</Key>
       <Description>
         OpenID Connect identity provider configuration for [AWS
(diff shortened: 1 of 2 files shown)


lpsinger:
@dakota002, I added a `handleDataRequest` function, and that does work to block access to resource routes, but Remix is calling `handleDataRequest` _after_ it has already called the route's loader. That's not ideal because it does not prevent our origin from doing work.

dakota002:
That makes sense, this comment is from a previous version, https://github.com/remix-run/remix/issues/4505#issuecomment-1301978829 but I think the same is applied now, specifically:

> handleDataRequest is for modifying the response of a data request.

So we probably wouldn't want to use it for preventing access

lpsinger:
I removed the entry.server module and instead put this check directly in the AWS Lambda handler in server.ts. I think it works now.

dakota002 (approved):
Nice, tested locally and behaves as expected

PR #2238, opened 2024-04-29
Use native fetch instead of Node.js polyfill

lpsinger:
As of Remix 2.9.0, Remix supports using the native fetch() added in Node.js 20 instead of a polyfill.


lpsinger:
Blocked by #2253.

lpsinger:
Fixed by #2407.

PR #2245, opened 2024-04-29
Fix primary nav responsiveness bugs

### LINKED ISSUE #2037: bug: hamburger menu and navbar menus do not automatically close when clicking away or changing responsive layout

lpsinger:
- When switching from mobile to desktop view, make sure to hide the mobile nav.
- When clicking outside the user menu in desktop view, make sure to collapse the user menu.

Fixes #2037.

This is simpler than #2201, but I think that it catches all of the corner cases.

### DIFF
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -13,7 +13,8 @@ import {
   Title,
   Header as USWDSHeader,
 } from '@trussworks/react-uswds'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
+import { useClickAnyWhere, useWindowSize } from 'usehooks-ts'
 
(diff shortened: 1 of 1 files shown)


tylerbarna (approved):
This seems to be a better implementation than #2201

dakota002 (approved):

PR #2248, opened 2024-05-01
Kafka circulars

### LINKED ISSUE #686: Distribute Circulars via Kafka
**Description**

There's community interest in receiving circulars via Kafka.  They're already stored in JSON objects, so we should setup a Kafka topic for them.

**Acceptance Criteria**
- add Kafka topic: gcn.circulars
- produce to gcn.circulars from circulars json files
- add circulars to Quickstart under JSON option
- announce it and add news item

dakota002:
# Description
Adds a table-streams event for Circulars over Kafka

This is currently a draft as there is either an issue in a dependency (possibly in cpp-zst-node16-fork or bindings) or in how we are building out these table-streams functions that is causing errors. 

Currently investigating how cpp-zst-node16-fork uses the bindings package and why it works in some projects and not others

# Related Issue(s)
Resolves #686

# Testing
To get this to work in local dev setting there are a few things that need to be modified under node_modules:

- `@remix-run/dev/dist/devServer_unstable/index.js` -> line 139 sets `stdio: pipe`, change this to 
```
      stdin: process.stdin,
      stdout: process.stdout,
```
in order to use the plugin-lambda-invoker.

- Add the following lines to the `try` array defined on the `defaults` object in `bindings/bindings.js` (currently trying to resolve this):

```
      ['module_root', 'node_modules', 'cppzst-node-16-fork', 'build', 'Debug', 'bindings'],
      ['module_root', 'node_modules', 'cppzst-node-16-fork', 'build', 'Release', 'bindings'],
```

Once those are in place, test by:
- Start a basic kafka consumer pointed to the dev server
- press `i` to trigger the invoker
- select  @tables-streams circulars-kafka-distribution


lpsinger:
The build job is failing with C++ compilation errors when building cpp-zst-node16-fork. Based on the compilation errors, I suspect that fixing it might be as simple as adding `-std=c++11` to the compiler flags, though I don't know how to do that.

As for the binding path issues, the fundamental problem is that this Zstandard compressor/decompressor library contains a [native Node addon](https://nodejs.org/api/addons.html) which is our bundler (esbuild) does not know how to handle. I think it should be sufficient to copy the native addons to the appropriate build output directory in https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/esbuild.js.

However, given the installation challenges, it may be worthwhile checking if there is a pure JavaScript or a WASM implementation of Zstandard that we could use instead, which would be trivial to install.

dakota002:
I think I got it working. The esbuild solution was a much easier find than the previous issues I was encountering. I adapted this https://esbuild.github.io/content-types/#file based on a comment I came across somewhere.

package.json
```diff
@@ -45,6 +45,7 @@
     "diff": "^5.2.0",
     "downshift": "^7.2.1",
     "email-validator": "^2.0.4",
+    "gcn-kafka": "github:dakota002/gcn-kafka-js#962d48862622e7abcb0088cbb6ca1bc3ee24f4ce",
```
lpsinger, line 48:
```suggestion
    "gcn-kafka": "^0.2.1",
```

package.json
```diff
@@ -77,7 +78,7 @@
   },
   "devDependencies": {
     "@architect/architect": "^10.16.3",
-    "@architect/plugin-lambda-invoker": "^1.2.0",
+    "@architect/plugin-lambda-invoker": "github:dakota002/plugin-lambda-invoker#8149c366f33af1e82ca2aeeda828c92ffc86e35c",
```
lpsinger, line 81: Is there a PR for your changes to plugin-lambda-invoker?

lpsinger (changes requested):

package.json
dakota002: https://github.com/architect/plugin-lambda-invoker/pull/4 Yup, it may need some work, but I recently rebased it

dakota002:
Marked as ready too soon, need to fix something else first

esbuild.js
```diff
@@ -22,6 +22,7 @@ const options = {
   target: ['node18'],
   minify: !dev,
   sourcemap: dev,
+  loader: { '.node': 'file' },
```
lpsinger, line 25: Would you please check that the native modules are _not_ copied over into the build output for functions that do not use gcn-kafka --- that is to say, that they are tree-shaken?

lpsinger (changes requested):
Rather than add a new Lambda, I would rather that we use the existing Lambda in app/table-streams/circulars. Look at app/scheduled/circulars for ideas about how to structure it. I imagine something roughly like:

```ts
// app/table-streams/circulars/actions/search.ts
export async function searchAction(...) {
  ...
}

// app/table-streams/circulars/actions/kafka
export async function kafkaAction(...) {
}

// app/table-streams/circulars/index.ts
import { kafkaAction } from './actions/kafka'
import { searchAction } from './actions/search'
import { createTriggerHandler } from '~/lib/lambdaTrigger.server'

  async (record: DynamoDBRecord) => {
    Promise.all([kafkaAction, searchAction].map((action) => action(record)))
  }
)


```

esbuild.js
dakota002: It was getting copied into the build folder directly, I updated it in the latest commit and it no longer gets copied. Also just tested and successfully produced and consumed

app/lib/lambdaTrigger.server.ts
```diff
@@ -41,3 +44,9 @@ export async function allSettledOrRaise<T extends readonly unknown[]>(
     }
   ).map(({ value }) => value) as { -readonly [P in keyof T]: Awaited<T[P]> }
 }
+
+export function unmarshallTrigger(
+  item?: Record<string, LambdaTriggerAttributeValue>
+) {
+  return unmarshall(item as Record<string, AttributeValue>)
+}
```
lpsinger, line 52: This file contains utilities for _all_ kinds of Lambda trigger handlers --- not specifically DynamoDB. Please keep the DynamoDB Table Streams related functionality in the app/table-streams directory.

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,34 @@
...
+  await producer.send({
+    topic,
+    messages: [
+      {
+        key: 'message',
+        value: JSON.stringify(circular),
+      },
+    ],
+  })
+  await producer.disconnect()
```
lpsinger, line 33: Please close the Kafka connection only when the Lambda runtime is terminating. That way we can amortize the cost of opening the connection across multiple Lambda invocations, only paying it once per cold start.

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,34 @@
...
+}).producer()
+
+export async function sendKafka(circular: Circular, topic: string) {
+  await producer.connect()
+  await producer.send({
+    topic,
+    messages: [
+      {
+        key: 'message',
+        value: JSON.stringify(circular),
```
lpsinger, line 29: Are we sure that we have stripped out sensitive fields such as OIDC sub claims?

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,34 @@
...
+import { getEnvOrDie } from './env.server'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+const client_id = getEnvOrDie('KAFKA_CLIENT_ID')
+const client_secret = getEnvOrDie('KAFKA_CLIENT_SECRET')
+
+const producer = new Kafka({
+  client_id,
+  client_secret,
+  domain: 'dev.gcn.nasa.gov', // Intentionally set to dev for testing
```
lpsinger, line 19: Don't hard code this. You could determine this from existing environment variables using `getHost()` or you could introduce a new environment variable to configure it.

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { Kafka } from 'gcn-kafka'
+
+import { getEnvOrDie } from './env.server'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+const client_id = getEnvOrDie('KAFKA_CLIENT_ID')
+const client_secret = getEnvOrDie('KAFKA_CLIENT_SECRET')
```
lpsinger, line 14: Document these environment variables in our contributing guide.

lpsinger (changes requested):

app/lib/lambdaTrigger.server.ts
dakota002: Gotcha, this was a change from when I had multiple files and needed to reuse the function, I'll revert it in the next commit

app/lib/kafka.server.ts
dakota002: No, but I'll add a delete statement to the method to make sure it is removed

app/lib/kafka.server.ts
lpsinger: You probably _should_ call disconnect when the Lambda runtime is terminating gracefully. [For Python Lambdas, AWS sends a SIGTERM signal to the function.](https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html) The documentation doesn't say that it does that for Node.js Lamdbas, but it probably does.

dakota002: This might be a [useful example](https://github.com/aws-samples/graceful-shutdown-with-aws-lambda/blob/main/nodejs-demo/hello-world/app.js), I will see if I can implement something similar

dakota002: Actually I am not sure if this will work as we think. Without the disconnect, the function hangs and throws an error 
image

I'll do some more testing though to be sure

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { getEnvOrDie, hostname } from './env.server'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+const client_id = getEnvOrDie('KAFKA_CLIENT_ID')
+const client_secret = getEnvOrDie('KAFKA_CLIENT_SECRET')
+const domain = hostname as
+  | 'gcn.nasa.gov'
+  | 'test.gcn.nasa.gov'
+  | 'dev.gcn.nasa.gov'
```
lpsinger, line 18: This is not quite correct; `hostname` might `'localhost'`. We have logic to determine the appropriate value for `domain` in ClientSampleCode.tsx, but it's written as a hook and can only be used in a React component. Please refactor that and provide both a hook or a non-hook version.

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const producer = new Kafka({
+  client_id,
+  client_secret,
+  domain,
+}).producer()
+
+export async function sendKafka(circular: Circular, topic: string) {
+  delete circular.sub
```
lpsinger, line 27: Does modifying the Circular in place like that possibly change its value in other functions that might require `sub` to be present? If so, you could remove the `sub` field non-destructively by doing something like:

```export async function sendKafka({sub, ...circular}: Circular, topic: string)```

Also, I notice that this function is in the directory app/lib, so it shouldn't have any Circular-specific code. Move Circular-specific code into table-streams/circulars. You might have a general function to send _any_ Kafka record here and a function that calls it in table-streams/circulars.

**Review by lpsinger — CHANGES_REQUESTED** [review]

File: app/lib/kafka.server.ts (hunk 5 above)
Comment (dakota002) [reply app/lib/kafka.server.ts]: Okay, I have something in the latest commit that should work. The listener gets added in, but locally there is no SIGTERM in sandbox. I'm not 100% confident in this implementation, but it should work

File: app/lib/env.server.ts [hunk 11]
```diff
@@ -44,6 +44,22 @@ function getHostname() {
   return new URL(getOrigin()).hostname
 }
 
+function getDomain() {
+  const hostname = getHostname() as
+    | 'gcn.nasa.gov'
+    | 'test.gcn.nasa.gov'
+    | 'dev.gcn.nasa.gov'
+    | undefined
```
lpsinger, line 52: getHostname is not constrained to return one of those values.

```suggestion
  const hostname = getHostname()
```

app/lib/env.server.ts
```diff
@@ -44,6 +44,22 @@ function getHostname() {
...
+
+  if (hostname === 'gcn.nasa.gov') {
+    return undefined
```
lpsinger, line 55:
```suggestion
    return null
```

app/lib/env.server.ts
```diff
@@ -44,6 +44,22 @@ function getHostname() {
...
+  } else if (hostname?.endsWith('gcn.nasa.gov')) {
+    return hostname
+  } else {
+    return 'dev.gcn.nasa.gov'
```
lpsinger, line 59: Please keep the default set to the public testing server. An external contributor would have access to that but not to dev.gcn.nasa.gov.

```suggestion
    return 'test.gcn.nasa.gov'
```

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+    messages: [
+      {
+        key: 'message',
+        value: JSON.stringify(item),
+      },
+    ],
+  })
+  if (hostname === 'localhost') {
+    await producer.disconnect()
+  }
```
lpsinger, line 38: What is this for?

```suggestion
```

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+  domain,
+}).producer()
+
+process.on('SIGTERM', async () => {
+  await producer.disconnect()
+  process.exit(0)
+})
+
+export async function sendKafka(item: any, topic: string) {
+  await producer.connect()
```
lpsinger, line 26: Please lift this out to module scope so that it is amortized across Lambda warm starts.

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+  await producer.send({
+    topic,
+    messages: [
+      {
+        key: 'message',
+        value: JSON.stringify(item),
```
lpsinger, line 32: Let's leave it up to the calling code to convert the message to a string, so that we aren't limited here to JSON.

```suggestion
        value: item,
```

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+  client_secret,
+  domain,
+}).producer()
+
+process.on('SIGTERM', async () => {
+  await producer.disconnect()
+  process.exit(0)
+})
+
+export async function sendKafka(item: any, topic: string) {
```
lpsinger, line 25: Switch order of arguments to more closely match the arguments to producer.send.

```suggestion
export async function sendKafka(topic: string, value: string) {
```

app/lib/kafka.server.ts
```diff
@@ -0,0 +1,39 @@
...
+  await producer.connect()
+  await producer.send({
+    topic,
+    messages: [
+      {
+        key: 'message',
```
lpsinger, line 31: I think that you can safely leave the key unspecified.

```suggestion
```

lpsinger (changes requested):

app/lib/env.server.ts
dakota002: True, but they are the only valid values for domain, so should this cast be done elsewhere?

dakota002: For example, in kafka.server.ts:

```js
const producer = new Kafka({
  client_id,
  client_secret,
  domain: domain as
    | 'gcn.nasa.gov'
    | 'test.gcn.nasa.gov'
    | 'dev.gcn.nasa.gov'
    | undefined,
}).producer()
```

app/lib/env.server.ts
dakota002: Will be in next commit

app/lib/kafka.server.ts
dakota002: Moving this up results in this error: 
image

app/lib/kafka.server.ts
dakota002: I left a comment in a previous thread, but there is no SIGTERM in the sandbox mode. If you don't disconnect eventually, the lambda hangs and throws an error

app/lib/env.server.ts
lpsinger: > True, but they are the only valid values for domain, so should this cast be done elsewhere?

They are the _only valid return values from this function_. However, `getHostname()` can return any string, or `undefined`. So this cast is incorrect.

app/lib/kafka.server.ts
lpsinger: Alright. Please submit a bug report for Architect and explain why this is here with a FIXME comment.

And please do close the connection at exit.

app/lib/kafka.server.ts
lpsinger: One workaround is to use `memoizee(..., { promise: true })`. We have several examples of that in our code.

Or we could have esbuild compile to ESM for us so that top-level awaits are legal.

lpsinger: Here is the change to switch to an ESM build:

```patch
diff --git a/esbuild.js b/esbuild.js
index 21a02fc0..c1b4a43d 100644
--- a/esbuild.js
+++ b/esbuild.js
@@ -12,11 +12,12 @@ const entryPoints = await glob(
  */
 const options = {
   bundle: true,
+  format: 'esm',
   entryPoints,
   logLevel: 'info',
   outdir: 'build',
   outbase: 'app',
-  outExtension: { '.js': '.cjs' },
+  outExtension: { '.js': '.mjs' },
   external: ['@aws-sdk/*', 'aws-sdk'],
   platform: 'node',
   target: ['node18'],
```

app/lib/kafka.server.ts
dakota002: It may not be architect specifically. For local testing we are triggering this with the invoker plugin. I'm testing their end function to see if maybe that can add an emit without totally also killing the main app too

dakota002: Nevermind, I am looking over the [sandbox invoke](https://github.com/architect/sandbox/blob/main/src/invoke-lambda/index.js) code and think its probably because of how this runs

app/lib/kafka.server.ts
dakota002: I'll opt for the memoizee, I tried the ESM build and got errors about dynamic imports of libraries not being supported

lpsinger: Yeah, CJS vs ESM is a mess.

lpsinger: The `await producer.connect()` _must_ be at module scope. Please add that to the promise-returning function that is memoized.

lpsinger (changes requested):

app/lib/env.server.ts
lpsinger: See #2295.

lpsinger:
Does this work now?

app/root.tsx
```diff
@@ -210,6 +210,18 @@ export function useHostname() {
...
+  const hostname = useHostname()
+
+  if (hostname === 'gcn.nasa.gov') {
+    return null
+  } else if (hostname?.endsWith('gcn.nasa.gov')) {
+    return hostname
+  } else {
+    return 'test.gcn.nasa.gov'
+  }
+}
```
lpsinger, line 223: Oops, you still have the old version of this function. Please rebase.

lpsinger (changes requested):

esbuild.js
```diff
@@ -22,6 +22,7 @@ const options = {
   target: ['node18'],
   minify: !dev,
   sourcemap: dev,
+  loader: { '.node': 'empty' },
```
lpsinger, line 25: Set this to `file` so that the NAPI module gets copied to the output directory.

```suggestion
  loader: { '.node': 'file' },
```

lpsinger (changes requested):

lpsinger (changes requested):

dakota002:
@lpsinger This still doesn't disconnect locally. It may work when deployed though since there will be different emitted events and not just an invoked function

lpsinger:
> @lpsinger This still doesn't disconnect locally. It may work when deployed though since there will be different emitted events and not just an invoked function

Does it disconnect locally if you hook into the SIGTERM signal?

lpsinger (commented):
Does this work now in your sandbox?

dakota002:
> Does this work now in your sandbox?

Yea, just ran it locally and had to add a couple lines but its looking good locally

app/routes/docs.contributing.configuration/route.mdx
```diff
@@ -170,5 +170,12 @@ account for GCN production by registering the email address
       </Description>
       <Default>TNS tooltips disabled</Default>
     </tr>
+    <tr>
+      <Key>`KAFKA_CLIENT_ID`, `KAFKA_CLIENT_SECRET`</Key>
+      <Description>
+        Client credentials for producing GCN Circulars over Kafka
+      </Description>
+      <Default>Kafka producer is disabled</Default>
+    </tr>
```
lpsinger, line 179: What were the pros and cons of creating dedicated client credentials vs. adding the new scope to the web app's existing app client?

dakota002: Ah so I tried using them, but it threw an error about the credentials already used. Let me run it real quick and I can copy the error here

lpsinger: You need to go into the Cognito console, find your app client, and add "Client credentials" to "OAuth 2.0 grant types".

dakota002: I think this warning is telling us that it won't work as we want it to:
image

If you select it, it removed the auth code grant

lpsinger: Okay, thanks for checking that!

lpsinger (approved):

lpsinger:
@dakota002, please follow the directions at https://gcn.nasa.gov/docs/contributing/configuration to configure the two new env variables on dev, then check that this is working as expected.

dakota002:
@lpsinger , I found an issue: the lambda is throwing the following error:
```
2024-05-31T13:53:45.696Z	undefined	ERROR	Uncaught Exception 	{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module '@mongodb-js/zstd-linux-arm64-gnu'\nRequire stack:\n- /var/task/index.cjs\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module '@mongodb-js/zstd-linux-arm64-gnu'",
        "Require stack:",
        "- /var/task/index.cjs",
        "- /var/runtime/index.mjs",
        "    at _loadUserApp (file:///var/runtime/index.mjs:1087:17)",
        "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1119:21)",
        "    at async start (file:///var/runtime/index.mjs:1282:23)",
        "    at async file:///var/runtime/index.mjs:1288:1"
    ]
}
```
If you look into the lambda code folder it has x64 versions of the zstd library. I think we might just need to specify an arm64 architecture in the `npm install` action so it builds the correct one for the lambda's environment. Does that make sense?

lpsinger:
See https://github.com/nasa-gcn/gcn.nasa.gov/pull/2328.

PR #2262, opened 2024-05-06
Try another syntax for disabling the Architect CDN

lpsinger:
This pragma is undocumented. This is another trial-and-error attempt to disable it.

See https://github.com/nasa-gcn/gcn.nasa.gov/pull/2261.


Courey (approved):

Courey:
does it run locally this way? Just asking because playwright is failing by timing out.

PR #2264, opened 2024-05-06
Mod submitter date update

### LINKED ISSUE #2237: edit submitter name/email and date in moderator interface
# Description
In order to backfill the Circulars early archive (-4 to 30), see issue #752, we need to be able to edit the name, email address, and date of the submission in the moderator interface.

# Acceptance criteria
The moderator interface allows the moderator to edit the from field and date.

dakota002:
# Description

Updates the correction request edit form to allow for changes to Submission Date (createdOn) and Submitter

Submission Date is only allowed to be edited if it does not already exist 

# Related Issue(s)

Resolves #2237

# Testing

- Sign in 
- Go to a circular
- Request a correction
- Fill out the form

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -35,6 +35,7 @@ import {
   createChangeRequest,
   get,
   getChangeRequests,
+  moderatorGroup,
   put,
   putVersion,
   search,
@@ -97,9 +98,22 @@ export async function action({ request }: ActionFunctionArgs) {
       if (circularId === undefined)
         throw new Response('circularId is required', { status: 400 })
       if (!user?.name || !user.email) throw new Response(null, { status: 403 })
-
+      let submitter, createdOnDate, createdOnTime, createdOn
+      if (user.groups.includes(moderatorGroup)) {
+        submitter = getFormDataString(data, 'submitter')
+        createdOnDate = getFormDataString(data, 'createdOnDate')
+        createdOnTime = getFormDataString(data, 'createdOnTime')
+        createdOn = Date.parse(`${createdOnDate} ${createdOnTime} UTC`)
+      }
+      if (!submitter || !createdOnDate || !createdOnTime || !createdOn)
+        throw new Response(null, { status: 400 })
       await createChangeRequest(
-        { circularId: parseFloat(circularId), ...props },
+        {
+          circularId: parseFloat(circularId),
+          ...props,
+          submitter,
+          createdOn,
+        },
         user
       )
       await postZendeskRequest({
--- a/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
@@ -9,11 +9,14 @@ import { Form, Link, useNavigation } from '@remix-run/react'
 import {
   Button,
   ButtonGroup,
+  DatePicker,
+  Grid,
   Icon,
   InputGroup,
   InputPrefix,
   Table,
   TextInput,
+  TimePicker,
 } from '@trussworks/react-uswds'
 import classnames from 'classnames'
 import { type ReactNode, useContext, useState } from 'react'
@@ -24,12 +27,17 @@ import { MarkdownBody } from '../circulars.$circularId.($version)/Body'
 import {
   type CircularFormat,
   bodyIsValid,
+  dateIsValid,
   subjectIsValid,
+  submitterIsValid,
 } from '../circulars/circulars.lib'
 import { RichEditor } from './RichEditor'
 import { CircularsKeywords } from '~/components/CircularsKeywords'
 import CollapsableInfo from '~/components/CollapsableInfo'
 import Spinner from '~/components/Spinner'
+import { useModStatus } from '~/root'
+
+import styles from './CircularsEditForm.module.css'
 
 function SyntaxExample({
   label,
@@ -103,20 +111,24 @@ export function SyntaxReference() {
 export function CircularEditForm({
   formattedContributor,
   circularId,
-  submitter,
+  defaultSubmitter,
   defaultFormat,
   defaultBody,
   defaultSubject,
   searchString,
+  defaultCreatedOnDate,
+  defaultCreatedOnTime,
   intent,
 }: {
   formattedContributor: string
   circularId?: number
-  submitter?: string
+  defaultSubmitter?: string
   defaultFormat?: CircularFormat
   defaultBody: string
   defaultSubject: string
   searchString: string
+  defaultCreatedOnDate?: string
+  defaultCreatedOnTime?: string
   intent: 'correction' | 'edit' | 'new'
 }) {
   let formSearchString = '?index'
@@ -130,9 +142,15 @@ export function CircularEditForm({
   const [body, setBody] = useState(defaultBody)
   const [subject, setSubject] = useState(defaultSubject)
   const [format, setFormat] = useState(defaultFormat)
+  const [date, setDate] = useState(defaultCreatedOnDate)
+  const [time, setTime] = useState(defaultCreatedOnTime ?? '12:00')
+  const dateValid = circularId ? dateIsValid(date, time) : true
+
+  const [submitter, setSubmitter] = useState(defaultSubmitter)
+  const submitterValid = circularId ? submitterIsValid(submitter) : true
   const bodyValid = bodyIsValid(body)
   const sending = Boolean(useNavigation().formData)
-  const valid = subjectValid && bodyValid
+  const valid = subjectValid && bodyValid && dateValid && submitterValid
   let headerText, saveButtonText
 
   switch (intent) {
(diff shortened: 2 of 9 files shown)


package-lock.json
```diff
@@ -126,7 +126,7 @@
         "yarn": "^1.22.21"
       },
       "engines": {
-        "node": ">=20"
+        "node": ">=20 <22"
```
lpsinger, line 129: This is `">=20"` on main. Would you please pull from main and rebase?

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -9,6 +9,7 @@ import { Form, Link, useNavigation } from '@remix-run/react'
 import {
   Button,
   ButtonGroup,
+  DatePicker,
```
lpsinger, line 12: I think that we record the time with second precision. We probably need a date-time picker.

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: How accurate do we care about making it? I may need to revise this a bit, as I assumed the only time this would be used is if the date is not defined, like on https://gcn.nasa.gov/circulars/12, but I am now seeing it DOES have a time (just 0).

I also assumed we did not know the exact hour/min/sec time that one of these older circulars came in, is that true?

lpsinger: In many cases we know down to the hour or even more accurately when these occurred because these were intensely studied events. The time may be apparent from the text of the Circular. There was even a book written for lay people about GRBs that probably has a pretty accurate timeline for some of those early GRBs.

Who knows, the timestamp on the filesystem on capella2 may even be accurate.

Are we sure that we want to disable date corrections for Circulars that already have dates? That makes this kind of a one-off feature.

dakota002: Hm thats a good point. I didn't know how much data was tracked on the initial files that came in, but if we know with that degree of accuracy we should preserve it.

 I am not opposed to making it enabled for any circular. Its entirely likely that some newer ones may need to be updated. 

The react uswds library does not contain a Datetime picker, just distinct Date and Time pickers. But I could probably whip  one up and commit it to their repo when it is ready

lpsinger: I think a date picker plus a time picker would be fine.

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
lpsinger, line 1: Why did we need to put everything into a Grid here?

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -164,150 +181,214 @@ export function CircularEditForm({
...
+    format !== defaultFormat ||
+    submitter !== updatedSubmitter?.trim() ||
+    Date.parse(`${defaultDate[0]} ${defaultDate[1].substring(0, 5)}`) !==
+      dateNumber
+
+  const userIsModerator = useModStatus()
+
+  useEffect(() => {
+    setDateString(`${date} ${time} UTC`)
+  }, [date, time])
```
lpsinger, line 193: Why do we need `useEffect` here?

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -164,150 +181,214 @@ export function CircularEditForm({
   const changesHaveBeenMade =
     body.trim() !== defaultBody.trim() ||
     subject.trim() !== defaultSubject.trim() ||
-    format !== defaultFormat
+    format !== defaultFormat ||
+    submitter !== updatedSubmitter?.trim() ||
+    Date.parse(`${defaultDate[0]} ${defaultDate[1].substring(0, 5)}`) !==
```
lpsinger, line 186: You're doing a lot of unnecessary conversion back and forth. I suggest instead:

```ts
export function CircularEditForm({
  ...
 defaultCreatedOn,
 ...
} : {
  defaultCreatedOn
})
{
  const [createdOn, setCreatedOn] = useState(defaultCreatedOn)
  ...
  const changesHaveBeenMade = ... || createdOn !== defaultCreatedOn
}
```

app/routes/circulars.moderation.$circularId.$requestor.tsx
```diff
@@ -9,6 +9,9 @@ import type { SEOHandle } from '@nasa-gcn/remix-seo'
 import type { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node'
 import { Form, redirect, useLoaderData } from '@remix-run/react'
 import { Button, ButtonGroup } from '@trussworks/react-uswds'
+import dayjs from 'dayjs'
+import locale from 'dayjs/locale/en'
+import RelativeTime from 'dayjs/plugin/relativeTime'
```
lpsinger, line 14: This and the following code:

```ts
dayjs.locale(locale)
dayjs.extend(RelativeTime)
```

are duplicated from TimeAgo.tsx. Please refactor.

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: For the Date and Time fields, they were VERY poorly aligned by default on both mobile and desktop. I had to use the Grid to give it some structure

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: the dateString needs to update consistently whenever date or time gets updated, can this be better written?

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: I originally went for something like this, but the default CreatedOn time have accuracy as HH:MM:SS, where these input are only allowing HH:MM, so it becomes a bit wonky

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: Update, useEffect is needed, since there are multiple state changes that effect the dateString. Without it, I noticed inconsistencies in updates to the field

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -117,6 +127,8 @@ export function CircularEditForm({
   defaultBody: string
   defaultSubject: string
   searchString: string
+  createdOnDate?: string
+  createdOnTime?: string
```
lpsinger, line 131: Would you please call these `defaultCreatedOnDate` and `defaultCreatedOnTime` for consistency with the other default form parameters?

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -130,9 +142,15 @@ export function CircularEditForm({
   const [body, setBody] = useState(defaultBody)
   const [subject, setSubject] = useState(defaultSubject)
   const [format, setFormat] = useState(defaultFormat)
+  const [date, setDate] = useState(createdOnDate)
+  const [time, setTime] = useState(createdOnTime ?? '12:00')
+  const dateValid = circularId ? dateIsValid(date, time) : true
+
+  const [updatedSubmitter, setUpdatedSubmitter] = useState(submitter)
```
lpsinger, line 149: Rename `submiter` to `defaultSubmitter` and `updatedSubmitter` to `submitter` for consistency with other fields.

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -151,117 +169,177 @@ export function CircularEditForm({
   }
   const bodyPlaceholder = useBodyPlaceholder()
 
+  /* FIXME: dateFormat does not work in the date selector, 
+    setting to YYYY-MM-DD prevents form submission. Once fixed, 
+    remove date replacement  */
```
lpsinger, line 174: Are you sure it doesn't work? We are using this branch of react-uswds: https://github.com/trussworks/react-uswds/pull/2726

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
lpsinger: Try https://github.com/trussworks/react-uswds/pull/2726/commits/d8ab66840b718ac83f634827b703a7a501d0eb81.

dakota002: This looks good, but one of the tests is failing, 
image
Would this prevent our build from working if I point to this commit?

lpsinger: I fixed that. Please use the latest code on https://github.com/trussworks/react-uswds/pull/2726.

dakota002: @lpsinger I left a comment on your PR, there is one more line that needs to be updated with this and then it should be working. I tested locally and found the error, but i can't make a PR to your branch because I rebased mine and now they are out of sync, but if you want to test it, just add this unit test to the `describe('with YYYY-MM-DD date format'` block:

```js
 it('returns false if the date is valid', () => {
      const testMin = new Date('May 1, 1988')
      expect(
        isDateInvalid('2023-02-16', INTERNAL_DATE_FORMAT, testMin)
      ).toEqual(false)
    })
```

dakota002:
@lpsinger Okay with the latest rebase it is working

app/routes/circulars._archive._index/route.tsx
```diff
@@ -97,10 +98,23 @@ export async function action({ request }: ActionFunctionArgs) {
...
         throw new Response('circularId is required', { status: 400 })
       if (!user?.name || !user.email) throw new Response(null, { status: 403 })
-
+      let submitter, createdOnDate, createdOnTime, createdOn
+      if (user.groups.includes(moderatorGroup)) {
+        submitter = getFormDataString(data, 'submitter')
+        createdOnDate = getFormDataString(data, 'createdOnDate')
+        console.log(createdOnDate)
+        createdOnTime = getFormDataString(data, 'createdOnTime')
+        console.log(createdOnTime)
```
lpsinger, line 107:
```suggestion
        createdOnTime = getFormDataString(data, 'createdOnTime')
```

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
```diff
@@ -97,10 +98,23 @@ export async function action({ request }: ActionFunctionArgs) {
...
       await createChangeRequest(
-        { circularId: parseFloat(circularId), ...props },
-        user
+        {
+          circularId: parseFloat(circularId),
+          ...props,
+        },
+        user,
+        submitter,
+        createdOn
```
lpsinger, line 117: Please include these as fields in the first argument.

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -150,118 +168,180 @@ export function CircularEditForm({
...
+                  dateFormat="YYYY-MM-DD"
+                />
+              </InputGroup>
+              <InputGroup className="tablet:grid-col-auto">
+                <input
+                  type="hidden"
+                  id="createdOnTime"
+                  name="createdOnTime"
+                  value={time}
+                />
```
lpsinger, line 247: Why do we still need this hidden field?

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
lpsinger: See https://github.com/dakota002/gcn.nasa.gov/pull/314.

lpsinger (changes requested):
- When the date, time, or submitter fields receive focus, the focus border should be on the input group, not the text field itself, so that it surrounds both the text field and the label. Compare with focus behavior for the subject field.
- Implement color-coded validation borders like we have for the other fields.

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: Its an easy work around for now since the TimePicker component does not post values correctly.

 I think the issue is that the underlying ComboBox component assigns the `name`  attribute to the `select` element, but the `value` to the `input` element, so when you post it just keeps posting the default value. These are the elements:

```jsx
<select
        {...selectProps}
        className="usa-select usa-sr-only usa-combo-box__select"
        name={name}
        aria-hidden
        tabIndex={-1}
        defaultValue={state.selectedOption?.value}
        data-testid="combo-box-select">
        {options.map((option) => (
          <option key={option.value} value={option.value}>
            {option.label}
          </option>
        ))}
</select>
<Input
        {...inputProps}
        role="combobox"
        onChange={(e): void => {
          if (inputProps?.onChange) {
            // Allow a custom input onChange handler
            inputProps?.onChange(e)
          }

          dispatch({ type: ActionTypes.UPDATE_FILTER, value: e.target.value })
        }}
        onClick={(): void => dispatch({ type: ActionTypes.OPEN_LIST })}
        onBlur={handleInputBlur}
        onKeyDown={handleInputKeyDown}
        value={state.inputValue}
        focused={state.focusMode === FocusMode.Input}
        aria-owns={listID}
        aria-controls={listID}
        aria-autocomplete="list"
        aria-describedby={assistiveHintID}
        aria-expanded={state.isOpen}
        aria-activedescendant={(state.isOpen && focusedItemId) || ''}
        id={id}
        disabled={isDisabled}
/>
```

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -150,118 +168,180 @@ export function CircularEditForm({
...
-            <input type="hidden" name="circularId" value={circularId} />
+        <Grid>
+          <Grid row>
+            {circularId !== undefined && (
+              <>
+                <input type="hidden" name="circularId" value={circularId} />
+                <InputGroup className="maxw-full">
+                  <InputPrefix className="wide-input-prefix">From</InputPrefix>
+                  <TextInput
+                    autoFocus
```
lpsinger, line 194: We autofocus on the subject field. No point in having two autofocus inputs.

```suggestion
```

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -130,9 +142,15 @@ export function CircularEditForm({
   const [body, setBody] = useState(defaultBody)
   const [subject, setSubject] = useState(defaultSubject)
   const [format, setFormat] = useState(defaultFormat)
+  const [date, setDate] = useState(defaultCreatedOnDate)
+  const [time, setTime] = useState(defaultCreatedOnTime ?? '12:00')
+  const dateValid = circularId ? dateIsValid(date, time) : true
+
+  const [submitter, setUpdatedSubmitter] = useState(defaultSubmitter)
```
lpsinger, line 149:
```suggestion
  const [submitter, setSubmitter] = useState(defaultSubmitter)
```

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -150,118 +168,180 @@ export function CircularEditForm({
...
+                    className="maxw-full"
+                    name="submitter"
+                    id="submitter"
+                    type="text"
+                    defaultValue={defaultSubmitter}
+                    onChange={(event) =>
+                      setUpdatedSubmitter(event.target.value)
+                    }
+                    required
+                    disabled={!userIsModerator}
```
lpsinger, line 204: Let's just not show this field if the user is not a moderator.

lpsinger (changes requested):

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
lpsinger: The select value _is_ included in the form data, but typing a valid value into the box does not necessarily update the select value. In other words, the time picker has useful form data if you choose a value from the combo box, but not if you type in a valid value.

This is clearly a bug in react-uswds. Would you please submit an issue upstream and add a FIXME comment in our code here linking to it and explaining why we have the hidden input?

dakota002: I think they are already aware: https://github.com/trussworks/react-uswds/issues/2806
I can make a PR for it once I finish up this stuff

dakota002:
> We generally show times in 24 hour format on this site. Would you please set the time picker to 24 hour mode?

I don't see that as a supported feature of this component, but I think it would be a good update later, along with the name fix for the TimePicker we were talking about above

dakota002:
Still working on resolving the input focus

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
lpsinger: Oh, cool! Thanks! May I suggest just adding a bit more context to the FIXME comment? Something to the effect of:

> The TimePicker component does not by itself contribute useful form data because only the <select> element has a name, and the <input> field does not. So the form data is only populated correctly if the user selects an option from the dropdown, but not if they type a valid value into the combo box.

lpsinger:
> > We generally show times in 24 hour format on this site. Would you please set the time picker to 24 hour mode?
> 
> I don't see that as a supported feature of this component, but I think it would be a good update later, along with the name fix for the TimePicker we were talking about above

Would you please submit an issue upstream and add a FIXME comment? 24-hour time formats are much more useful in astronomy.

lpsinger:
> Still working on resolving the input focus
> 
> @lpsinger This is proving more challenging than I imagined. The date and time picker inputs use the input group elements and don't really play well with the input prefix component by default. I think this can be fixed in the future, but shouldn't hold up the rest of this PR

Let me take one quick swing at this and see if I can figure it out. Would you please rebase your branch from main?

lpsinger (changes requested):
Please add a validation border around the From field.

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
dakota002: Yup! I will just use your wording for it

dakota002:
> > Still working on resolving the input focus
> > @lpsinger This is proving more challenging than I imagined. The date and time picker inputs use the input group elements and don't really play well with the input prefix component by default. I think this can be fixed in the future, but shouldn't hold up the rest of this PR
> 
> Let me take one quick swing at this and see if I can figure it out. Would you please rebase your branch from main?

Done!

lpsinger (changes requested):
Would you please make my suggestions from https://github.com/dakota002/gcn.nasa.gov/pull/314 about only putting the date and time picker in a Grid and not the entire component?

dakota002:
> Would you please make my suggestions from [dakota002#314](https://github.com/dakota002/gcn.nasa.gov/pull/314) about only putting the date and time picker in a Grid and not the entire component?

Woops, I missed that one, can you rebase it then surely

app/routes/circulars/circulars.lib.ts
```diff
@@ -130,6 +132,15 @@ export function formatIsValid(format: string): format is CircularFormat {
   return (circularFormats as any as string[]).includes(format)
 }
 
+/** For updated dates, check that the date is valid */
+export function dateIsValid(date?: string, time?: string) {
+  return !Number.isNaN(Date.parse(`${date} ${time} UTC`))
```
lpsinger, line 137: This doesn't work in my browser (Safari on macOS), and it prevents the submit button from ever being enabled.

From my browser's JavaScript console:

```
> Date.parse('2023-09-08 15:59 UTC')
NaN
```

lpsinger (changes requested):

app/routes/circulars/circulars.lib.ts
dakota002: Does the following parse correctly on safari? 
```js
Date.parse('2023-09-08T15:59:00.000Z')
```

Based it off https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

lpsinger: Yes.

dakota002: Okay i will adjust the formatting to something that will parse correctly

lpsinger (approved):

lpsinger:
I get an HTTP 400 error when I submit the correction request form if I am _not_ a moderator. Please fix.

PR #2268, opened 2024-05-10
Add user group management 

dakota002:
# Description

Adds admin pages for managing user's group associations

# Testing

Ran locally, tested adding/removing groups that user's belong to

### DIFF
--- a/app/components/UserLookup.tsx
+++ b/app/components/UserLookup.tsx
@@ -1,0 +1,180 @@
+import { useFetcher } from '@remix-run/react'
+import type { action } from 'app/routes/api.users'
+import classnames from 'classnames'
+import type { UseComboboxProps, UseComboboxStateChange } from 'downshift'
+import { useCombobox } from 'downshift'
+import { debounce } from 'lodash'
+import {
+  forwardRef,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useState,
+} from 'react'
+
+import { formatAuthor } from '~/routes/circulars/circulars.lib'
+
+import loaderImage from 'nasawds/src/img/loader.gif'
+
+export interface UserLookup {
+  sub?: string
+  email: string
+  name?: string
+  affiliation?: string
+}
+
+interface UserComboBoxProps
+  extends Omit<
+    UseComboboxProps<UserLookup>,
+    'items' | 'onInputValueChange' | 'itemToString'
+  > {
+  disabled?: boolean
+  className?: string
+  group?: string
+  onChange?: () => void
+}
+
+interface UserComboBoxHandle {
+  reset: () => void
+}
+
+const UserLookupComboBox = forwardRef<UserComboBoxHandle, UserComboBoxProps>(
+  ({ disabled, className, group, onChange, ...props }, ref) => {
+    const fetcher = useFetcher<typeof action>()
+    const [items, setItems] = useState<UserLookup[]>([])
+
+    useEffect(() => {
+      setItems(fetcher.data?.users ?? [])
+    }, [fetcher.data])
+
+    if (onChange) onChange()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const onInputValueChange = useCallback(
+      debounce(
+        ({ inputValue, isOpen }: UseComboboxStateChange<UserLookup>) => {
+          if (inputValue && isOpen) {
+            const data = new FormData()
+            data.set('filter', inputValue.split(' ')[0])
+            data.set('intent', 'filter')
+            data.set('group', group ?? '')
+            fetcher.submit(data, { method: 'POST', action: '/api/users' })
+          } else {
+            setItems([])
+          }
+        },
+        500,
+        { trailing: true }
+      ),
+      []
+    )
+
+    const {
+      reset,
+      isOpen,
+      highlightedIndex,
+      selectedItem,
+      getMenuProps,
+      getInputProps,
+      getItemProps,
+      getToggleButtonProps,
+    } = useCombobox<UserLookup>({
+      items,
+      onInputValueChange,
+      itemToString(item) {
+        return item ? formatAuthor(item) : ''
+      },
+      ...props,
+    })
+
+    useImperativeHandle(
+      ref,
+      () => ({
+        reset,
+      }),
+      [reset]
+    )
+
+    const loading = fetcher.state === 'submitting'
+    const pristine = Boolean(selectedItem)
+
+    return (
+      <div
+        data-testid="combo-box"
+        data-enhanced="true"
+        className={classnames('usa-combo-box', className, {
+          'usa-combo-box--pristine': pristine || loading,
+        })}
+      >
+        <input
+          autoCapitalize="off"
+          autoComplete="off"
+          className="usa-combo-box__input"
+          disabled={disabled}
+          {...getInputProps()}
+          // Funky escape sequence is a zero-width character to prevent Safari
+          // from attempting to autofill the user's own email address, which
+          // would be triggered by the presence of the string "email" in the
+          // placeholder.
+          placeholder="Name or em&#8203;ail address of user"
+        />
+        <span className="usa-combo-box__clear-input__wrapper" tabIndex={-1}>
+          <button
+            type="button"
+            className="usa-combo-box__clear-input"
+            aria-label="Clear the select contents"
+            style={
+              loading ? { backgroundImage: `url('${loaderImage}')` } : undefined
+            }
+            onClick={() => reset()}
+            hidden={(!pristine || disabled) && !loading}
+            disabled={disabled}
+          >
+            &nbsp;
+          </button>
+        </span>
+        <span className="usa-combo-box__input-button-separator">&nbsp;</span>
+        <span className="usa-combo-box__toggle-list__wrapper" tabIndex={-1}>
+          <button
+            type="button"
+            className="usa-combo-box__toggle-list"
+            {...getToggleButtonProps()}
+          >
+            &nbsp;
+          </button>
+        </span>
+        <ul
+          {...getMenuProps()}
+          className="usa-combo-box__list"
+          role="listbox"
+          hidden={!isOpen}
+        >
+          {isOpen &&
+            (items.length ? (
+              items.map((item, index) => (
+                <li
+                  key={item.sub}
+                  className={classnames('usa-combo-box__list-option', {
+                    'usa-combo-box__list-option--focused':
+                      index === highlightedIndex,
+                    'usa-combo-box__list-option--selected':
+                      selectedItem?.sub === item.sub,
+                  })}
+                  {...getItemProps({ item, index })}
+                >
+                  {formatAuthor(item)}
+                </li>
+              ))
+            ) : (
+              <li className="usa-combo-box__list-option--no-results">
+                No results found
+              </li>
+            ))}
+        </ul>
+      </div>
+    )
+  }
+)
+
+UserLookupComboBox.displayName = 'UserLookupComboBox'
+export default UserLookupComboBox
--- a/app/components/UserLookup.tsx
+++ b/app/components/UserLookup.tsx
@@ -1,16 +1,17 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
 import { useFetcher } from '@remix-run/react'
 import type { action } from 'app/routes/api.users'
 import classnames from 'classnames'
 import type { UseComboboxProps, UseComboboxStateChange } from 'downshift'
 import { useCombobox } from 'downshift'
 import { debounce } from 'lodash'
-import {
-  forwardRef,
-  useCallback,
-  useEffect,
-  useImperativeHandle,
-  useState,
-} from 'react'
+import { useCallback, useEffect, useState } from 'react'
 
 import { formatAuthor } from '~/routes/circulars/circulars.lib'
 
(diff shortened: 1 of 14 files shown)


lpsinger:
Could you upload a few screen shots to illustrate the new functionality?

app/routes/admin.users._index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const user = await getUser(request)
+  if (!user?.groups.includes('gcn.nasa.gov/gcn-admin'))
+    throw new Response(null, { status: 403 })
+  const users = await listUsers()
+  return { users }
+}
+
+export default function Users() {
```
lpsinger, line 24:
```suggestion
export default function () {
```

app/routes/admin.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { getUser } from './_auth/user.server'
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const user = await getUser(request)
+  if (!user?.groups.includes('gcn.nasa.gov/gcn-admin'))
+    throw new Response(null, { status: 403 })
+  return null
+}
+
+export default function Admin() {
```
lpsinger, line 21:
```suggestion
export default function () {
```

app/lib/cognito.server.ts
```diff
@@ -58,6 +69,55 @@ export function extractAttributeRequired(
...
+    throw new Response('Requested user does not exist', {
+      status: 400,
+    })
+
+  const groups = (await listGroupsForUser(user.Username)).map(
+    (group) => group.GroupName
+  )
+  return { ...user, groups }
+}
+export async function listUsers() {
```
lpsinger, line 93:
```suggestion
}

export async function listUsers() {
```

app/lib/cognito.server.ts
```diff
@@ -58,6 +69,55 @@ export function extractAttributeRequired(
...
+        UserPoolId,
+        Filter: `sub = "${escapedSub}"`,
+      })
+    )
+  )?.Users?.[0]
+
+  if (!user?.Username)
+    throw new Response('Requested user does not exist', {
+      status: 400,
+    })
```
lpsinger, line 86: This is copied from endorsements.sever.ts. Please refactor. Don't duplicate code.

app/lib/cognito.server.ts
```diff
@@ -58,6 +69,55 @@ export function extractAttributeRequired(
...
+
+  const groups = (await listGroupsForUser(user.Username)).map(
+    (group) => group.GroupName
+  )
+  return { ...user, groups }
```
lpsinger, line 91: Seems like this is mixing up two distinct pieces of functionality in one function.

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from '~/lib/cognito.server'
+
+export async function loader({
+  params: { userId },
+  request,
+}: LoaderFunctionArgs) {
+  const currentUser = await getUser(request)
+  if (!currentUser?.groups.includes('gcn.nasa.gov/gcn-admin'))
+    throw new Response(null, { status: 403 })
+  if (!userId) throw new Response(null, { status: 400 })
```
lpsinger, line 29:
```suggestion
  if (!userId) throw new Response(null, { status: 404 })
```

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,114 @@
...
+  return { user, allGroups }
+}
+
+export async function action({ request }: ActionFunctionArgs) {
+  const user = await getUser(request)
+  if (!user?.groups.includes('gcn.nasa.gov/gcn-admin'))
+    throw new Response(null, { status: 403 })
+  const data = await request.formData()
+  const { Username, ...selectedGroups } = Object.fromEntries(data)
+  if (!Username) throw new Response(null, { status: 400 })
```
lpsinger, line 41: Is it necessary to include the username in the form data? Isn't it present in the route params?

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,114 @@
...
+      .filter((x) => !currentUserGroups.includes(x))
+      .map((x) => addUserToGroup(Username.toString(), x))
+  )
+  // Remove user from these
+  promises.push(
+    ...currentUserGroups
+      .filter((x) => !selectedGroupsNames.includes(x))
+      .map((x) => removeUserFromGroup(Username.toString(), x))
+  )
+  await Promise.all(promises)
```
lpsinger, line 59:
```suggestion
  await Promise.all([
    ...selectedGroupsNames
      .filter((x) => !currentUserGroups.includes(x))
      .map((x) => addUserToGroup(Username.toString(), x)),
    ...currentUserGroups
      .filter((x) => !selectedGroupsNames.includes(x))
      .map((x) => removeUserFromGroup(Username.toString(), x))
  ])
```

app/routes/admin.users._index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  useEffect(() => {
+    setDisplayUsers(
+      users.filter(
+        (user) =>
+          user.email.toLowerCase().includes(filterString.toLowerCase()) ||
+          user.name?.toLowerCase().includes(filterString.toLowerCase())
+      )
+    )
+  }, [users, filterString])
```
lpsinger, line 37: This can be done without useEffect.

app/routes/api.circulars.ts
```diff
@@ -100,7 +100,7 @@ async function getUserAttributes(Username: string) {
   }
 }
 
-async function getUserGroups(Username: string) {
+export async function getUserGroups(Username: string) {
```
lpsinger, line 103: Since we are resuing this function in multiple routes now, move this to a library module like cognito.server.ts.

lpsinger (changes requested):

app/lib/cognito.server.ts
dakota002: Yea, good catch. I will fix this in the next commit

app/routes/api.circulars.ts
dakota002: I actually don't think this was intended to be modified, but I will refactor it to use some of the new functions since it is basically doing the same thing anyway

app/lib/cognito.server.ts
```diff
@@ -162,6 +233,13 @@ export async function listGroupsForUser(Username: string) {
   return groups
 }
 
+export async function getUserGroupStrings(Username: string) {
+  const Groups = await listGroupsForUser(Username)
+  return Groups?.map(({ GroupName }) => GroupName).filter(Boolean) as
+    | string[]
+    | undefined
+}
```
lpsinger, line 241: This function is only used in endorsements.server.ts. Would you please move it there?

app/lib/cognito.server.ts
```diff
@@ -122,6 +186,13 @@ export async function getGroups() {
   return groups
 }
 
+export async function allGroupNames(): Promise<string[]> {
+  const names = (await getGroups())
+    .map((group) => group.GroupName)
+    .filter((group) => typeof group === 'string') as string[]
+  return names
+}
+
```
lpsinger, line 195: I think it would be nice to show on the page both the group names _and_ their descriptions. So no need for this specialized function that only returns the names.

app/lib/cognito.server.ts
```diff
@@ -58,6 +69,59 @@ export function extractAttributeRequired(
...
+            name: extractAttribute(user.Attributes, 'name'),
+            affiliation: extractAttribute(
+              user.Attributes,
+              'custom:affiliation'
+            ),
+          }))
+      )
+  }
+  return users
+}
```
lpsinger, line 123: This function may take a painfully long time to execute in production with thousands of users. Instead of displaying a list of all of the users, could we have an autocomplete search box like we do for peer endorsements?

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <h1>Manage User Settings</h1>
+      {user.Attributes?.find((x) => x.Name == 'email')?.Value}
+      <h2>Groups</h2>
+      <fetcher.Form method="POST">
+        <input
+          type="hidden"
+          name="Username"
+          id="Username"
+          value={user.Username}
+        />
```
lpsinger, line 78: Remove hidden field. We have to validate any input we get from any form, which would mean validating that this username matches the sub claim, which would defeat the performance optimization of storing it in a hidden field.

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,111 @@
...
+        {allGroups.map(
+          (groupName) =>
+            groupName && (
```
lpsinger, line 81: How would `groupName` ever be falsey?

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Checkbox
+      id={groupName}
+      name={groupName}
+      label={groupName}
+      checked={checked}
+      onChange={() => setChecked(!checked)}
+    />
+  )
```
lpsinger, line 110:
```suggestion
  return (
    <Checkbox
      id={groupName}
      name={groupName}
      label={groupName}
      defaultChecked={userGroups.includes(groupName)}
    />
  )
```

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,111 @@
...
+            )
+        )}
+        <Button type="submit">Save</Button>
+      </fetcher.Form>
+    </>
+  )
+}
+
+function GroupsCheckbox({
+  userGroups,
```
lpsinger, line 94: Could we do this processing to determine if a field should be checked by default on the loader side instead of in the component?

app/routes/admin.users._index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Table stackedStyle="default" fullWidth>
+        <thead>
+          <tr>
+            <th>Email</th>
+            <th>Name</th>
+            <th></th>
+          </tr>
+        </thead>
+        <tbody>
+          {displayUsers &&
```
lpsinger, line 55: How would `displayUsers` be falsey?

app/lib/cognito.server.ts
```diff
@@ -18,15 +18,26 @@ import {
...
 const UserPoolId = process.env.COGNITO_USER_POOL_ID
 
+export interface UserData {
+  email: string
+  sub?: string
+  name?: string
+  affiliation?: string
+  receive?: boolean
+  submit?: boolean
+}
```
lpsinger, line 39: This should not go in `cognito.server.ts`. This file contains utilities for dealing with Cognito and should not have any details about our own data model.

lpsinger (changes requested):

app/lib/cognito.server.ts
dakota002: Yea, I will implement that in the next commit

app/routes/admin.users.$userId.tsx
dakota002: This shouldn't be false in practice, but the GroupType object has the GroupName property defined as either string or undefined in the cognito idp library

app/routes/admin.users.$userId.tsx
dakota002: Yup! will also be in next commit

app/routes/admin.users._index.tsx
dakota002: This will actually just be removed

dakota002:
Okay now it is ready

app/components/UserLookup.tsx
lpsinger, line 1: Add copyright/license banner.

app/components/UserLookup.tsx
```diff
@@ -0,0 +1,180 @@
...
+              </li>
+            ))}
+        </ul>
+      </div>
+    )
+  }
+)
+
+UserLookupComboBox.displayName = 'UserLookupComboBox'
+export default UserLookupComboBox
```
lpsinger, line 180: My personal preference is not to have default exports in component modules.

```suggestion
```

app/components/UserLookup.tsx
```diff
@@ -0,0 +1,180 @@
...
+  group?: string
+  onChange?: () => void
+}
+
+interface UserComboBoxHandle {
+  reset: () => void
+}
+
+const UserLookupComboBox = forwardRef<UserComboBoxHandle, UserComboBoxProps>(
+  ({ disabled, className, group, onChange, ...props }, ref) => {
```
lpsinger, line 42: The following will give the React component a display name:

```suggestion
  function UserLookupComboBox({ disabled, className, group, onChange, ...props }, ref) {
```

lpsinger (changes requested):

app/components/UserLookup.tsx
```diff
@@ -0,0 +1,180 @@
...
+                No results found
+              </li>
+            ))}
+        </ul>
+      </div>
+    )
+  }
+)
+
+UserLookupComboBox.displayName = 'UserLookupComboBox'
```
lpsinger, line 179: Remove, see https://github.com/nasa-gcn/gcn.nasa.gov/pull/2268#discussion_r1700739216.

```suggestion
```

lpsinger (changes requested):

app/routes/api.users.ts
lpsinger, line 1: Uh oh. This API endpoint will allow any authenticated user to list the members of any group!

We should only allow authenticated users to list Circulars moderators. Admin users may list other groups.

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function GroupsCheckbox({
+  groupName,
+  description,
+  selected,
+}: {
+  groupName: string
+  description: string
+  selected: boolean
```
lpsinger, line 112: Might as well call this prop `defaultChecked` to match the prop name of the Checkbox component itself.

lpsinger (changes requested):

app/lib/cognito.server.ts
```diff
@@ -116,7 +183,14 @@ export async function getGroups() {
   const groups: GroupType[] = []
   for await (const page of pages) {
     const nextGroups = page.Groups
-    if (nextGroups) groups.push(...nextGroups)
+    if (nextGroups)
+      groups.push(
+        ...nextGroups.filter(
+          (group) =>
+            group.GroupName !== undefined &&
+            !group.GroupName.startsWith('us-east-1')
```
lpsinger, line 191: Actually I think that we should only _include_ groups that start with `gcn.nasa.gov/`. Do we have a constant somewhere that defines that group prefix?

app/routes/admin.users.$userId.tsx
```diff
@@ -0,0 +1,124 @@
...
+  ) as string[]
+  const selectedGroupsNames = Object.keys(selectedGroups)
+
+  await Promise.all([
+    ...selectedGroupsNames
+      .filter((x) => !currentUserGroups.includes(x))
+      .map((x) => addUserToGroup(userId, x)),
+    ...currentUserGroups
+      .filter((x) => !selectedGroupsNames.includes(x))
+      .map((x) => removeUserFromGroup(userId, x)),
```
lpsinger, line 73: There is some missing validation here. Submitting this form must not add or remove the special groups that Cognito uses to keep track of which users federated how.

lpsinger (changes requested):

app/lib/cognito.server.ts
dakota002: What about the igwn alerts?

dakota002: I dont think we do, but i can add one if not

lpsinger: I think that group name also starts with gcn.nasa.gov/, doesn’t it?

dakota002: Yep, I was mistaken, should be all good now

lpsinger (approved):

PR #2279, opened 2024-05-15
Make clearing query in search bar show the full circular archive instead of hiding results

### LINKED ISSUE #1110: clean variable in circulars index causing issues with displaying results

tylerbarna:
# Description
The clean variable causes a bug where queries with no results are cleared but the full list of circulars are not shown.

# Related Issue(s)
resolves #1110 

# Testing
- make a search in the circulars archive
- hover over the search bar
- press the grey x
- query should be cleared and the full archive should be shown again

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -165,7 +165,6 @@ export default function () {
   if (searchString) searchString = `?${searchString}`
 
   const [inputQuery, setInputQuery] = useState(query)
-  const clean = inputQuery === query
 
   const formId = useId()
   const submit = useSubmit()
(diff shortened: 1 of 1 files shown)


dakota002:
Can you attach some screenshots or small clip of the behavior with this change?

tylerbarna:
> Can you attach some screenshots or small clip of the behavior with this change?


https://github.com/nasa-gcn/gcn.nasa.gov/assets/57914086/a6ee38a5-966d-436a-972c-999402c21c39

tylerbarna:
It doesn't seem that changing the input of the query changes the results until one presses enter or the search button, with the exception of when one deletes the entire query by pressing backspace/delete. In that case, the full archive will load without pressing enter or the search button.

PR #2280, opened 2024-05-16
Add conditional wildcard queries for object searches

### LINKED ISSUE #2281: Add ability to search for specific event identifiers
# Description
Currently, if a user searches something like "230924" the results for IGWN event 230924an don't load. We can add a conditional check for event identifiers that submits a [wildcard search](https://opensearch.org/docs/latest/query-dsl/term/wildcard/) that will allow the relevant results to be returned.

(Based on Helpdesk request)

# Acceptance criteria

- [ ] Have search function detect if an event identifier is submitted as a search
- [ ] Add wildcards to search queries that contain event identifiers in the backend
- [ ] Wildcard searches can be slower and more costly than standard multi-match searches, so only have the wildcard search trigger in the specific case of an event identifier. If the search is not an event identifier, the current search strategy should occur

tylerbarna:
# Description
If a user submits a query that consists of what could be an object identifier (eg `230919a`), the search function will automatically add wildcard characters to the query and submit a wildcard search. This is done by using a regex. If the regex is true, it strips out the alphabetical characters on either side of the string and then inititates a wildcard search. This wildcard search does not occur if the query does not match the regex.

# Related Issue(s)
Resolves #2281 

# Testing
Test added functionality:
- Navigate to circulars archive
- Initiate a search for some object identifier, like 230919a
- Results should now show any circulars where the object identifier is in the subject
- Compare this to current [search](gcn.nasa.gov/circulars), where same query will not return results

Test standard query:
- Navigate to circulars archive
- Initiate a search that the regex should evaluate as falsey
- Compare the results to the current [search](gcn.nasa.gov/circulars), it should be the same

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -149,6 +149,10 @@ export async function search({
 
   const [startTime, endTime] = getValidDates(startDate, endDate)
 
+  const isWildcardQuery = query && /^[a-zA-Z]*[0-9]+[a-zA-Z]*$/.test(query)
+
+  const cleanedQuery = isWildcardQuery ? query.replace(/[a-zA-Z]/g, '') : query
+
   const sortObj =
     sort === 'relevance' && query
       ? {}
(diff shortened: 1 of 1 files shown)


tylerbarna:
Made redundant by planned implementation of lucene style queries in #2288

PR #2290, opened 2024-05-20
Reset Circulars search results on empty query

### LINKED ISSUE #1110: clean variable in circulars index causing issues with displaying results

lpsinger:
The intended behavior of the search box on the Circulars archive page is:

- Hide the list of results and number of results when the user starts typing search terms.
- Display the list of results and number of results when the user submits the form.
- If the user does not submit the form, but wipes out the search terms either by deleting the text in the search box or by clicking the 'x', then show the Circulars (without filtering by search terms) once again.

This behavior broke when we added filtering by date. Fix it, and also prevent the scroll position from jumping when submitting or resetting the form.

Fixes #1110. Closes #2279.

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -151,7 +151,7 @@ export default function () {
   searchParams.delete('index')
 
   const limit = searchParams.get('limit') || '100'
-  const query = searchParams.get('query') || undefined
+  const query = searchParams.get('query') || ''
   const startDate = searchParams.get('startDate') || undefined
   const endDate = searchParams.get('endDate') || undefined
   const sort = searchParams.get('sort') || 'circularID'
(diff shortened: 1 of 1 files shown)


dakota002 (approved):
Looks good to me

tylerbarna (approved):

PR #2306, opened 2024-05-27
Return to stable uswds and nasawds releases

lpsinger:
- https://github.com/uswds/uswds/issues/5825 has been fixed.
- https://github.com/lpsinger/react-uswds/tree/gcn is the same as https://github.com/trussworks/react-uswds/pull/2726, but rebased.
- I had to update some CSS where we have heavily customized the appearance of some USWDS components.


dakota002 (approved):
Looks good to me

PR #2330, opened 2024-05-31
Place circulars table-stream Lambda inside our VPC

lpsinger:
This Lambda talks to our Kafka cluster, which in our dev deployment only accepts external traffic from the NASA network. Place the Lambda inside our VPC so that it can talk to the Kafka cluster regardless.


lpsinger:
@dakota002, I've manually deployed this to dev. Would you please check if your Lambda can connect to the Kafka cluster now?

dakota002 (approved):

dakota002:
Yea, I will run some tests now

dakota002:
Still no luck, same error is getting thrown in cloudwatch

lpsinger:
This didn't help. In this setup, the Lambda still resides “in” the AWS-managed VPC; Lambda VPC private networking simply sets up VPC-to-VPC NAT which is not sufficient to allow ingress given our existing NACLs.

PR #2366, opened 2024-06-12
Handle signals in both Lambda and arc sandbox

lpsinger:
Our problems with handling signals in the Architect sandbox might have been fixed by #2364.

### DIFF
--- a/app/lib/kafka.server.ts
+++ b/app/lib/kafka.server.ts
@@ -18,53 +18,24 @@ const kafka = new Kafka({
   domain,
 })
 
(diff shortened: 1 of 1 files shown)


dakota002:
This is likely the correct solution, however the issue still exists on windows machines. From the docs for the underlying package, [tree-kill](https://www.npmjs.com/package/tree-kill):

"For Windows, this uses `taskkill /pid PID /T /F` to kill the process tree. Note that on Windows, sending the different kinds of POSIX signals is not possible."

PR #2370, opened 2024-06-14
Fixes Edit page loading and adds some simple playwright tests

### LINKED ISSUE #2368: Edit Circular form does not populate date and time fields
Please add some Playwright tests for Circulars submission, moderation, and editing workflows. We have a lot of broken corner cases.

# Current behavior

The date and time fields are blank.

Screenshot 2024-06-13 at 19 54 41

# Expected behavior

Fields should be populated

# Steps to reproduce

Navigate to https://dev.gcn.nasa.gov/circulars/edit/33671

# Environment

Safari, macOS

dakota002:
Resolves #2368

This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ sam.yaml
 /playwright/.cache/
 # https://playwright.dev/docs/auth
 /__playwright__/.auth
-/__playwright__/auth.setup.ts
 auth.json
 
 # Draw.io auto-save files
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ sam.yaml
 /playwright/.cache/
 # https://playwright.dev/docs/auth
 /__playwright__/.auth
+/__playwright__/auth.setup.ts
 auth.json
 
 # Draw.io auto-save files
--- a/__playwright__/auth.setup.ts
+++ b/__playwright__/auth.setup.ts
@@ -6,18 +6,20 @@ dotenv.config()
 
 const authFile = '__playwright__/.auth/user.json'
 
-const testUsername = process.env.TEST_USERNAME
-const testPassword = process.env.TEST_PASSWORD
+const testUsername = 'TEST_USERNAME'
+const testPassword = 'TEST_PASSWORD'
 
 setup('authenticate', async ({ page }) => {
   // Perform authentication steps. Replace these actions with your own.
   if (!testUsername || !testPassword)
     throw new Error('Please define test account info')
   await page.goto('/login')
-  await page.getByRole('textbox', { name: 'name@host.com' }).fill(testUsername)
-  await page.getByRole('textbox', { name: 'Password' }).fill(testPassword)
-  await page.getByRole('button', { name: 'submit' }).click()
+  await page.getByPlaceholder('Enter any login').fill(testUsername)
+  await page.getByPlaceholder('and password').fill(testPassword)
+  await page.getByRole('button', { name: 'Sign-in' }).click()
   await page.waitForURL('/')
-  await expect(page.getByRole('button', { name: testUsername })).toBeVisible()
+  await expect(
+    page.getByRole('button', { name: 'user@example.com' })
+  ).toBeVisible()
   await page.context().storageState({ path: authFile })
 })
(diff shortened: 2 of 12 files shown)


dakota002:
Adding another couple of tests still

lpsinger:
> This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD

Are you sure? The sandbox IdP will accept _any_ username and password.

dakota002:
> > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> 
> Are you sure? The sandbox IdP will accept _any_ username and password.

True, but since these tests cover submission and corrections the test user will need the appropriate permissions

lpsinger:
> > > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> > 
> > 
> > Are you sure? The sandbox IdP will accept _any_ username and password.
> 
> True, but since these tests cover submission and corrections the test user will need the appropriate permissions

How is that related to the username and password?

lpsinger:
This is currently blocking deployments.

dakota002:
> > > > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> > > 
> > > 
> > > Are you sure? The sandbox IdP will accept _any_ username and password.
> > 
> > 
> > True, but since these tests cover submission and corrections the test user will need the appropriate permissions
> 
> How is that related to the username and password?

The tests run `auth.setup.ts` first to sign in a user, then use that auth state for tests

lpsinger:
> > > > > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> > > > 
> > > > 
> > > > Are you sure? The sandbox IdP will accept _any_ username and password.
> > > 
> > > 
> > > True, but since these tests cover submission and corrections the test user will need the appropriate permissions
> > 
> > 
> > How is that related to the username and password?
> 
> The tests run `auth.setup.ts` first to sign in a user, then use that auth state for tests

Yes, but the sandbox IdP will accept _any_ username and password --- it doesn't even check them. Why not just hardcode some arbitrary string?

dakota002:
> > > > > > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> > > > > 
> > > > > 
> > > > > Are you sure? The sandbox IdP will accept _any_ username and password.
> > > > 
> > > > 
> > > > True, but since these tests cover submission and corrections the test user will need the appropriate permissions
> > > 
> > > 
> > > How is that related to the username and password?
> > 
> > 
> > The tests run `auth.setup.ts` first to sign in a user, then use that auth state for tests
> 
> Yes, but the sandbox IdP will accept _any_ username and password --- it doesn't even check them. Why not just hardcode some arbitrary string?

This is half way to the solution, my latest commit now uses this. The other thing is that the fake account just needed the additional  cognito groups and if you test this locally just make sure you switch to using the sandbox mode if you aren't already

lpsinger:
> > > > > > > This will also require adding 2 new env variables, TEST_USERNAME and TEST_PASSWORD
> > > > > > 
> > > > > > 
> > > > > > Are you sure? The sandbox IdP will accept _any_ username and password.
> > > > > 
> > > > > 
> > > > > True, but since these tests cover submission and corrections the test user will need the appropriate permissions
> > > > 
> > > > 
> > > > How is that related to the username and password?
> > > 
> > > 
> > > The tests run `auth.setup.ts` first to sign in a user, then use that auth state for tests
> > 
> > 
> > Yes, but the sandbox IdP will accept _any_ username and password --- it doesn't even check them. Why not just hardcode some arbitrary string?
> 
> This is half way to the solution, my latest commit now uses this. The other thing is that the fake account just needed the additional cognito groups and if you test this locally just make sure you switch to using the sandbox mode if you aren't already

So we need to pre-populate the sandbox IdP with some fake accounts with predefined group claims.

dakota002:
For the scope of this issue, i think it will be fine as is. I do not know what crazy config changes we will also need to make to the playwright.config file to point to specific auth states per test, or there is probably another solution in just running some auth steps before the tests directly, but again I think that would be best left to a separate issue 

 The tests failed, but just because of some timeouts. I have another commit to push if they timeout again that may help

lpsinger (changes requested):

lpsinger (approved):

PR #2400, opened 2024-06-24
Add more ESM-only packages

lpsinger:
PR #2395 added some more ESM-only modules, breaking local development. Add them to the list of ESM-only modules which must always be bundled.


tylerbarna (approved):

lpsinger:
One moment. We may need to revert these updates: https://github.com/nasa-gcn/gcn.nasa.gov/pull/2401

lpsinger:
Nope, never mind, the Playwright tests were already broken, apparently.

PR #2417, opened 2024-06-26
Allow for testing of different roles in playwright config

dakota002:
# Description
- Updates the Playwright config to show how we can create projects to run with different roles (i.e. circular submitters, moderators, admins, etc.)
- Updates the sandbox to allow us to define profiles for different use cases and includes an example admin role


lpsinger (approved):
Very cool!

PR #2432, opened 2024-07-01
Fixes missing fields causing unexpected error in moderator views

### LINKED ISSUE #2431: Moderator page throws unexpected error if a Diff'd string is missing
# Current behavior
Unexpected Error prevents further interaction

# Expected behavior
Undefined strings should behave as empty strings to prevent the error from throwing

# Steps to reproduce
- Sign in as a moderator
- Navigate to the review page
- Review any current pending edit
-Error pops up

# Environment
All

dakota002:
Fixes #2431 

I am working on adding some playwright tests for this page, but this should resolve the issue so that we aren't blocked in approving changes

### DIFF
--- a/app/routes/circulars.moderation.$circularId.$requestor.tsx
+++ b/app/routes/circulars.moderation.$circularId.$requestor.tsx
@@ -129,7 +129,7 @@ function DiffedContent({
 }) {
   const diff = methodMap[method && newString ? method : 'lines'](
     oldString ?? '',
-    newString ?? 'Missing, please fix'
+    newString ?? ''
   )
 
   return (
--- a/app/routes/circulars.moderation.$circularId.$requestor.tsx
+++ b/app/routes/circulars.moderation.$circularId.$requestor.tsx
@@ -127,7 +127,10 @@ function DiffedContent({
   newString: string
   method?: 'words' | 'lines'
 }) {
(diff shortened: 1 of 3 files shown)


lpsinger (commented):
It's the format field that was missing, isn't it? It's not an error for it to be missing.

dakota002:
> It's the format field that was missing, isn't it? It's not an error for it to be missing.

Also the created on field, should I remove that message?

lpsinger:
> > It's the format field that was missing, isn't it? It's not an error for it to be missing.
> 
> Also the created on field, should I remove that message?

If the field is absent in original _and_ in the new version, then of course that's not an error. If it's absent in the old version and present in the new version, then we should render that as if it was an empty string in the original. If it's present in the original but absent in the new version, well --- that should be impossible.

dakota002:
> > > It's the format field that was missing, isn't it? It's not an error for it to be missing.
> > 
> > 
> > Also the created on field, should I remove that message?
> 
> If the field is absent in original _and_ in the new version, then of course that's not an error. If it's absent in the old version and present in the new version, then we should render that as if it was an empty string in the original. If it's present in the original but absent in the new version, well --- that should be impossible.

This was actually a case of the last one you mentioned, present in the original, missing in the new because the change requests were made before the UI supported createdOn being a changeable field

lpsinger (approved):

PR #2505, opened 2024-07-31
updated synonymGroups table-stream with tests

### LINKED ISSUE #2504: Update table-stream for new synonymGroups
Update the table-stream for the new design of synonymGroups for opensearch. 

As this cannot be tested locally, include testing of adding and removing records via this table-stream.

Courey:
# Description
This change is to modify the synonymGroups table-stream to match the format we decided at our in person team meeting. It includes full test coverage. 

# Related Issue(s)
Resolves #2504 

# Testing
Table-streams are not able to run locally because the architect sandbox uses dynalite which does not support table-streams. So this has not been run. However, there are 3 tests to cover each of the three actions possible, MODIFY, INSERT, and REMOVE. 

I would appreciate someone with dev access helping me test this on the dev server to ensure that it works properly.


lpsinger (changes requested):
I think that there are some logical errors here. For example, your handler for removal of a DynamoDB record deletes the item in the OpenSearch index. However, it ought to _update_ the item in OpenSearch with the remaining event names that still correspond to the same UUID.

app/table-streams/synonyms/index.ts
```diff
@@ -12,12 +12,14 @@ import { errors } from '@opensearch-project/opensearch'
...
+  StreamRecord,
 } from 'aws-lambda'
 
 import { createTriggerHandler } from '~/lib/lambdaTrigger.server'
-import type { Synonym } from '~/routes/synonyms/synonyms.lib'
+import type { Synonym, SynonymGroup } from '~/routes/synonyms/synonyms.lib'
+import { getSynonymsByUuid } from '~/routes/synonyms/synonyms.server'
 
-const index = 'synonyms'
+const index = 'synonymGroups'
```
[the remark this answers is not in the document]
Courey: well it's not a synonym. those are structurally different than what we are putting in opensearch. I didn't want anyone to be confused by that.

lpsinger (commented):
Would you please address the remaining outstanding comments from my first review?

app/table-streams/synonyms/index.ts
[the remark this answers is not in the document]
lpsinger: Fair enough.

app/table-streams/synonyms/index.ts
```diff
@@ -34,26 +34,59 @@ async function removeIndex(id: string) {
...
 export const handler = createTriggerHandler(
   async ({ eventName, dynamodb }: DynamoDBRecord) => {
-    const id = unmarshallTrigger(dynamodb!.Keys).id as string
+    if (!eventName || !dynamodb) return
     const promises = []
-
-    if (eventName === 'REMOVE') {
+    const id = unmarshallTrigger(dynamodb!.Keys).synonymId
+    const synonym = unmarshallTrigger(dynamodb!.NewImage) as Synonym
+    const existingGroup = await getIndex(synonym)
```
lpsinger, line 66: Remember that we are treating DynamoDB as the source of truth, not OpenSearch. Please get the list of synonyms matching the UUID of the altered record using DynamoDB.

app/table-streams/synonyms/index.ts
```diff
@@ -34,26 +34,59 @@ async function removeIndex(id: string) {
...
+        eventIds: group.eventIds.filter((id) => id !== synonym.eventId),
+      } as SynonymGroup
+      promises.push(putIndex(modifiedGroup))
     } /* (eventName === 'INSERT' || eventName === 'MODIFY') */ else {
-      const synonym = unmarshallTrigger(dynamodb!.NewImage) as Synonym
-      promises.push(putIndex(synonym))
+      if (!group.eventIds.includes(synonym.eventId)) {
+        group.eventIds.push(synonym.eventId)
+      }
+      promises.push(putIndex(group as SynonymGroup))
```
lpsinger, line 88: As I said the other day, you shouldn't need to compute the terms to add or remove if you just query all of the terms that match the UUID of the DynamoDB record that triggered the Lambda.

sandbox-search.js
```diff
@@ -11,13 +11,28 @@ export default async function () {
...
   const { circulars, synonyms } = JSON.parse(text)
 
+  const groupedSynonyms = {}
+  synonyms.forEach((synonym) => {
+    if (groupedSynonyms[synonym.synonymId]) {
+      groupedSynonyms[synonym.synonymId].push(synonym.eventId)
+    } else {
+      groupedSynonyms[synonym.synonymId] = [synonym.eventId]
+    }
+  })
```
lpsinger, line 21: Is there a builtin JavaScript method or a Lodash helper to group an iterable by the value of a key function?

lpsinger (changes requested):

app/table-streams/synonyms/index.ts
```diff
@@ -34,26 +35,31 @@ async function removeIndex(id: string) {
...
 
 export const handler = createTriggerHandler(
   async ({ eventName, dynamodb }: DynamoDBRecord) => {
-    const id = unmarshallTrigger(dynamodb!.Keys).id as string
+    if (!eventName || !dynamodb) return
     const promises = []
-
-    if (eventName === 'REMOVE') {
+    const id = unmarshallTrigger(dynamodb!.Keys).synonymId
+    const synonym = unmarshallTrigger(dynamodb!.NewImage) as Synonym
```
lpsinger, line 52:
```suggestion
    const { synonymId } = unmarshallTrigger(dynamodb!.NewImage) as Synonym
```

app/table-streams/synonyms/index.ts
```diff
@@ -34,26 +35,31 @@ async function removeIndex(id: string) {
...
+    const group = {
+      synonymId: synonym.synonymId,
+      eventIds: dynamoSynonyms.map((synonym) => synonym.eventId),
+    }
+    if (eventName === 'REMOVE' && group.eventIds.length === 0) {
       promises.push(removeIndex(id))
     } /* (eventName === 'INSERT' || eventName === 'MODIFY') */ else {
-      const synonym = unmarshallTrigger(dynamodb!.NewImage) as Synonym
-      promises.push(putIndex(synonym))
+      promises.push(putIndex(group))
```
lpsinger, line 61: You can simply the above as follows:

```ts
if (dynamoSynonyms.length > 0)
{
  await putIndex({
     synonymId,
     eventIds: dynamoSynonyms.map((synonym) => synonym.eventId),
 })
} else {
  await removeIndex(synonymId)
}
```

Then you can also remove the final `await Promises.all`.

lpsinger (changes requested):

sandbox-search.js
```diff
@@ -11,13 +11,23 @@ export default async function () {
   const text = await readFile('sandbox-seed.json', { encoding: 'utf-8' })
   const { circulars, synonyms } = JSON.parse(text)
 
+  const groupedSynonyms = synonyms.reduce((accumulator, synonym) => {
```
lpsinger, line 14: I was thinking something more along the lines of https://lodash.com/docs/#groupBy.

lpsinger (changes requested):

sandbox-search.js
Courey: is there a reason or is it just preference?

lpsinger: It's terser and it's more obvious what the code does. I think it is more idiomatic. I would have suggested [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy), but it is a relatively recent addition to ECMAScript and requires a newer version of Node.js than we are targeting.

sandbox-search.js
```diff
@@ -6,16 +6,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { readFile } from 'fs/promises'
+import _ from 'lodash'
```
lpsinger, line 9: Use the following import style to match what we do elsewhere in this project:

```suggestion
import groupBy from 'lodash/groupBy'
```

This is to facilitate tree-shaking. That doesn't matter in this context, but it's good to match the same style anyway.

lpsinger (changes requested):

sandbox-search.js
```diff
@@ -6,18 +6,24 @@
...
 
 export default async function () {
   const text = await readFile('sandbox-seed.json', { encoding: 'utf-8' })
   const { circulars, synonyms } = JSON.parse(text)
+  const groupedSynonyms = _.groupBy(synonyms, 'synonymId')
+  const groups = Object.keys(groupedSynonyms).map((id) => ({
+    synonymId: id,
+    eventIds: groupedSynonyms[id],
+  }))
```
lpsinger, line 18:
```suggestion
  const groups = Object.entries(groupBy(synonyms, 'synonymId')).map(
    ([synonymId, values]) => [{synonymId, eventIds: values.map(({eventId}) => eventId)}])
```

sandbox-search.js
```diff
@@ -6,18 +6,24 @@
...
 
   return [
     ...circulars.flatMap((item) => [
       { index: { _index: 'circulars', _id: item.circularId.toString() } },
       item,
     ]),
-    ...synonyms.flatMap((item) => [
-      { index: { _index: 'synonyms', _id: item.id } },
+    ...groups.flatMap((item) => [
```
lpsinger, line 25:
```suggestion
    ...groups.map((item) => [
```

lpsinger (changes requested):

lpsinger (approved):

Issue #1266, opened 2023-08-21
Persist all new incoming Kafka records to an s3 bucket


dakota002:
Set up a bucket and utilize kafka connect to store records until we set up a long-term db solution

https://docs.confluent.io/kafka-connectors/s3-sink/current/overview.html

jracusin:
Acceptance criteria

- setup S3 bucket
- save all notices in all formats including GCN Classic (voevent, text, binary) and JSON records distributed over Kafka

PR #2531, opened 2024-08-08
Simple plugin to add an S3 bucket to the cloudformation template

dakota002:
From [these docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-s3-bucket.html) it looks like all the properties are optional, so I think this should be enough to create the new S3 bucket

### DIFF
--- a/app.arc
+++ b/app.arc
@@ -159,6 +159,9 @@ volumeSize 10
 dedicatedMasterCount 3
 dedicatedMasterType t3.small.search
 
+@storage-private
+notices
+
 @plugins
 plugin-remix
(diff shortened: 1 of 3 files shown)


lpsinger:
What about https://www.npmjs.com/package/@architect/plugin-storage-private?

dakota002:
Neat! I read over their code a bit and seems straight forward, I'll try implementing it

dakota002:
First bullet point of [1266](https://github.com/nasa-gcn/gcn.nasa.gov/issues/1266)

lpsinger:
Closing in favor of similar change on the gan-infrastructure side.

PR #2538, opened 2024-08-14
adds synonyms moderation front end

### LINKED ISSUE #2116: synonym moderation UI
Create a front end for synonyms moderation behind a feature flag. It should include the ability to create, delete, and update synonyms.

Acceptance criteria:
- a feature flag exists hiding this work
- a route exists where moderators can see an index of synonyms
- a route exists where moderators can create synonym groups
- a route exists for a detailed view of a synonym group where editing is possible
- font, styling, and spacing should be consistent with the rest of the website
- the data displayed for the group name should be the eventIds, not the uuids.

Courey:
# Description
This is the front end for the moderation portion of synonyms (not the circular index grouped by synonyms, but the ability to create, put, delete the synonyms themselves). It is behind both a feature flag and a moderator check. The majority of the code is behind a feature flag and moderation gate, but I did create a component for pagination to reduce redundancy so I could reuse it for this work. In order to do that I had to update the circulars index page as well. That portion is NOT behind a feature flag.

# Related Issue(s)
Resolves #2116

# Testing
This is the moderation index with pagination. Links for each group are to the detail page
Screenshot 2024-08-13 at 4 34 31 PM

This is the detail page for a synonym group. The button to save is disabled because no additional eventIds have been added. It is enabled when a new eventId is added.
Screenshot 2024-08-14 at 10 04 06 AM

This is the error shown if you enter an eventId that doesn't exist.
Screenshot 2024-08-14 at 2 50 48 PM

This is the create view with no eventIds added.
Screenshot 2024-08-14 at 10 05 12 AM

This is the same view with eventIds added.
Screenshot 2024-08-14 at 10 05 44 AM

This is the error shown when you try to create a synonym group with an invalid eventId
Screenshot 2024-08-14 at 2 37 14 PM

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -14,14 +14,7 @@ import {
   useSearchParams,
   useSubmit,
 } from '@remix-run/react'
-import {
-  Alert,
-  Button,
-  Icon,
-  Label,
-  Select,
-  TextInput,
-} from '@trussworks/react-uswds'
+import { Alert, Button, Icon, Label, TextInput } from '@trussworks/react-uswds'
 import clamp from 'lodash/clamp'
 import { useId, useState } from 'react'
 
@@ -41,13 +34,13 @@ import {
   putVersion,
   search,
 } from '../circulars/circulars.server'
-import CircularPagination from './CircularPagination'
 import CircularsHeader from './CircularsHeader'
 import CircularsIndex from './CircularsIndex'
 import { DateSelector } from './DateSelectorMenu'
 import { SortSelector } from './SortSelectorButton'
 import Hint from '~/components/Hint'
 import { ToolbarButtonGroup } from '~/components/ToolbarButtonGroup'
+import PaginationSelectionFooter from '~/components/pagination/PaginationSelectionFooter'
 import { origin } from '~/lib/env.server'
 import { getFormDataString } from '~/lib/utils'
 import { postZendeskRequest } from '~/lib/zendesk.server'
@@ -281,40 +274,13 @@ export default function () {
             totalItems={totalItems}
             query={query}
           />
-          <div className="display-flex flex-row flex-wrap">
-            <div className="display-flex flex-align-self-center margin-right-2 width-auto">
-              <div>
-                <Select
-                  id="limit"
-                  title="Number of results per page"
-                  className="width-auto height-5 padding-y-0 margin-y-0"
-                  name="limit"
-                  defaultValue="100"
-                  form={formId}
-                  onChange={({ target: { form } }) => {
-                    submit(form)
-                  }}
-                >
-                  <option value="10">10 / page</option>
-                  <option value="20">20 / page</option>
-                  <option value="50">50 / page</option>
-                  <option value="100">100 / page</option>
-                </Select>
-              </div>
-            </div>
-            <div className="display-flex flex-fill">
-              {totalPages > 1 && (
-                <CircularPagination
-                  query={query}
-                  page={page}
-                  limit={parseInt(limit)}
-                  totalPages={totalPages}
-                  startDate={startDate}
-                  endDate={endDate}
-                />
-              )}
-            </div>
-          </div>
+          <PaginationSelectionFooter
+            query={query}
+            page={page}
+            limit={parseInt(limit)}
+            totalPages={totalPages}
+            form={formId}
+          />
         </>
       )}
     </>
--- a/app/routes/synonyms.$synonymId.tsx
+++ b/app/routes/synonyms.$synonymId.tsx
@@ -7,25 +7,62 @@
  */
 import type { ActionFunctionArgs, LoaderFunctionArgs } from '@remix-run/node'
 import { redirect } from '@remix-run/node'
-import { Form, Link, useLoaderData } from '@remix-run/react'
-import { Button, ButtonGroup, FormGroup, Icon } from '@trussworks/react-uswds'
-import { useState } from 'react'
-import invariant from 'tiny-invariant'
-
 import {
+  Form,
+  Link,
+  useFetcher,
+  useLoaderData,
+  useSubmit,
+} from '@remix-run/react'
+import type { ModalRef } from '@trussworks/react-uswds'
+import {
+  Button,
+  ButtonGroup,
+  CardBody,
+  FormGroup,
+  Grid,
+  Icon,
+  Modal,
+  ModalFooter,
+  ModalHeading,
+  ModalToggleButton,
+  TextInput,
+} from '@trussworks/react-uswds'
+import { useEffect, useRef, useState } from 'react'
+import invariant from 'tiny-invariant'
+import { useOnClickOutside } from 'usehooks-ts'
+
+import { getUser } from './_auth/user.server'
+import { moderatorGroup } from './circulars/circulars.server'
+import {
+  autoCompleteEventIds,
   deleteSynonyms,
   getSynonymsByUuid,
   putSynonyms,
 } from './synonyms/synonyms.server'
+import DetailsDropdownContent from '~/components/DetailsDropdownContent'
+import { ToolbarButtonGroup } from '~/components/ToolbarButtonGroup'
 import { getFormDataString } from '~/lib/utils'
 
-export async function loader({ params: { synonymId } }: LoaderFunctionArgs) {
+export async function loader({
+  request,
+  params: { synonymId },
+}: LoaderFunctionArgs) {
+  const user = await getUser(request)
+  if (!user?.groups.includes(moderatorGroup))
+    throw new Response(null, { status: 403 })
+
   invariant(synonymId)
   const synonyms = await getSynonymsByUuid(synonymId)
   const eventIds = synonyms.map((synonym) => synonym.eventId)
-
+  const url = new URL(request.url)
+  const query = url.searchParams.get('query')
+  const { options } = query
+    ? await autoCompleteEventIds({ query })
+    : { options: [] }
   return {
     eventIds,
+    options,
   }
 }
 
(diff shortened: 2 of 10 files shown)


Courey:
> When you delete synonyms, does it double check you meant to delete them? If not, it would be good to do so.

It does not currently, but that is definitely something I can add in!

Courey:
Here are updated images to show the additional request for a modal making sure you want to delete a synonym group.
I also updated the trashcan icon next to each eventId so that it says remove instead of just uses an icon.
Screenshot 2024-08-16 at 11 41 34 AM
Screenshot 2024-08-16 at 11 41 56 AM
Screenshot 2024-08-16 at 11 44 38 AM
Screenshot 2024-08-16 at 11 44 56 AM

app/routes/synonyms.new.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { useId, useState } from 'react'
+
+import { createSynonyms } from './synonyms/synonyms.server'
+import { ToolbarButtonGroup } from '~/components/ToolbarButtonGroup'
+import { getFormDataString } from '~/lib/utils'
+
+export async function action({ request }: ActionFunctionArgs) {
+  const data = await request.formData()
+  const eventIds = getFormDataString(data, 'synonyms')?.split(',')
+  if (!eventIds) return 'Must include at least 1 eventId.'
```
dakota002, line 21:
```suggestion
  if (!eventIds) throw new Response(null, { status: 400 })
```

app/routes/synonyms.new.tsx
```diff
@@ -0,0 +1,110 @@
...
+  return (
+    <>
+      <h1>Create New Synonym Group</h1>
+      {error && (
+        <Alert type="error" heading="Error:" headingLevel="h4">
+          {error}
+        </Alert>
+      )}
+      <ToolbarButtonGroup className="position-sticky top-0 bg-white margin-bottom-1 padding-top-1 z-300">
+        <Form preventScrollReset className="" id={formId}>
```
dakota002, line 46:
```suggestion
        <Form preventScrollReset id={formId}>
```

app/routes/synonyms/route.tsx
```diff
@@ -10,12 +10,14 @@ import { Outlet } from '@remix-run/react'
 import { GridContainer } from '@trussworks/react-uswds'
 
 import { getUser } from '../_auth/user.server'
+import { feature } from '~/lib/env.server'
 
 export async function loader({ request }: LoaderFunctionArgs) {
   const user = await getUser(request)
   const isModerator = user?.groups.includes('gcn.nasa.gov/circular-moderator')
+  const featureFlagOn = feature('SYNONYMS')
```
dakota002, line 18:
```suggestion
```

app/routes/synonyms/route.tsx
```diff
@@ -10,12 +10,14 @@ import { Outlet } from '@remix-run/react'
...
 
-  if (!isModerator) {
+  if (!isModerator || !featureFlagOn) {
```
dakota002, line 20:
```suggestion
  if (!isModerator || !feature('SYNONYMS')) {
```
No need to define as a const

app/routes/synonyms/synonyms.lib.ts
```diff
@@ -17,3 +17,14 @@ export interface SynonymGroup {
   synonymId: string
   eventIds: string[]
 }
+
+export interface SynonymValidityCheck {
+  eventId: string
+  count: number
+}
```
dakota002, line 24: Looks like this is unused

```suggestion
```

dakota002 (changes requested):

app/routes/synonyms._index.tsx
lpsinger, line 1: Add copyright/license banner.

app/routes/circulars._archive._index/PaginationSelectionFooter.tsx
```diff
@@ -0,0 +1,54 @@
...
+  totalPages,
+  limit,
+  query,
+  formId,
+}: {
+  page: number
+  totalPages: number
+  limit?: number
+  query?: string
+  formId: string
```
lpsinger, line 17:
```suggestion
  form: string
```

to match the name of the prop on `<input>` elements.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -40,10 +33,10 @@ import {
   putVersion,
   search,
 } from '../circulars/circulars.server'
-import CircularPagination from './CircularPagination'
```
lpsinger, line 43: This module is unused now, isn't it? Please refactor to eliminate dead code.

app/routes/synonyms/route.tsx
```diff
@@ -10,12 +10,13 @@ import { Outlet } from '@remix-run/react'
...
 
 import { getUser } from '../_auth/user.server'
+import { feature } from '~/lib/env.server'
 
 export async function loader({ request }: LoaderFunctionArgs) {
   const user = await getUser(request)
   const isModerator = user?.groups.includes('gcn.nasa.gov/circular-moderator')
 
-  if (!isModerator) {
+  if (!isModerator || !feature('SYNONYMS')) {
```
lpsinger, line 19: We can probably do without this feature flag check here because the synonyms interface is only visible to moderators.

app/routes/synonyms.$synonymId.tsx
```diff
@@ -44,12 +56,12 @@ export async function action({
...
       getFormDataString(data, 'deleteSynonyms')?.split(',') || ([] as string[])
     const filtered_subtractions = subtractions.filter((sub) => sub)
-    await putSynonyms({
+    const response = await putSynonyms({
       synonymId,
       additions: filtered_additions,
       subtractions: filtered_subtractions,
     })
-    return null
+    return response?.error
```
lpsinger, line 64: Generally the front-end code should be written to assume that the code takes the happy path. Use an [error boundary](https://remix.run/docs/en/main/route/error-boundary) if you want to fine-tune how errors are presented.

```suggestion
    return null
```

app/routes/synonyms/synonyms.server.ts
```diff
@@ -148,7 +181,19 @@ export async function putSynonyms({
...
+    error: null,
+  } as PutSynonymResponse
   if (!subtractions?.length && !additions?.length) return
+  if (additions?.length) {
+    const isValid = await validateEventIds({ eventIds: additions })
+    if (!isValid) {
+      response.error = 'Invalid eventId'
+      return response
+    }
+  }
```
lpsinger, line 196: Raise an exception if the input is invalid. If you want to be RESTfully correct, then raise a 400 Bad Request error.

app/routes/synonyms/synonyms.server.ts
```diff
@@ -175,6 +220,8 @@ export async function putSynonyms({
     },
   }
   await client.batchWrite(params)
+  response.success = true
+  return response
```
lpsinger, line 224: The function should succeed and return, or fail and never return (i.e., raise an error).

lpsinger (changes requested):
I don't see anything in the action functions that checks that the user is authorized.

app/routes/synonyms/synonyms.server.ts
Courey: You do not want the banner that Judy requested? I believe the request was for a banner to alert the moderator that an eventId was incorrect so they could check the input and re-enter it. I could just make it 400 but in that case it would not have the banner telling the user why it failed.
Screenshot 2024-08-16 at 11 41 56 AM

app/routes/circulars._archive._index/PaginationSelectionFooter.tsx
Courey: I am happy to change it. I did it this way for consistency as you originally did it here https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/circulars._archive._index/route.tsx#L190 
Would you like me to also change that so that it's consistent across routes? or since the elements the form id is being passed to are different, is there not a need for that level of consistency in variable names?

app/routes/synonyms/synonyms.server.ts
lpsinger: What does it mean for an eventId to be invalid?

Courey: that it does not appear as the eventId on any circular in the database. In the above image, all of the eventIds exist except "nope". Since "nope" was not found to be an eventId on any circular, that eventId is invalid. I can remove the validation so that a moderator could put anything, but I was worried that in that particular scenario they might add something they thought was correct but was mistyped. So what I was trying to accomplish there was a "hey, you entered an eventId that doesn't exist anywhere, that means what you think you linked to the group would not include any circulars. check over these ids below and see if you entered something wrong and then try saving again."

Courey: If you just want it to throw an error response, I can try to make that informative in a way that gives some indication of what went wrong, but I haven't explored those options yet, so I am not sure yet what that would look like.

Courey: I can also adapt the message so that it is much clearer what an invalid eventId is

app/routes/circulars._archive._index/PaginationSelectionFooter.tsx
lpsinger: > or since the elements the form id is being passed to are different, is there not a need for that level of consistency in variable names?

That one.

app/routes/synonyms/synonyms.server.ts
lpsinger: > that it does not appear as the eventId on any circular in the database. In the above image, all of the eventIds exist except "nope". Since "nope" was not found to be an eventId on any circular, that eventId is invalid. I can remove the validation so that a moderator could put anything, but I was worried that in that particular scenario they might add something they thought was correct but was mistyped. So what I was trying to accomplish there was a "hey, you entered an eventId that doesn't exist anywhere, that means what you think you linked to the group would not include any circulars. check over these ids below and see if you entered something wrong and then try saving again."

How about we have a combo box that auto-completes from existing event IDs, and the submit button is disabled until you choose a valid one?

@dakota002 may have refactored our combo box code to the point that it is reasonably reusable.

Courey: I can definitely look into that.

app/routes/circulars._archive._index/route.tsx
Courey: I'm not sure what you mean here? It is used. I moved it into components>pagination>GCNPagination.tsx and it is used in the PaginationFooter which is also moved to that same directory.

Courey:
With nothing on the page yet
Screenshot 2024-09-13 at 1 09 56 PM

when you type 3 or more characters
Screenshot 2024-09-13 at 1 10 33 PM

When you hover over one of the options
Screenshot 2024-09-13 at 1 10 43 PM

lpsinger (commented):
Sorry this took me so long to get to. There are some merge conflicts. Would you please rebase from main?

Squashing your existing commits before rebasing might make it easier.

app/routes/synonyms.$synonymId.tsx
```diff
@@ -61,81 +98,103 @@ export async function action({
...
 
 export default function () {
-  const { eventIds } = useLoaderData<typeof loader>()
+  const { eventIds, options } = useLoaderData<typeof loader>()
+  const uniqueOptions = Array.from(new Set(options)) as string[]
   const [deleteSynonyms, setDeleteSynonyms] = useState([] as string[])
   const [synonyms, setSynonyms] = useState(eventIds || [])
   const [addSynonyms, setAddSynonyms] = useState([] as string[])
-  const [newSynonym, setNewSynonym] = useState('')
+  const uniqueSynonyms = Array.from(new Set(synonyms)) as string[]
```
lpsinger, line 106: Under what circumstances would the synonyms fail to be unique? I think our data model guarantees that they will be unique.

app/routes/synonyms.$synonymId.tsx
```diff
@@ -61,81 +98,103 @@ export async function action({
 }
 
 export default function () {
-  const { eventIds } = useLoaderData<typeof loader>()
+  const { eventIds, options } = useLoaderData<typeof loader>()
+  const uniqueOptions = Array.from(new Set(options)) as string[]
   const [deleteSynonyms, setDeleteSynonyms] = useState([] as string[])
   const [synonyms, setSynonyms] = useState(eventIds || [])
   const [addSynonyms, setAddSynonyms] = useState([] as string[])
```
lpsinger, line 105: Why are all of these type assertions (`X as Y`) necessary? Why is it failing to infer types?

app/routes/synonyms._index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import type { SynonymGroup } from './synonyms/synonyms.lib'
+import { searchSynonymsByEventId } from './synonyms/synonyms.server'
+import { ToolbarButtonGroup } from '~/components/ToolbarButtonGroup'
+import PaginationSelectionFooter from '~/components/pagination/PaginationSelectionFooter'
+
+import searchImg from 'nasawds/src/img/usa-icons-bg/search--white.svg'
+
+export async function loader({ request: { url } }: LoaderFunctionArgs) {
+  const { searchParams } = new URL(url)
```
lpsinger, line 34: Probably should check that the user is a moderator in the loader here.

app/routes/synonyms.$synonymId.tsx
```diff
@@ -148,7 +207,7 @@ export default function () {
           <input type="hidden" name="deleteSynonyms" value={deleteSynonyms} />
           <input type="hidden" name="addSynonyms" value={addSynonyms} />
```
lpsinger, line 208: Since these state variables are both sets, I think you can just emit an input for each of their members, for easier deserialization.

According to (for example) https://webmasters.stackexchange.com/questions/9775/is-it-valid-to-have-multiple-form-input-elements-with-the-same-name, you should get all of the values of identically-named inputs in the form data.

lpsinger (changes requested):

app/routes/synonyms._index.tsx
Courey: That is done at the parent route level [here](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/synonyms/route.tsx#L18). I'm happy to also add it here, but I don't think it's necessary because the entire synonyms route is mod gated.

lpsinger: I think it is necessary. As far as I can recall, Remix will request the loader data for all of the ancestor routes in parallel, so authorization-dependent code _must_ be checked even in descendent routes.

app/routes/synonyms.$synonymId.tsx
Courey: Example: If the user inputs GRB 230919A and then using the input again searches and selects 230919A, only 1 230919A is added to the synonym group ul. This prevents showing multiples of the same event id like this:
Screenshot 2024-10-10 at 10 14 03 AM

With the uniqueness check, it only displays one of each synonym even when added multiple times:
Screenshot 2024-10-10 at 10 15 44 AM

Either way, it will only be sent to the server once, so it won't break anything. It could just be confusing if someone added it twice and expected there to be more list items upon refresh than it has since it will remove the duplications.

app/routes/synonyms.$synonymId.tsx
Courey: That is because of a quirk in how react useState behaves. [Here is a video](https://www.totaltypescript.com/tips/avoid-unexpected-behaviour-of-react-s-usestate) that shows the exact case. If you pass useState an empty array, it infers that the type is a never[]. So if you try to add anything to it, you can't because it's a never[]. 
Screenshot 2024-10-10 at 10 27 56 AM
You can solve for the issue by using useState<string[]>([]) or by casting the array to a string array like I did here. I'll go ahead and change it to to use useState<string[]>([]) since you probably prefer the terseness and not casting.

app/routes/synonyms.$synonymId.tsx
Courey: I'm not sure I fully understand what you are asking for here. 
Do you mean doing something like this?
```
{addSynonyms.map((synonym)=>(<input key={synonym} type="hidden" name="addSynonyms" value={synonym} id={`add-${synonym}`}/>))}
{deleteSynonyms.map((synonym)=>(<input key={synonym} type="hidden" name="deleteSynonyms" value={synonym} id={`del-${synonym}`}/>))}
```
If so, that does not work, it just sends the first value and does not return an array of multiple values.

Courey: Starting state:
Screenshot 2024-10-10 at 11 21 40 AM

Two removed, two added:
Screenshot 2024-10-10 at 11 23 12 AM

The code where I am logging in the action:
Screenshot 2024-10-10 at 11 23 36 AM

The terminal output:
Screenshot 2024-10-10 at 11 23 23 AM

app/routes/synonyms.$synonymId.tsx
lpsinger: Yes, I prefer that. Better than overriding TypeScript. Thanks!

app/routes/synonyms.$synonymId.tsx
lpsinger: That might be a limitation of our own function, `getFormDataString`. I'm going to do a little experiment with my browser. Will report back in a moment...

lpsinger: Take the following little HTML page, `test.html`:

```html
<!DOCTYPE html>
<form>
  <input name="foo" value="bar">
  <input name="foo" value="bat">
  <input type="submit">
</form>
```

Open it in a browser and click the submit button. The address bar updates to `file:///.../test.html?foo=bar&foo=bat`, showing that at least for URL-encoded form submission, both values are included.

lpsinger (approved):

PR #2546, opened 2024-08-21
Updates the change request functions to automatically close the corresponding zendesk ticket

dakota002:
# Description
Adds a function to the Zendesk server file to close tickets by Id, and calls the function in both the approve and reject paths

Also fixes a typo

# Testing
- As a circular submitter, create a change request
- Check zendesk to see that the ticket was created
- As a moderator either approve or reject the request
- Go back to Zendesk and see the ticket status updated to "solved"

### DIFF
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -72,5 +72,4 @@ export async function closeZendeskTicket(ticketId: number) {
     console.error(response)
     throw new Error(`Request failed with status ${response.status}`)
   }
-  return await response.json()
 }
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -43,6 +43,34 @@ export async function postZendeskRequest(request: ZendeskRequest) {
 
   if (!response.ok) {
     console.error(response)
-    throw new Error(`Reqeust failed with status ${response.status}`)
+    throw new Error(`Request failed with status ${response.status}`)
   }
+  return await response.json()
+}
+
+export async function closeZendeskTicket(ticketId: number) {
+  const response = await fetch(
+    `https://nasa-gcn.zendesk.com/api/v2/tickets/${ticketId}.json`,
+    {
+      method: 'PUT',
+      headers: {
+        'Content-Type': 'application/json',
+        ...getBasicAuthHeaders(
+          `${getEnvOrDie('ZENDESK_TOKEN_EMAIL')}/token`,
+          getEnvOrDie('ZENDESK_TOKEN')
+        ),
+      },
+      body: JSON.stringify({
+        ticket: {
+          status: 'solved',
+        },
+      }),
+    }
+  )
+
+  if (!response.ok) {
+    console.error(response)
+    throw new Error(`Request failed with status ${response.status}`)
+  }
+  return await response.json()
 }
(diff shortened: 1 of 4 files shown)


jracusin:
Let's only automatically close zendesk tickets when accepted, not rejected, because that'll need more explanation.

app/lib/zendesk.server.ts
```diff
@@ -43,6 +43,34 @@ export async function postZendeskRequest(request: ZendeskRequest) {
...
+        },
+      }),
+    }
+  )
+
+  if (!response.ok) {
+    console.error(response)
+    throw new Error(`Request failed with status ${response.status}`)
+  }
+  return await response.json()
```
lpsinger, line 75: We don't really use the response, do we?

```suggestion
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -110,22 +110,25 @@ export async function action({ request }: ActionFunctionArgs) {
...
+        },
+      })
+
       await createChangeRequest(
         {
           circularId: parseFloat(circularId),
           ...props,
           submitter,
           createdOn,
+          zendeskTicketId: zendeskResponse.request.id,
```
lpsinger, line 128:
```suggestion
      const {request: {id: zendeskTicketId}} = await postZendeskRequest({
        requester: { name: user.name, email: user.email },
        subject: `Change Request for Circular ${circularId}`,
        comment: {
          body: `${user.name} has requested an edit. Review at ${origin}/circulars`,
        },
      })

      await createChangeRequest(
        {
          circularId: parseFloat(circularId),
          ...props,
          submitter,
          createdOn,
          zendeskTicketId,
```

app/routes/circulars/circulars.server.ts
```diff
@@ -539,7 +539,8 @@ export async function approveChangeRequest(
   })
 
   await deleteChangeRequestRaw(circularId, requestorSub)
-
+  if (changeRequest.zendeskTicketId)
+    await closeZendeskTicket(changeRequest.zendeskTicketId)
```
lpsinger, line 543: The only reason that we would need this is to handle change requests that pre-date the deployment of this PR, right?
```suggestion
    await closeZendeskTicket(changeRequest.zendeskTicketId)
```

Also, can we do some of these steps concurrently (`Promise.all`)?

lpsinger (changes requested):

app/lib/zendesk.server.ts
dakota002: Nope, this one can be removed

app/routes/circulars/circulars.server.ts
dakota002: Yup, and I can check on the concurrent thing. I assume probably

app/routes/circulars/circulars.lib.ts
```diff
@@ -39,6 +39,7 @@ export interface CircularChangeRequest extends CircularMetadata {
   format: CircularFormat
   submitter: string
   createdOn: number
+  zendeskTicketId?: number
```
lpsinger, line 42:
```suggestion
  zendeskTicketId: number
```

lpsinger (changes requested):
I think you may have inadvertently added the Zendesk ticket ID to the GCN Circular.

dakota002:
> I think you may have inadvertently added the Zendesk ticket ID to the GCN Circular.

I don't believe so, where do you see this?

lpsinger (approved):

PR #2548, opened 2024-08-22
Adds some minor validation to catch html in the contact form fields

dakota002:
This will hopefully reduce the spam tickets we receive 

@lpsinger @Courey Please let me know if there are additional options you can think of to catch other potentially malicious requests

### DIFF
--- a/app/routes/contact.tsx
+++ b/app/routes/contact.tsx
@@ -42,6 +42,11 @@ export async function loader() {
 export const headers: HeadersFunction = ({ loaderHeaders }) =>
   pickHeaders(loaderHeaders, ['Link'])
 
+function containsHtml(input: string): boolean {
+  const htmlRegex = /<[a-z/][\s\S]*>/g
+  return htmlRegex.test(input)
+}
+
 export async function action({ request }: ActionFunctionArgs) {
   const data = await request.formData()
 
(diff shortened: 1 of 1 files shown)


lpsinger:
I don't know about this. Legitimate support requests could contain markup.

dakota002:
Hm that is fair. What are your thoughts on updating it so that at least the checks against the subject, name, and email are not containing any, and a server side update that the email is address is correctly formatted? 

Also are there any different checks you think we should add?

lpsinger (commented):
I am concerned that this might exclude legitimate content, especially content that contains sample XML content from VOEvents.

dakota002:
Right, what are your thoughts on my previous comment though? Email Name and Subject fields only then?

lpsinger:
> Right, what are your thoughts on my previous comment though? Email Name and Subject fields only then?

I can imagine that the subject field might also legitimately contain markup.

dakota002:
Gotcha, what is your preference for this PR? Close out, or some changes?

lpsinger:
I'm looking at the names on the tickets that we marked as spam, and this pattern would catch only some of them. It sure would be nice to have some spam filtering for tickets opened through the web form, but I think it would have to be a bit more sophisticated than a single regex.

Issue #2562, opened 2024-09-03
Add optional email/kafka distribution for Circular revisions


jracusin:
# Description

When authors request a major change to a Circular (e.g. trigger time or magnitudes/flux change), give moderator option to distribute new version via email/Kafka.

# Acceptance criteria

- add Circular JSON schema revision number as field
- Moderator button to optionally distribute new Circular version over email and Kafka
- Add reminder to moderators of what constitutes a major change
- update documentation to say that we may distribute a new version if there's a significant data change

lpsinger:
The Kafka part of this is dead simple: I think that we should _always_ send revised Circulars over Kafka. https://github.com/nasa-gcn/gcn.nasa.gov/pull/2568

dakota002:
@lpsinger and @jracusin can you revise the acceptance criteria for this ticket? If we are going to always send the revised circular, what changes are left to make?

lpsinger:
Are we? I think it is clear that we should always send the Kafka record. I am less sure if we always want to send the email.

dakota002:
Ah I see, I thought by your update we were just going to always send the messages either way. All good then

dakota002:
@jracusin @lpsinger for the reminder to the mods, is something like this what y'all have in mind
![image](https://github.com/user-attachments/assets/bc00da1c-f4d9-4a45-819a-727e010a0657)

lpsinger:
> @jracusin @lpsinger for the reminder to the mods, is something like this what y'all have in mind?

No. If there is a checkbox, then the documentation about it should be in the label for the checkbox, not at the top of the page.

dakota002:
Gotcha, as for the changes themselves, what are they?

lpsinger:
None as far as I can tell. Just that detailed instructions related to a field in a form should usually go in or near the label for the control itself. Here's an example:

![Screenshot 2024-09-06 at 15 13 01](https://github.com/user-attachments/assets/ad842085-8dac-4e83-8cad-9378cd76d1c0)

PR #2568, opened 2024-09-04
Always send new versions of GCN Circulars over Kafka

lpsinger:
See #2562.


dakota002:
Does this include the version as a field in the kafka message? Or some other way to denote that it is an update and that the user's know they aren't receiving a duplicate of a previous message?

lpsinger:
> Does this include the version as a field in the kafka message?

Yes.

dakota002 (approved):

Issue #189, opened 2022-07-25
Implement Kafka to email bridge


lpsinger:
1. [x] Decide on implementation strategy:
    * a Docker container that is always running, subscribing to all public topics, and making AWS API calls? 
    * use an existing AWS service to bridge Confluent Kafka to something that can trigger Lambda invocations?
    * bridge directly to something that can trigger email notifications? (e.g. Kafka -> SQS -> SNS -> email)
2. [x] Subscribe to necessary Kafka topics
3. [x] Read subscription data
4. [x] Send emails

Depends on #185, #186, #188.

dakota002:
https://github.com/nasa-gcn/gcn-email-service

I have spent some time setting this up, it works locally with hard coded values, but I would like to get it out on dev along with the front end update for the subscriptions

lpsinger:
Renamed to: https://github.com/nasa-gcn/gcn-email

Let's keep this open until everything is set for deployment to prod.

PR #2584, opened 2024-09-16
Address issue raised on gcn-schema #189

jessiethw:
Clarify the case for when the "most_probable_direction" key is included in the lvk_nu_track_search follow up GCN, raised in nasa-gcn/gcn-schema#189

The most probable direction is included when the p-value for the generic transient search is <10%. This PR updates the IceCube docs to reflect this case.


lpsinger (approved):
@jessiethw, this looks good to me. Can you recommend someone as a reviewer? Would @blaufuss be a good choice?

jessiethw:
> @jessiethw, this looks good to me. Can you recommend someone as a reviewer? Would @blaufuss be a good choice?

thanks, and yes that would be great!

blaufuss (approved):
Thanks @jessiethw

PR #2585, opened 2024-09-17
Updates the change request form to populate with previously requested changes if they exist

dakota002:
Currently requesting sequential changes for a circular will replace whatever request already exists with the new changes. 

This pulls that info in now so if a user is to say submit a request to update the subject, then come back and later update the body before their request is approved, their request will now have both changes and be tied to the same zendesk ticket

### DIFF
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -45,7 +45,13 @@ export async function postZendeskRequest(request: ZendeskRequest) {
     console.error(response)
     throw new Error(`Request failed with status ${response.status}`)
   }
-  return await response.json()
+
+  const responseJson = await response.json()
+  if (!responseJson.request.id) {
+    throw new Response(null, { status: 500 })
+  }
+
+  return responseJson.request.id
 }
 
 export async function closeZendeskTicket(ticketId: number) {
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -49,7 +49,9 @@ export async function postZendeskRequest(request: ZendeskRequest) {
   const responseJson = await response.json()
   if (!responseJson.request.id) {
     console.error(responseJson)
-    throw new Error("ZenDesk request succeeded, but did not return a request ID")
+    throw new Error(
+      'ZenDesk request succeeded, but did not return a request ID'
+    )
   }
 
   return responseJson.request.id
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -48,7 +48,8 @@ export async function postZendeskRequest(request: ZendeskRequest) {
 
   const responseJson = await response.json()
   if (!responseJson.request.id) {
-    throw new Response(null, { status: 500 })
+    console.error(responseJson)
+    throw new Error("ZenDesk request succeeded, but did not return a request ID")
   }
 
   return responseJson.request.id
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -112,14 +112,15 @@ export async function action({ request }: ActionFunctionArgs) {
       if (!createdOnDate || !createdOn)
         throw new Response(null, { status: 400 })
 
-      let zendeskTicketId
+      let zendeskTicketId: number | undefined
 
       try {
         zendeskTicketId = (
           await getChangeRequest(parseFloat(circularId), user.sub)
         ).zendeskTicketId
       } catch (error) {
-        console.info('No existing request found')
+        const err = error as Response
+        if (err.status !== 404) throw err
       }
 
       if (!zendeskTicketId) {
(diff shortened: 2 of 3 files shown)


app/routes/circulars._archive._index/route.tsx
```diff
@@ -111,23 +112,36 @@ export async function action({ request }: ActionFunctionArgs) {
...
-      })
+      let zendeskTicketId
+
+      try {
+        zendeskTicketId = (
+          await getChangeRequest(parseFloat(circularId), user.sub)
+        ).zendeskTicketId
+      } catch (error) {
+        console.info('No existing request found')
+      }
```
lpsinger, line 123: - Catch 404 error (not found) and don't log anything.
- For any other error, just raise it.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -111,23 +112,36 @@ export async function action({ request }: ActionFunctionArgs) {
...
+          request: { id: zendeskTicketId },
+        } = await postZendeskRequest({
+          requester: { name: user.name, email: user.email },
+          subject: `Change Request for Circular ${circularId}`,
+          comment: {
+            body: `${user.name} has requested an edit. Review at ${origin}/circulars`,
+          },
+        }))
+        if (!zendeskTicketId) throw new Response(null, { status: 500 })
+      }
```
lpsinger, line 136: Can `postZendeskRequest` even return without raising an exception if it fails to create a ticket?

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
dakota002: I'm not sure, this is more to resolve the type issue of it being either a number or undefined

dakota002: This is all it returns if it fails:

```js
  if (!response.ok) {
    console.error(response)
    throw new Error(`Request failed with status ${response.status}`)
  }
```

This check is just to resolve the type issue of it being potentially undefined

app/routes/circulars.correction.$circularId.tsx
```diff
@@ -37,7 +37,8 @@ export async function loader({
   try {
     existingRequest = await getChangeRequest(parseFloat(circularId), user.sub)
   } catch (error) {
-    console.info('No existing request found')
+    const err = error as Response
+    if (err.status !== 404) throw err
```
lpsinger, line 41: Avoid type assertions. Use type narrowing wherever possible.

```suggestion
    if (!(err instanceof Response && err.status === 404)) throw err
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -112,14 +112,15 @@ export async function action({ request }: ActionFunctionArgs) {
...
+      let zendeskTicketId: number | undefined
 
       try {
         zendeskTicketId = (
           await getChangeRequest(parseFloat(circularId), user.sub)
         ).zendeskTicketId
       } catch (error) {
-        console.info('No existing request found')
+        const err = error as Response
+        if (err.status !== 404) throw err
```
lpsinger, line 123: Use type narrowing, not type assertions (see same comment below).

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
lpsinger: Could we have `postZendeskRequest` raise an exception if it does not return a new ticket ID?

lpsinger (changes requested):

dakota002:
Sure! I'll get that in the next commit

dakota002:
Sure, I can update it in the next commit

dakota002:
@lpsinger Also, do you think it makes more sense that this method should then just return the Id?

lpsinger:
> @lpsinger Also, do you think it makes more sense that this method should then just return the Id?

Good idea.

app/lib/zendesk.server.ts
```diff
@@ -45,7 +45,13 @@ export async function postZendeskRequest(request: ZendeskRequest) {
     console.error(response)
     throw new Error(`Request failed with status ${response.status}`)
   }
-  return await response.json()
+
+  const responseJson = await response.json()
+  if (!responseJson.request.id) {
+    throw new Response(null, { status: 500 })
```
lpsinger, line 51:
```suggestion
    console.error(responseJson)
    throw new Error("ZenDesk request succeeded, but did not return a request ID")
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -111,15 +112,26 @@ export async function action({ request }: ActionFunctionArgs) {
...
+
+      if (!zendeskTicketId) {
+        zendeskTicketId = await postZendeskRequest({
+          requester: { name: user.name, email: user.email },
+          subject: `Change Request for Circular ${circularId}`,
+          comment: {
+            body: `${user.name} has requested an edit. Review at ${origin}/circulars`,
+          },
+        })
+        if (!zendeskTicketId) throw new Response(null, { status: 500 })
```
lpsinger, line 133: This line cannot be reached.

```suggestion
```

lpsinger (changes requested):

lpsinger (approved):

PR #2587, opened 2024-09-18
Set IceCube notice field names in a code font

lpsinger:


lpsinger:
@blaufuss, would you please review?

blaufuss (approved):
thanks for the cleanup!

Issue #2288, opened 2024-05-20
Circulars: add support for Lucene-style search queries in archive


tylerbarna:
# Description
Lucene style search queries allow for users to conduct more granular searches and are supported by [OpenSearch](https://opensearch.org/docs/latest/query-dsl/full-text/index/)

# Acceptance criteria

- [x] Change OpenSearch backend to make use of the full-text/lucene query methods
- [x] Allow for searching with wildcards
- [x] Allow for searching by author
- [x] Allow for specifying a search of the body and/or the subject
- [x] Current default search behaviour should not be affected by lucene queries.
- [x] Provide documentation on supported lucene query options. The documentation in [ADS](https://ui.adsabs.harvard.edu/) is a good example
- [x] (optional) Implement a quick field system along the lines of ADS
- [ ] (optional) Have method for searching by date range using lucene style query
- [ ] (optional) Ensure the lucene style date range updates the state of the date range element UI

tylerbarna:
Upon further consideration, I don't think the addition of the date range filtering into the lucene search is necessary for the feature to be delivered. This would require refactoring the structure of our search function, which effectively conducts a search and then filters the results of that search based on the date range input. It's unlikely that most users will need/want to manually input these dates as part of a search query rather than using the existing date filter. Additionally, there's the concern of date format standardization (eg some users will use MM-DD-YYYY while others will use DD-MM-YYYY). This adds another area where user inputs could result in an error, which would trigger the fallback to `multi_match`, which does require the existing date filtering.

This could be something to address in the future, but the implementation of Lucene search in recent PRs meets the requirements we outlined.

PR #2589, opened 2024-09-18
Circulars: Lucene Shortcuts and Archive Page Documentation

tylerbarna:
# Description
This adds a brief explanation of Lucene queries to the Circulars Archive page as well as shortcut buttons to populate the search bar with common Lucene search syntax 

# Related Issue(s)

Tracked by #2288 

# Testing
1. Navigate to Circulars Archive
2. Click on the "Advanced Search" (Tentative Label) Accordion Menu
3. The menu should expand
4. Any Shortcut buttons should be clickable, resulting in the addition of the selected lucene query to the query. This should be reflected directly in the search bar and should not clear the existing query

### DIFF
--- a/app/routes/circulars._archive._index/CircularsLuceneMenu.tsx
+++ b/app/routes/circulars._archive._index/CircularsLuceneMenu.tsx
@@ -23,9 +23,23 @@ export function LuceneAccordion({
           Lucene Search Syntax Page
         </Link>
         {'. '}
+        <div>Lucene Examples</div>
         <div>
-          <Button type="button" onClick={() => populateSearch('test')}>
-            child
+          <Button
+            type="button"
+            onClick={() => populateSearch('subject:"SWIFT"')}
+          >
+            subject:"SWIFT"
+          </Button>
+
+          <Button type="button" onClick={() => populateSearch('body:"GRB"')}>
+            body:"GRB"
+          </Button>
+          <Button
+            type="button"
+            onClick={() => populateSearch('body:"Tomas Ahumada Mena"')}
+          >
+            body:"Tomas Ahumada Mena"
           </Button>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


app/routes/circulars._archive._index/CircularsLuceneMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+        body). Further documentation can be found on the{' '}
+        <Link className="usa-link" to="/docs/circulars/lucene">
+          Lucene Search Syntax Page
+        </Link>
+        {'. '}
+        <h4>Lucene Examples:</h4>
+        <div>
+          <Button
+            type="button"
+            className="usa-button--outline"
```
lpsinger, line 30:
```suggestion
            outline
```

and same for the other buttons.

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
```diff
@@ -227,6 +228,7 @@ export default function () {
             name="query"
             type="search"
             defaultValue={inputQuery}
+            value={inputQuery}
```
lpsinger, line 231: I don't remember where I read this, but I think it's considered a common mistake to have an input that, on change, updates a state variable, _and_ whose value comes from a state variable. @dakota002, @Courey, have you heard that? Or am I misremembering?

dakota002: I'm not sure on that, I think it might be dependent on where in the general hierarchy of components those state values are used. Like if they need to affect a value in a child/parent component. And if they are basic inputs vs more complex components. But I think generally speaking your idea sounds correct

I think there was something going on here where, when the page reloads due to a form submission, the value was not updating correctly, like the `defaultValue` would still show the previous value, even though the query string would update. @tylerbarna is this correct or am I thinking of another thing?

tylerbarna: @dakota002 without setting `value` the search bar won't update to reflect the current query string inside the search bar until the page is reloaded, so there's not visual feedback that one of the quick action buttons have been pressed until you reload the page.

tylerbarna: @lpsinger apologies for missing this last week, I'm going to close this PR and open a new one with just the in-page documentation

tylerbarna:
The documentation portion of this PR has been moved to #2713. The quick action buttons will be addressed in a future PR.

PR #2301, opened 2024-05-24
Add Kafka ACL admin to website 

dakota002:
# Description

New route to manage a DynamoDB representation of the ACLs defined on the Kafka brokers

PR #2590, opened 2024-09-20
Moves the definition of the send function to resolve issues with kafka related imports in the remix build

### LINKED ISSUE #2588: Deployments failing due to missing zstd libraries in Remix server bundles
# Current behavior

Deployments are failing: https://github.com/nasa-gcn/gcn.nasa.gov/actions/runs/10925788532/job/30328269477

In  the CloudFormation console, I can see that the resource `AnyCatchallHTTPLambdaAliaslive` is failing to update with the following error message:

`Resource handler returned message: "Provisioned Concurrency configuration failed to be applied. Reason: FAILED" (RequestToken: ..., HandlerErrorCode: NotStabilized)`

This usually means that when AWS Lambda tries to start our code, it is crashing. Indeed, in the CloudWatch log, I see this exception:

```json
{
    "errorType": "Runtime.ImportModuleError",
    "errorMessage": "Error: Cannot find module '/_static/app/_assets/zstd.linux-arm64-gnu-4QNIIDDD.node'\nRequire stack:\n- /var/task/index.cjs\n- /var/runtime/index.mjs",
    "stack": [
        "Runtime.ImportModuleError: Error: Cannot find module '/_static/app/_assets/zstd.linux-arm64-gnu-4QNIIDDD.node'",
        "Require stack:",
        "- /var/task/index.cjs",
        "- /var/runtime/index.mjs",
        "    at _loadUserApp (file:///var/runtime/index.mjs:1087:17)",
        "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1119:21)",
        "    at async start (file:///var/runtime/index.mjs:1282:23)",
        "    at async file:///var/runtime/index.mjs:1288:1"
    ]
}
```

This broke due to #2568 which caused the Remix server code to depend up on gcn-kafka-js which in turn requires the zstd library. Native libraries are not bundled by default.

[Our build script for the custom Lambdas, `esbuild.js`](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/esbuild.js), contains a custom esbuild plugin to bundle them. I don't know at the moment if it is possible to customize Remix's build process in a similar manner.

Possible solutions:
- Ensure that it is always the table stream Lambda that is sending Kafka records.
- Extend the Remix build process to bundle native modules.

# Expected behavior

Lambda should start successfully.

# Steps to reproduce

N/A

# Environment

N/A

dakota002:
Temporary solution for #2588

It looks like the issue started with [this commit](https://github.com/nasa-gcn/gcn.nasa.gov/commit/dd4f9d6971c3e35e1d6c46fe69180f4affeacc4b)

This PR moves the definition of the send function so it can still be reusable and should remove any dependency that would have caused the `.node` library to be needed. 

The next step is to figure out the best way to include the zstd library in the build, as future updates will possibly require it for more Kafka functionality

### DIFF
--- a/app/lib/email.server.ts
+++ b/app/lib/email.server.ts
@@ -5,7 +5,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import { services } from '@architect/functions'
+import { services, tables } from '@architect/functions'
 import type {
   BulkEmailEntry,
   SendBulkEmailCommandInput,
@@ -17,15 +17,20 @@ import {
   SendBulkEmailCommand,
   SendEmailCommand,
 } from '@aws-sdk/client-sesv2'
+import { paginateQuery, paginateScan } from '@aws-sdk/lib-dynamodb'
+import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
 import chunk from 'lodash/chunk'
 
-import { hostname } from './env.server'
+import { hostname, origin } from './env.server'
 import { getEnvBannerHeaderAndDescription } from './utils'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+import { formatCircularText } from '~/routes/circulars/circulars.lib'
 import { encodeToURL } from '~/routes/unsubscribe.$jwt/jwt.server'
 
 const client = new SESv2Client({})
 // https://docs.aws.amazon.com/ses/latest/dg/quotas.html
 const maxRecipientsPerMessage = 50
+const fromName = 'GCN Circulars'
 
 interface MessageProps {
   /** The name to show in the From: address. */
@@ -60,7 +65,7 @@ function getFrom(fromName: string) {
   return `${fromName} <no-reply@${hostname}>`
 }
 
-async function send(sendCommandInput: SendEmailCommandInput) {
+async function sendSES(sendCommandInput: SendEmailCommandInput) {
   const command = new SendEmailCommand(sendCommandInput)
   try {
     await client.send(command)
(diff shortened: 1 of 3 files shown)


lpsinger:
This won't fix the issue. The problem is that the zstd shared library is not getting bundled.

dakota002:
> This won't fix the issue. The problem is that the zstd shared library is not getting bundled.

I'm investigating this as an option for the first possible solution you mentioned. 
"Ensure that it is always the table stream Lambda that is sending Kafka records"

The table-stream lambda already is the only thing sending Kafka records. But, since it was exporting a function that gets used in the remix lambda, I'm trying to determine whether or not that meant that its dependencies are getting tangled up in an unintended way. 

The main issue that I am seeing though is that this may be suggesting the problem is also going to be present in https://github.com/nasa-gcn/gcn.nasa.gov/pull/2301

Still investigating though, I think the better solution will be to figure out the inclusion of the lib, but in the meantime this may be a good option to unblock the broken deployments

lpsinger:
> I'm investigating this as an option for the first possible solution you mentioned.

I understand now. Go for it!

lpsinger (commented):
How will this fix the issue?

dakota002:
> How will this fix the issue?

I linked the commit that caused this in the description at the top. Specifically I believe that exporting/importing the `send` function from  `app/table-streams/circulars/index.ts` caused the dependencies (the zstd node lib required by gcn-kafka) to be required by the main build. 

Moving the send function out should resolve the issue, since this will be back, more or less, to a successful build state

dakota002:
I'm pretty sure that's not the case though, where are you referring to? The only place I see our Kafka send function used is in the table-stream

lpsinger:
> I'm pretty sure that's not the case though, where are you referring to? The only place I see our Kafka send function used is in the table-stream

Right, I understand now. OK, this will work!

app/lib/email.server.ts
lpsinger, line 1: The file `email.server.ts` contains functions for sending any kind of email. Nothing in this file now is specific to GCN Circulars. Please put the Circulars specific methods somewhere else.

lpsinger (changes requested):

app/lib/email.server.ts
dakota002: Good point, I moved them to the circulars server in the latest commit

lpsinger, line 1: Are the remaining changes to this file still necessary?

lpsinger (changes requested):

app/lib/email.server.ts
dakota002: nope! Reverted in latest commit

lpsinger (approved):

PR #2610, opened 2024-10-04
update burstcube status

### LINKED ISSUE #2583: Update BurstCube mission page with status
# Description

Update BurstCube mission page given end of mission and status of GCN notices usage.

# Acceptance criteria

Mission page updated.

jracusin:
# Description
Updated BurstCube mission status and explained why page is being maintained as a reference for future missions.

# Related Issue(s)
Resolves #2583 

# Testing
Rendered in local version.

### DIFF
--- a/app/routes/missions.burstcube/route.mdx
+++ b/app/routes/missions.burstcube/route.mdx
@@ -7,10 +7,6 @@ import logo from './logo.png'
 
 import { Alert } from '@trussworks/react-uswds'
 
-<Alert type="warning" heading="Not yet active" headingLevel="h4">
-  BurstCube Notice types are not yet in production.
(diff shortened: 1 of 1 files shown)


app/routes/missions.burstcube/route.mdx
```diff
@@ -7,10 +7,6 @@ import logo from './logo.png'
 
 import { Alert } from '@trussworks/react-uswds'
```
dakota002, line 8: I think you can remove this import since you removed the component below (unless there is another spot where it's used, but I'm pretty sure it's not the case)

dakota002 (changes requested):
Just the one line needs to be removed, otherwise looks good

PR #2617, opened 2024-10-10
Circulars Archive Group View

### LINKED ISSUE #2544: Circulars Grouped by Synonyms index
update the circulars index to have a grouped option,
pagination must work across both index and grouped views,
synonym groups should be listed in a style similar to an inbox, where the contents of that group can be opened and each event lists a sub grouping of circulars related to that event.

Courey:
This work is based off #2538 and will be turned into a non-draft PR once that work has been merged in and I rebase off main.

# Description
This work includes:
- a button to toggle from the index view to the group view
- a drop down list of the circular titles when viewed from the circulars archive
- a link to each circular group in the circulars archive group view
- a route for a synonym group overview
- full circular data in a drop down on the group overview page with a link to each individual circular
- a feature flag hiding the synonyms work

# Related Issue(s)
Resolves #2544

# Testing
This has been thoroughly manually tested locally, but should be tested on dev WITHOUT the feature flag on to ensure that no existing functionality has broken. 

# Images
Screenshot 2024-10-10 at 3 26 08 PM
Screenshot 2024-10-10 at 3 26 19 PM
Screenshot 2024-10-10 at 3 26 34 PM
Screenshot 2024-10-10 at 3 26 44 PM


lpsinger (changes requested):
Please rebase.

Courey:
Added "Open/Close All" button on group overview page as requested by @jracusin 
Screenshot 2024-10-16 at 11 47 38 AM
Screenshot 2024-10-16 at 11 47 32 AM

Courey:
> Please rebase.

That was the plan as I said in the original PR description:

> This work is based off https://github.com/nasa-gcn/gcn.nasa.gov/pull/2538 and will be turned into a non-draft PR once that work has been merged in **_and I rebase off main_**.

I rebased and took the PR out of draft status.

lpsinger (changes requested):
There are more conflicts. Please rebase again.

lpsinger (changes requested):
First, some high-level UX feedback:

- [ ] Please use numbered lists, not bulleted lists, for Circulars in the index view. The list number value should be the Circular number.
- [ ] For now, don't have disclosure arrows on the per-group page. Just display all of the Circulars belonging to the group. We can fine-tune the UI to quickly navigate within a group in a future PR.

Courey:
Removed details element from group overview:
Screenshot 2024-10-22 at 09-36-40 GCN - Circulars

Made ul an ol with circularId values:
Screenshot 2024-10-22 at 9 36 13 AM

app/components/circularDisplay/AstroDataContext.tsx
```diff
@@ -29,6 +29,7 @@ export const AstroDataContext = createContext<AstroDataContextProps>({})
 /**
  * An Astro Flavored Markdown enriched link.
  */
+// eslint-disable-next-line react/display-name
```
lpsinger, line 32: Why was this addition necessary?

app/components/pagination/PaginationSelectionFooter.tsx
```diff
@@ -9,40 +9,51 @@ import { useSubmit } from '@remix-run/react'
...
             <option value="10">10 / page</option>
             <option value="20">20 / page</option>
-            <option value="50">50 / page</option>
-            <option value="100">100 / page</option>
+            {!isGroupView && (
+              <>
+                <option value="50">50 / page</option>
+                <option value="100">100 / page</option>
+              </>
+            )}
```
lpsinger, line 56: Why don't we show options for 50 or 100 results per page for the group view?

app/components/pagination/PaginationSelectionFooter.tsx
```diff
@@ -53,6 +64,7 @@ export default function PaginationSelectionFooter({
             page={page}
             limit={limit}
             totalPages={totalPages}
+            view={view}
```
lpsinger, line 67: Why does the Pagination component need to know about the view?

app/routes/synonyms.$synonymId.tsx
```diff
@@ -292,7 +292,7 @@ export default function () {
               unstyled
               className="padding-105 text-center"
             >
-              Cancel
+              Go back
```
lpsinger, line 295: Please keep the text `Cancel`. This was from feedback on an earlier PR that we have merged.

lpsinger (changes requested):

app/components/pagination/PaginationSelectionFooter.tsx
Courey: each group could have many circulars. If each group had on average 10 circulars, if I included 100 groups at a time, that would be querying 1000 circulars to load the page. It seemed reasonable to limit it. Do you think large circular queries would be problematic?

app/routes/synonyms.$synonymId.tsx
Courey: this one is mystifying me. I'm assuming a change snuck in during the rebase, but when I look at the branch locally, this is what I see:
Screenshot 2024-10-22 at 1 42 24 PM
I can change the B to be capitalized in "Go back" and it reflects the change but the diff says it was "Go back" previously.
Screenshot 2024-10-22 at 1 43 56 PM

Courey: if you click  "view file" in the three dot menu you can scroll down and it also says go back:
[app/routes/synonyms.$synonymId.tsx](https://github.com/nasa-gcn/gcn.nasa.gov/blob/2ac116c459bc8a799a5cbd76cfcd553b621799d1/app/routes/synonyms.%24synonymId.tsx)

app/components/pagination/PaginationSelectionFooter.tsx
Courey: because in getPageLink if the view is not set as a search param, then when it creates the link, it doesn't include view. If it doesn't include view, then the view defaults to index. If it's always index, then the pagination links will never work for groups. To show you what I mean, here is what happens when I remove the view from getPageLinks. When I hover over the 2 button, this is the link. You will see that it doesn't include the view so when that page is navigated to, it's the index view which is the default.
pagination_view

app/components/circularDisplay/AstroDataContext.tsx
Courey: when I moved the component into the components directory, it is judged as a component definition and causes this warning:
Screenshot 2024-10-22 at 1 58 18 PM

Courey: I was trying to make the move have no changes to the code if possible

app/routes/synonyms.$synonymId.tsx
lpsinger: OK, but I think that this should be `Cancel`, don't you?

app/components/pagination/PaginationSelectionFooter.tsx
lpsinger: Not necessarily. I would prefer not limiting it for now.

lpsinger:
> Removed details element from group overview: Screenshot 2024-10-22 at 09-36-40 GCN - Circulars
> 
> Made ul an ol with circularId values: Screenshot 2024-10-22 at 9 36 13 AM

Instead of the boxes, could we just put a horizontal rule between Circulars for now?

app/routes/group.$synonymId.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { ToolbarButtonGroup } from '~/components/ToolbarButtonGroup'
+import { PlainTextBody } from '~/components/circularDisplay/Body'
+import { FrontMatter } from '~/components/circularDisplay/FrontMatter'
+import { feature } from '~/lib/env.server'
+
+export async function loader({
+  params: { synonymId },
+  request: { url },
+}: LoaderFunctionArgs) {
+  if (!feature('SYNONYMS')) throw new Response(null, { status: 403 })
```
lpsinger, line 26:
```suggestion
  if (!feature('SYNONYMS')) throw new Response(null, { status: 404 })
```

app/routes/group.$synonymId.tsx
```diff
@@ -0,0 +1,81 @@
...
+  if (!synonymId) throw new Response('Id is required', { status: 400 })
```
lpsinger, line 27: No need to return a sane HTTP response here, because this route is only reachable if the synonymId parameter is a non-empty string.

```suggestion
  invariant(synonymId)
```

app/routes/group.$synonymId.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const { searchParams } = new URL(url)
+  const view = searchParams.get('view') || 'group'
+  const limit = searchParams.get('limit') || 20
+  const page = searchParams.get('page') || 1
+  const synonym = (await getSynonymById(synonymId)) as SynonymGroup
```
lpsinger, line 33: Avoid type assertion here. Why was it failing to infer the type?

app/routes/group.$synonymId.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const members = await getAllSynonymMembers(synonym.eventIds)
+
+  return { members, eventIds: synonym.eventIds, view, limit, page }
+}
+
+export default function Group() {
+  const { members, eventIds, view, limit, page } =
+    useLoaderData<typeof loader>()
+  const searchString = `?view=${view}&limit=${limit}&page=${page}`
```
lpsinger, line 42: Can you use URLSearchParams or similar here to construct this query string in a more structured way, rather than doing manual string formatting?

app/routes/group/route.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { feature } from '~/lib/env.server'
+import type { BreadcrumbHandle } from '~/root/Title'
+
+export const handle: BreadcrumbHandle = {
+  breadcrumb: 'Circulars',
+}
+
+export function loader() {
+  if (!feature('SYNONYMS')) throw new Response(null, { status: 403 })
```
lpsinger, line 19:
```suggestion
  if (!feature('SYNONYMS')) throw new Response(null, { status: 404 })
```

app/routes/synonyms/synonyms.server.ts
```diff
@@ -281,3 +273,88 @@ export async function autoCompleteEventIds({
...
+      ':eventId': eventId,
+    },
+  })
+  return Items as Circular[]
+}
+
+export async function getAllSynonymMembers(eventIds: string[]) {
+  const promises = eventIds.map((eventId) => {
+    return getSynonymMembers(eventId)
+  })
```
lpsinger, line 357:
```suggestion
  const promises = eventIds.map(getSynonymMembers)
```

lpsinger (changes requested):

app/components/pagination/PaginationSelectionFooter.tsx
Courey: I think that's a good idea. It did feel odd to have to add view specific code to it.

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <summary>
+              <Link to={`/group/${group.synonymId}${searchString}`}>
+                {group.eventIds.join(', ')}
+              </Link>
+            </summary>
+            <ol className="margin-left-3">
+              {members.map(({ circularId, subject }) => {
+                return (
+                  <li key={circularId} value={circularId}>
+                    <Link to={`/circular/${circularId}`}>{subject}</Link>
```
lpsinger, line 43:
```suggestion
                    <Link to={`/circulars/${circularId}`}>{subject}</Link>
```

app/routes/group.$synonymId.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <Icon.ArrowBack
+              role="presentation"
+              className="position-absolute top-0 left-0"
+            />
+          </div>
+          &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Back
+        </Link>
+      </ToolbarButtonGroup>
+
+      <h1>{`Group ${eventIds.join(', ')}`}</h1>
```
lpsinger, line 65:
```suggestion
      <h1>{eventIds.join(', ')}</h1>
```

app/routes/group.$synonymId.tsx
lpsinger, line 1: Add breadcrumb to create a meaningful title on this page.

lpsinger (changes requested):
- [ ] Please place the new event pages under /circulars/group/<event-name> rather than /group/<event-name>.
- [ ] The synonym IDs make for URLs that are not very presentable. Instead, can we use a slugified form of the event name (any of the event names that belong to the group) as the path component?

Courey:
> The synonym IDs make for URLs that are not very presentable. Instead, can we use a slugified form of the event name (any of the event names that belong to the group) as the path component?

I have a couple of concerns about this if I'm understanding correctly. 
1. It doesn't seem very RESTful. the id of the group would be the synonymId, not the eventId (because sometimes it is comprised of more than one eventId). it is the synonym we are representing here, not the event.
2. Let's say we had GRB 111224A, GRB 111224B, and GRB 111224C. If we slugified that into something like `GRB_111224A-GRB_111224B-GRB_111224C` if a user bookmarked that uri and later a moderator decided that GRB 111224C was actually an unrelated event and removed that from the group, then the users bookmarked uri wouldn't work anymore. It is fragile. 
3. We could handle for that so it just picks one eventId and pulls up the record based on that, if the group that eventId was in changed, it would pull up a totally different group than originally. so `circulars/group/GRB_111224C` would now bring up just that record instead of the original group it belonged to before a moderator changing it. So it wouldn't go to the same group as when the bookmark/link was created.
5. If we disregarded any concerns about uri fragility or possibly pulling up the wrong record than intended, it also changes how things work on the back end:
    - we would have to parse the slug which isn't a huge deal, but it's not as terse as just looking up by synonymId.
    - If we slugified all the eventIds in the group, we would be able to get the records based on those eventIds. doing it that way may include records from a different group as it's not specifying the synonymId. 
    - if we try to handle that by checking to see if the records have the same synonymId, and we have two records and they both have different synonymIds, which synonymId wins out? 
    - if we accepted just one eventId of the group, we'd have to make a db call to get the synonymId for that eventId, then using those results, we'd have to make another call to get all the other eventIds with that syonymId. So it increases queries.

Can you explain what about uuids feels not presentable to you?

lpsinger:
> > The synonym IDs make for URLs that are not very presentable. Instead, can we use a slugified form of the event name (any of the event names that belong to the group) as the path component?
> 
> I have a couple of concerns about this if I'm understanding correctly.
> 
> 1. It doesn't seem very RESTful. the id of the group would be the synonymId, not the eventId (because sometimes it is comprised of more than one eventId). it is the synonym we are representing here, not the event.
> 2. Let's say we had GRB 111224A, GRB 111224B, and GRB 111224C. If we slugified that into something like `GRB_111224A-GRB_111224B-GRB_111224C` if a user bookmarked that uri and later a moderator decided that GRB 111224C was actually an unrelated event and removed that from the group, then the users bookmarked uri wouldn't work anymore. It is fragile.
> 3. We could handle for that so it just picks one eventId and pulls up the record based on that, if the group that eventId was in changed, it would pull up a totally different group than originally. so `circulars/group/GRB_111224C` would now bring up just that record instead of the original group it belonged to before a moderator changing it. So it wouldn't go to the same group as when the bookmark/link was created.
> 4. If we disregarded any concerns about uri fragility or possibly pulling up the wrong record than intended, it also changes how things work on the back end:
>    
>    * we would have to parse the slug which isn't a huge deal, but it's not as terse as just looking up by synonymId.
>    * If we slugified all the eventIds in the group, we would be able to get the records based on those eventIds. doing it that way may include records from a different group as it's not specifying the synonymId.
>    * if we try to handle that by checking to see if the records have the same synonymId, and we have two records and they both have different synonymIds, which synonymId wins out?
>    * if we accepted just one eventId of the group, we'd have to make a db call to get the synonymId for that eventId, then using those results, we'd have to make another call to get all the other eventIds with that syonymId. So it increases queries.

I proposed using "any of the event names that belong to the group" as the path component. So /circulars/groups/GW170817, /circulars/groups/GRB170817A, /circulars/groups/GW170817+GRB170817A would all represent the same resource. (One would need to be reported as the canonical URL for SEO purposes; probably the version with the maximal set of terms.)

Another option (but not my preference) would be to do what Stack Overflow does: have an opaque UUID followed by a slug that is ignored. So for example, all of the following refer to the same resource:

- https://stackoverflow.com/questions/79182387/sequelize-typescript-error-property-optypes-and-does-not-exist-on-type-wh
- https://stackoverflow.com/questions/79182387
- https://stackoverflow.com/questions/79182387/foobar

although the first is canonical.

> Can you explain what about uuids feels not presentable to you?

It's a usability issue: https://en.wikipedia.org/wiki/Clean_URL

Courey:
> I proposed using "any of the event names that belong to the group" as the path component. So /circulars/groups/GW170817, /circulars/groups/GRB170817A, /circulars/groups/GW170817+GRB170817A would all represent the same resource. (One would need to be reported as the canonical URL for SEO purposes; probably the version with the maximal set of terms.)

That still does not solve for or answer the issues I brought up in the above comment. If we take your example of  /circulars/groups/GW170817, /circulars/groups/GRB170817A, /circulars/groups/GW170817+GRB170817A. Jane Doe is interested in this grouping that she discovered when searching for GRB170817A. She bookmarks /circulars/groups/GRB170817A so she can go back and look at this group. A moderator looks at the grouping and decides that while the events were very close together, they are actually different events. So the moderator removes GRB170817A from the group with GRB170817. So now when Jane visits the website looking for the information from both GRB170817A and GRB170817, the link she bookmarked would take her to the group that only has GRB170817A in it. 

This request would be like removing the circularId from the circulars path and replacing it with the circular subject instead. But you shouldn't do that because the subject could be edited and if so, it would break links because it's fragile. 

clean_url

circular_path

synonym_path

The only difference in these patterns is UUID vs integer id. But the integer ID isn't giving any additional information in a human readable format. I have no insight into what circular 12345 is about. 

To make it restful, it should follow the pattern /{resource}/{id}

https://restfulapi.net/resource-naming/
Screenshot 2024-11-14 at 10 07 33 AM

lpsinger:
I don't disagree that RESTfulness is a desirable property for URL routing, but an entity need not have a _unique_ URL, does it? By analogy, files on a filesystem don't have unique paths.

It's true that we are using Circular IDs as URL path components, but Circular IDs are part of the publicly-visible bibliographic record of Circulars. UUIDs are an internal detail.

Courey:
I disagree that what you are proposing is "clean". 
from your wiki link on clean urls:

> Other reasons for using clean URLs include [search engine optimization](https://en.wikipedia.org/wiki/Search_engine_optimization) (SEO),[[1]](https://en.wikipedia.org/wiki/Clean_URL#cite_note-cws-1) conforming to the [representational state transfer](https://en.wikipedia.org/wiki/Representational_state_transfer) (REST) style of software architecture, and ensuring that individual [web resources](https://en.wikipedia.org/wiki/Web_resource) remain consistently at the same URL. This makes the [World Wide Web](https://en.wikipedia.org/wiki/World_Wide_Web) a more stable and useful system, and allows more durable and reliable [bookmarking](https://en.wikipedia.org/wiki/Bookmark_(World_Wide_Web)) of web resources

your request violates:

> conforming to the [representational state transfer](https://en.wikipedia.org/wiki/Representational_state_transfer) (REST) style of software architecture

because calling a resource by one of it's members is not restful. not restful != clean

and 

> ensuring that individual [web resources](https://en.wikipedia.org/wiki/Web_resource) remain consistently at the same URL. This makes the [World Wide Web](https://en.wikipedia.org/wiki/World_Wide_Web) a more stable and useful system, and allows more durable and reliable [bookmarking](https://en.wikipedia.org/wiki/Bookmark_(World_Wide_Web)) of web resources

because of the situations outlined above when the members of the group change. If I have a group of GRB A, GRB B, GRB C, and GRB D. If I bookmark `circulars/groups/GRB_D` it it had been removed, it would take me to the new group for GRB D without any reference to why GRB A, GRB B, and GRB C that were there before are now missing. This is not a durable uri because it is based on changeable values. One would instead expect to visit the group bookmarked originally and see that GRB D was no longer there, then one could search for GRB D to see what new group it was part of. Just having it jump to the new group with no indication of why the new group is not the old group is not an anticipated behavior and it does not make for reliable bookmarking. Since the slugs would be based on changeable values, bookmarks would always be fragile which is also not clean.

Additionally, is it really worth the complexity that it creates? 
things that would have to be accounted for:
1. slugging and de-slugging singular eventIds
2. handling slugging and de-slugging  more than 1 eventId
3. looking for a synonym group by getting the synonym off the de-slugified eventId and then making a second query for the members of the group by the results of that query. 
4. keeping track of SEO for changing group members

lpsinger:
In a URL like /circulars/group/GW170817, the "group" component is a the resource and "GW170817" is the ID. I view the event names as permanent, stable IDs. The UUIDs are not.

Although this approach appears RESTful to me, I don't want to get bogged down in that argument, because this path is not an API endpoint and it is not all that important whether it is RESTful or not. It's more important that they are human-readable and predictable.

Here are some examples of style guidelines and advice related to human-readable URLs:
- https://developers.google.com/search/docs/crawling-indexing/url-structure
- https://www2.gov.bc.ca/gov/content/governments/services-for-government/service-experience-digital-delivery/web-content-development-guides/cms-lite-manual/content-menu/topic-template/hrus
- https://webmasters.stackexchange.com/questions/134183/is-uuid-based-url-bad
- https://webmasters.stackexchange.com/questions/133701/uuid-or-slug-first-in-the-url-for-better-ranking-does-it-change-with-canonicals
- https://www.hanselman.com/blog/urls-are-ui
- https://medium.com/design-bridges/designing-better-links-for-websites-and-emails-a-guideline-5b8638ce675a

> Additionally, is it really worth the complexity that it creates?
> things that would have to be accounted for:
> 
> 1. slugging and de-slugging singular eventIds
> 2. handling slugging and de-slugging more than 1 eventId
> 3. looking for a synonym group by getting the synonym off the de-slugified eventId and then making a second query for the members of the group by the results of that query.
> 4. keeping track of SEO for changing group members

That looks like a reasonable implementation plan. Is there a good open-source slugging library we could use?

Courey:
So if the group consists of events GRB 123456A, GRB 123456B, and GRB 123456C, is GRB 123456A the group? or is GRB 123456A a member of the group?

if we have a data structure that is like so:
```
{
  synonymId: 1,
  eventIds: [GRB 123456A, GRB 123456B, GRB 123456C]
}
```
Which is how we are representing the concept of a synonym group. I acknowledge that how it is stored in dynamo does not make it apparent, but that is because we had to do some unusual things to use a document store relationally. But the general concept of a group is the above structure. 

According to that record above, is the resource id of the group 1 or is it GRB 123456A?

If I remove GRB 123456C from the synonym group, thus making it a group unto itself I would have these two records:

```
{
  synonymId: 1,
  eventIds: [GRB 123456A, GRB 123456B]
}

{
  synonymId: 2,
  eventIds: [GRB 123456C]
}
```

Is the group the same for GRB 123456C as it was before? it used to be in group 1, now it is in group 2. Are group 1 & 2 the same resource?

lpsinger:
> So if the group consists of events GRB 123456A, GRB 123456B, and GRB 123456C, is GRB 123456A the group? or is GRB 123456A a member of the group?

I would consider the group to be a collection of _circulars_. I don't care that there is a DynamoDB table of synonyms. Synonyms are not a user-visible aspect of our application (except to moderators).

In your example, I suggest that all of the following URLs resolve to the same resource:

- /circulars/group/grb_123456a
- /circulars/group/grb_123456b
- /circulars/group/grb_123456c
- /circulars/group/grb_123456a-grb_123456b
- /circulars/group/grb_123456a-grb_123456c
- /circulars/group/grb_123456b-grb_123456a
- /circulars/group/grb_123456b-grb_123456c
- /circulars/group/grb_123456c-grb_123456a
- /circulars/group/grb_123456c-grb_123456b
- /circulars/group/grb_123456a-grb_123456b-grb_123456c
- /circulars/group/grb_123456a-grb_123456c-grb_123456b
- /circulars/group/grb_123456b-grb_123456a-grb_123456c
- /circulars/group/grb_123456b-grb_123456c-grb_123456a
- /circulars/group/grb_123456c-grb_123456a-grb_123456b
- /circulars/group/grb_123456c-grb_123456b-grb_123456a

The _canonical_ URL should be /circulars/group/grb_123456a-grb_123456b-grb_123456c, which consists of all of the synonyms in lexical order.

To resolve the group for a given URL, take the entity ID and split on the separator (in the example above, a dash). Take just the first name in the group, and look up its synonyms. Return all Circulars for all synonyms belonging to that group.

Courey:
Okay, so according to your example, I have bookmarked group:
`/circulars/group/grb_123456c-grb_123456a`
except between the time I bookmarked it and the time that I went back to look at it, a moderator removed grb_123456c from the group. Which do I go to? The group for grb_123456c or the group for grb_123456a? What if they bookmarked that link but were going back to look at grb_123456a, so if we just went to the first one in the list, it would send them to the wrong group.

Courey:
in the above case `/circulars/group/grb_123456c-grb_123456a` would take you to a group that didn't actually include grb_123456a. Is that expected?

Courey:
so you are saying that if they looked up `/circulars/group/grb_123456c-grb_123456a` and c had been removed from the group, then it would take them to group c while the bookmarked uri still says that it's taking them to something with c and a in it. why would a bookmark that tells me that it includes both c and a take me somewhere that didn't include a?

lpsinger:
> so you are saying that if they looked up `/circulars/group/grb_123456c-grb_123456a` and c had been removed from the group, then it would take them to group c while the bookmarked uri still says that it's taking them to something with c and a in it. why would a bookmark that tells me that it includes both c and a take me somewhere that didn't include a?

I'll refer you again to this algorithm:

> To resolve the group for a given URL, take the entity ID and split on the separator (in the example above, a dash). Take just the first name in the group, and look up its synonyms. Return all Circulars for all synonyms belonging to that group.

The alternatives that I see are to raise a 404 error if the URL does not contain all of the event names, or to only return the circulars belonging to the event names in the URL. Both of those alternatives would be more surprising to the user.

Courey:
> The alternatives that I see are to raise a 404 error if the URL does not contain all of the event names, or to only return the circulars belonging to the event names in the URL. Both of those alternatives would be more surprising to the user.

But what wouldn't be surprising to anyone in any case is if the user had a link to a group id instead of to the changing members of the group. 

if you had the uri as `/group/123`, Since the resource is the group itself, not it's members, the resource id would be the group id (in our case the synonymId). 
synonyms_all_together_1

that way, when the editable field gets changed, the uri is not comprised of editable members, so it doesn't get confusing.
synonyms_structure_2

You would always go to the group and who the members of that group are can change freely without impacting anything while also following REST. That's why it's the convention. if the user bookmarks `group/123` then no matter what was in that group, they know they would always be getting the intended group.

lpsinger:
The URL structure does not need to reflect the database structure. The UUIDs are a bit of dirty laundry that I don't want users to see, ever.

Courey:
There are several characters we use in eventIds that are not alphanumeric. they are:
'/', '.', ' ', '-', '+'

Some of these are acceptable in URIs, others are reserved and can not be used.
'.' and '-' are acceptable, the rest are not.

Here are examples of the different characters that are used in eventIds:
HAWC-230527A
SGR 1935+2154
GRB 231115A
LIGO/Virgo S190425z
GRB 160225.81

1. Since we are using the slug to look up the synonym, we need to ensure that any slugification can be de-slugified exactly. Typical slugging behavior is not reversible, which creates a challenge in our case. You can not exactly de-slug them because they typically convert unsafe characters to dashes (or a specified character) and we would not be able to tell what the original characters were. Since we are using the eventId as an id for lookup, this is not acceptable.
the slugified versions of the examples above would be as follows:
HAWC-230527A
SGR-1935-2154
GRB-231115A
LIGO-Virgo-S190425z
GRB-160225-81

If we tried to use the slug to query the database, we would not be able to return them exactly to their original format, we would just know that reserved or unacceptable characters had been turned into dashes. This would have a similar impact to SEO as the other options where the keywords are not exact. (GRB-160225-81 would be GRB, 160225, and 81 instead of 160225.81)
If we used a relational databse, we could use a series of wildcard where clauses to search for the appropriate record, but in dynamodb this is not supported. We could use a table scan (which is expensive and not performant) and/or multiple queries with a begins with and contains. This is not exact and the results could be wrong in addition to it being not very performant and expensive.
Pure sluggification is not an option in our case.

2. The best option for reversability would be to percent encode. 
The cons of this approach are that it is ugly and not super human readable since it would have percent encoded values sprinkled in. 
Crawlers can understand them, but it sort of defeats the purpose of using something human readable. 
SEO impact would be that the keywords may be obscured which may impact searchability.

here are the examples above url encoded:
HAWC-230527A
SGR%201935%2B2154
GRB%20231115A
LIGO%2FVirgo%20S190425z
GRB%20160225.81

3. An approach that includes exact reversibility would be to create a custom encoder/decoder. 
This would be a function that replaces the reserved characters with a string representing each character. 
This would get us a moderately human readable slug that can be reversed. 
It would have meaningful separators, so while it is not as clean as a normal slug, it is still fairly SEO friendly. 
Keywords would be preserved, so SEO could still extract useful keywords.

Encoded with our custom encoder they would appear like so:
HAWC--dash--230527A
SGR--space--1935--plus--2154
GRB--space--231115A
LIGO--slash--Virgo--space--S190425z
GRB--space--160225--dot--81

it's unusual, but functional. It can be revesed exactly so the value can be used in a query and it would preserve keywords.
It would obscure parts of the id for SEO though. For example GRB--space--160225--dot--81 would have keywords GRB, 160225, and 81 instead of GRB and 160225.81

None of the options are ideal and all will have some impact on SEO.

Option 1 is not an option.
Option 2 is not super human readable and has minimal SEO keyword impact with no impact to Crawlers
Option 3 is weird but reversible and preserves most keywords though it does have some minimal SEO impact similar to the pure slug option.

Courey:
An additional option would be to add a field to the synonyms table called `slug` and when an eventId is created or updated, we could replace all reserved characters with dashes. For example, the record would be:
```
{
  synonymId: 'uuid-uuid-uuid-uuid',
  eventId: 'LIGO/Virgo S190425z',
  slug: 'LIGO-Virgo-S190425z'
}
```
Then we would have to add an additional index to the slug field. That would change the lookup slightly from a pkey lookup to a GSI lookup.

We would need an additional backfill for the existing synonyms to create the slug.

lpsinger:
@Courey, if we want the slugs to be case insensitive (and I think we do), then we'll need to add the slugs to the database anyway because DynamoDB keys are case sensitive.

Given that, I think we can abandon the requirement that the slugs are reversible, and just replace anything that _would_ require percent encoding with a dash.

There are a number of slugging libraries on NPM. I believe that we already have a dependency on https://www.npmjs.com/package/github-slugger.

Courey:
Additional work for this feature:
- [ ] add slug field to dynamodb record (all lowercase with dashes)
- [ ] add index to the slug field
- [ ] add code to save slug in any place that a synonym could be created or updated (using github slugger package)
- [ ] create function to get synonym by slug
- [ ] update path to be `/circulars/events/<slug>`
- [ ] create additional ticket to backfill any existing synonyms' slugs
- [ ] update existing ticket and code for synonym backfill to include slug

lpsinger:
This is great, @Courey. I am swamped until the end of the day but I'll try to post a review by tomorrow. Meanwhile, @dakota002, would you please give it a once over?

Courey:
The tests have been fixed. 

The only other weird thing I see is that in the initial PR for this feature, I squashed (the first commit in this PR). But it looks like it squashed from a previous commit into this one. It was based on the synonym moderation ui branch since they shared work. When I squashed for the initial PR, it included these commits:
```
adds synonyms moderation front end
adds validation and error handling
fixes tests
adds modal warning prior to delete
changes removal button to words instead of only icons
code review change requests
removing unused className
formatting
autofill moderator synonym eventId selector
adding create sad path test
removing feature flag check
adding a 3 second debounce
Adds grouped view to circulars archive index
adds missing route and flag checks
```
These commits are already in main:
```
 adds synonyms moderation front end
adds validation and error handling
fixes tests
adds modal warning prior to delete
changes removal button to words instead of only icons
code review change requests
removing unused className
formatting
autofill moderator synonym eventId selector
adding create sad path test
removing feature flag check
adding a 3 second debounce
```

Which means the only things that _should_ have been included in that particular commit were these two:
```
Adds grouped view to circulars archive index
adds missing route and flag checks
```

I am unsure why they are showing up in the diff though because they are already in main. Still looking into this and deciding if there would be any detrimental impact.

__tests__/synonyms.server.ts
```diff
@@ -10,6 +17,12 @@ import {
 } from '~/routes/synonyms/synonyms.server'
 
 jest.mock('@architect/functions')
+jest.mock('github-slugger', () => ({
+  slug: (eventId: string) => {
+    return eventId.toLowerCase()
+  },
+}))
+
```
lpsinger, line 25: Would you please add a slug to explain why you are mocking github-slugger?

app/routes/synonyms/synonyms.lib.ts
```diff
@@ -5,15 +5,23 @@
...
 }
 
 /* Layout of materialized view in OpenSearch */
 export interface SynonymGroup {
   synonymId: string
   eventIds: string[]
+  slugs: string[]
+}
+
+export interface SynonymGroupWithMembers {
```
lpsinger, line 24: Might be a little cleaner to extend SynonymGroup rather than have it as a member.

app/routes/synonyms/synonyms.server.ts
```diff
@@ -72,7 +90,7 @@ export async function searchSynonymsByEventId({
     },
   } = await client.search({
     index: 'synonym-groups',
-    from: page && limit && (page - 1) * limit,
+    from: page * limit,
```
lpsinger, line 93: Would you please explain this change to me?

app/routes/synonyms/synonyms.server.ts
```diff
@@ -145,7 +151,7 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
     RequestItems: {
       [TableName]: synonymousEventIds.map((eventId) => ({
         PutRequest: {
-          Item: { synonymId: uuid, eventId },
+          Item: { synonymId: uuid, eventId, slug: slug(eventId) },
```
lpsinger, line 154: Hoist `slug(eventId)` to function scope to avoid calling it multiple times.

app/routes/circulars.events.$slug.tsx
```diff
@@ -0,0 +1,94 @@
...
+  params: { slug },
+  request: { url },
+}: LoaderFunctionArgs) {
+  if (!feature('SYNONYMS')) throw new Response(null, { status: 404 })
+  invariant(slug)
+
+  const { searchParams } = new URL(url)
+  const view = searchParams.get('view') || 'group'
+  const limit = searchParams.get('limit') || '20'
+  const page = searchParams.get('page') || '1'
```
lpsinger, line 38: There's no need to get these in the loader; they are available on the client side.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -273,18 +332,30 @@ export default function () {
...
+            />
+          )}
+          {isGroupView && synonymFlagIsOn && (
+            <SynonymGroupIndex
+              allItems={items as SynonymGroupWithMembers[]}
+              searchString={searchString}
+              totalItems={totalItems}
+              query={query}
+            />
+          )}
```
lpsinger, line 350: Slightly simpler:

```tsx
{isGroupView && synonymFlagIsOn ? (<Component1/>) : (<Component2/>)}
```

lpsinger (changes requested):

app/routes/synonyms/synonyms.server.ts
Courey: If you look at the circulars._archive._index/route file, you will see that in the search function it has `page - 1` passed into the search function. So that everything was the same whether it was synonym groups or not, I wanted to keep that as `page - 1` being passed in. If I didn't adjust it, it would be `page -1 -1` when it hit that line. The way that impacted things was that it would show the first page of results twice and you would not be able to see the last page of results.
If you look at this changed file https://github.com/nasa-gcn/gcn.nasa.gov/pull/2617/files#diff-e08b0c9f75242b7710fc6321f15b142e5694a47c82317853845577883ea2f6a9 you will see that I also had to adjust it there to bring it in line with this approach. 
it seemed better to keep it the same as circulars than to do something odd like page + 1 in the loader while leaving the page - 1 in the server function. This way they are all three the same.

app/routes/synonyms/synonyms.server.ts
Courey: I'm not sure that would work. It occurs once in the moderatorCreateSynonyms() map function. If it was hoisted to the function level, all the synonyms would have one singular slug (based on which?) eventId and that's not consistent with how it works. You would want each individual eventId to be slugged since they are different.

It occurs once in tryInitSynonym(), so no real benefit to hoisting it there. 

it occurs in two map functions in putSynonyms(). Again, each synonym being updated or created would need a unique slug based on its eventId. If hoisted in this function, all updated and created synonyms would have the same slug regardless of different eventIds. 

Can you clarify where and why it should be hoisted to function level? Are you wanting all synonyms in a group to have the same slug regardless of eventId?

lpsinger: I see. I simply misread the code. Thanks!

__tests__/synonyms.server.ts
Courey: I added a comment above each mock. Let me know if you wanted something different.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -164,39 +182,58 @@ export async function action({ request }: ActionFunctionArgs) {
...
-
-  const limit = searchParams.get('limit') || '100'
-  const query = searchParams.get('query') || ''
-  const startDate = searchParams.get('startDate') || undefined
-  const endDate = searchParams.get('endDate') || undefined
-  const sort = searchParams.get('sort') || 'circularID'
+  const { query, startDate, endDate, sort, view } = handleSearchParams(
+    searchParams,
+    synonymFlagIsOn
+  )
```
lpsinger, line 228: Please undo this refactoring.

Refactoring to make code more readable is a good thing, but this refactoring is not related to the objectives of this PR. Please don't mix up refactoring with feature or bug fix PRs. It increases the review burden because I have to check that the refactoring itself is correct.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -68,7 +78,15 @@ export async function loader({ request: { url } }: LoaderFunctionArgs) {
     sort,
   })
   const requestedChangeCount = (await getChangeRequests()).length
-  return { page, ...results, requestedChangeCount }
+
+  return {
+    page,
+    ...results,
+    requestedChangeCount,
+    synonymFlagIsOn,
```
lpsinger, line 86: No need to pass this through the loader. This value is already available in components through the `useFeature` hook.

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
Courey: I will undo it, but in the software industry there is something called the campfire rule. This means that if your feature touches existing code, refactoring for readability, performance, etc is encouraged. Not making improvements when working in an area is unusual for most engineers, so that might be something we want to put in the contributing guide or wherever we put our team norms.

lpsinger: These are good improvements! But please do the refactoring in a separate PR. I would like to review the refactoring itself independently from the functional changes.

When you are working on a new feature or a bug fix and you notice a helpful refactoring, please do the refactoring in a separate commit and open a separate PR for it.

Courey:
Is the review complete? I would like to make all requested changes at once because I do a thorough testing locally to ensure none of the refactoring breaks anything. When we do the review in review after review, it eats up a lot of time with the testing because I have to retest all of it each time. I don't want to skip testing because since we merge each others branches, I want to make sure it is completely tested each time I push in case it gets merged in.

lpsinger:
> Is the review complete?

I'm done. I have some look-and-feel requests about the view toggle button, but I had planned to do that in a separate issue after we merge this.

PR #2619, opened 2024-10-10
Add Synonym link to NavBar for moderators

### LINKED ISSUE #2618: Add Synonyms Moderation Button to Circular index page
Add Synonyms Moderation route link/button to Circulars index page. 
It should:
- be moderator gated

Courey:
# Description
Adding moderator gated link to Synonyms moderation route.

With the link not visible as a non-logged in user:
Screenshot 2024-10-10 at 5 06 57 PM

With the link visible:
Screenshot 2024-10-10 at 5 06 44 PM

With the link not visible as a logged in user:
Screenshot 2024-10-10 at 5 10 19 PM

Mobile view:
Screenshot 2024-10-10 at 5 13 54 PM
Screenshot 2024-10-10 at 5 13 47 PM

and because I know you'll ask about it, the reason I added `style={{ width: 'max-content' }}` to the Title is because without it, the additional characters in the PrimaryNav cause the "General Coordinates Network" to all stack on top of each other which makes the header very noticeably taller than normal. Here is an image without that style added:
Screenshot 2024-10-10 at 4 06 11 PM

I can add "Syno" without anything changing at all, but if you add the rest of the word things start to change slightly to fit all the text in the nav bar.

# Related Issue(s)
Resolves #2618 

# Testing
Manual testing both as a logged in moderator and a logged out user. This should be tested on dev with a logged out user, a logged in non-moderator user, and a logged in moderator.

### DIFF
--- a/app/root/header/Header.tsx
+++ b/app/root/header/Header.tsx
@@ -111,7 +111,7 @@ export function Header() {
       <USWDSHeader basic className={`usa-header--dark ${styles.header}`}>
         <div className="usa-nav-container">
           <div className="usa-navbar">
-            <Title>
+            <Title style={{ width: 'max-content' }}>
               <Link to="/">
                 <Meatball className="width-auto" />
                 <span>General Coordinates Network</span>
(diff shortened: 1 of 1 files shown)


lpsinger (commented):
I don't think that this should be a top-level nav button. Where else could we put the link?

Courey:
This version puts it on the circulars archive index page as a button next to the change request button #2630 
closing in favor of that version.

PR #2633, opened 2024-10-18
Fix link for Einstein Probe example alert JSON

jak574:
# Description
The link to an example JSON alert for Einstein probe on this page: https://gcn.nasa.gov/missions/einstein-probe leads to a 404 due to an error in the linked filename. 

# Testing
Tested in local development environment.


lpsinger (approved):
Good catch!

PR #2640, opened 2024-10-22
Fermi notices

dakota002:
Shortened list of new fermi notices. Just the `gcn.notices.fermi.gbm` (alert is implied) and an enum can be used to determine the sub type of the alert


dakota002:
Somehow super k stuff got in there and i didn't notice, ignore this

Issue #2605, opened 2024-09-30
Fermi-GBM JSON Notices


jracusin:
# Description
GBM new public JSON notices.

# Acceptance criteria
- create topics/acls
- JSON schema
- mission page updates
- add to Quickstart
- announcement

dakota002:
Schema PR: https://github.com/nasa-gcn/gcn-schema/pull/199

dakota002:
Quickstart update (behind feature flag) https://github.com/nasa-gcn/gcn.nasa.gov/pull/2631

@shb46 are there more topics, or a more specific topic, that you would like listed? This PR will add `gcn.notices.fermi.gbm` to the list under the JSON option

shb46:
@dakota002: Our idea was each notice to have its own topic to enable subscription to individual types.  Let me know if it is a problem.  So the topic would be `gcn.notices.fermi.gbm.alert` .

dakota002:
> @dakota002: Our idea was each notice to have its own topic to enable subscription to individual types. Let me know if it is a problem. So the topic would be `gcn.notices.fermi.gbm.alert` .

Not a problem at all! Are there more than `gcn.notices.fermi.gbm.alert` that you would like added to the list?

shb46:
@dakota002: GBM produces 5 public notice types and a few internal ones.  Besides the alert The public ones are "flight position", "ground position",  "final position" and "subthreshold".  If you want to be consistent, these can be abbreviated as they appear in the email subscription page.

dakota002:
@shb46 are those going to all be distributed under the same topic with some flag denoting the difference, or each as individual topics, (ie. `gcn.notices.fermi.gbm.flight_position`, `gcn.notices.fermi.gbm.ground_position`, `gcn.notices.fermi.gbm.final_position`, etc)

shb46:
@dakota002: Individual topics.

Vidushi-GitHub:
@shb46 would you like to update the Fermi-GBM mission page (https://gcn.nasa.gov/missions/fermi) for new Notices?
You can fork the page from https://github.com/nasa-gcn/gcn.nasa.gov, app->route has the previous page. A new section titled as "JSON-Serialized GCN Notices Types" can be added.
As reference, you can see the IceCube update for GW follow-up or Einstein Probe mission page.

Otherwise, please send me the information that you used for the README file. I can create PR and ask for your feedback.

Vidushi-GitHub:
[...]

PR #2641, opened 2024-10-22
Reduce the fermi json notices list

dakota002:


lpsinger (commented):
Where is the discussion that led to this proposed change?

shb46:
@lpsinger: See #2605

shb46 (approved):

PR #2642, opened 2024-10-22
create synonym group for eventId if none exists

### LINKED ISSUE #2661: create synonym for new circulars automatically if one does not already exist

Courey:
# Description
This change adds functionality to check if there is an existing synonym group for an eventId upon circular creation. If there is an eventId and no synonym group for that eventId, it creates one. If there is no eventId it skips. If there is an eventId and the synonym group already exists for that eventId it skips.

# Related Issue(s)
Resolves #2661 

# Testing
I tested this locally. 
However, because it is in a critical area, we need to thoroughly test it on dev.

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -6,7 +6,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { tables } from '@architect/functions'
-import { PutItemCommand } from '@aws-sdk/client-dynamodb'
 import { type DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
 import { search as getSearchClient } from '@nasa-gcn/architect-functions-search'
 import crypto from 'crypto'
@@ -156,18 +155,19 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
 
 export async function tryInitSynonym(eventId: string) {
   const db = await tables()
-  const client = db._doc as unknown as DynamoDBDocument
-  const TableName = db.name('synonyms')
+
   try {
-    const params = {
-      TableName,
-      Item: {
-        synonymId: { S: crypto.randomUUID() },
-        eventId: { S: eventId },
+    await db.synonyms.update({
+      Key: { eventId },
+      UpdateExpression: 'set #synonymId = :synonymId',
+      ExpressionAttributeNames: {
+        '#synonymId': 'synonymId',
+      },
+      ExpressionAttributeValues: {
+        ':synonymId': crypto.randomUUID(),
       },
       ConditionExpression: 'attribute_not_exists(eventId)',
-    }
-    await client.send(new PutItemCommand(params))
+    })
   } catch (error) {
     if ((error as Error).name !== 'ConditionalCheckFailedException') throw error
   }
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -6,7 +6,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { tables } from '@architect/functions'
-import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
+import { PutItemCommand } from '@aws-sdk/client-dynamodb'
+import { type DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
 import { search as getSearchClient } from '@nasa-gcn/architect-functions-search'
 import crypto from 'crypto'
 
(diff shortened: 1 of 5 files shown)


lpsinger:
Why is this necessary? Is it because, currently, the grouped view does not show events that do not have an entry in the synonyms table?

Courey:
> Why is this necessary? Is it because, currently, the grouped view does not show events that do not have an entry in the synonyms table?

yup. Judy wanted to all events to have an entry, not just the manually created ones.

Courey:
And in addition to this change, I am aware that there needs to be a backfill to get things to work. I need a backfill to populate the eventId in circulars that existed before I added in the eventId on circular creation. I then need a backfill to populate the synonyms with the existing eventIds. I have combined both backfills into one for efficiency sake. This is the backfill script I have currently:
```
import { ReturnValue, paginateScan } from '@aws-sdk/client-dynamodb'
import { type DynamoDBDocument, UpdateCommand } from '@aws-sdk/lib-dynamodb'

import {
  type Circular,
  parseEventFromSubject,
} from '../circulars/circulars.lib'
import { createSynonyms, synonymExists } from '../synonyms/synonyms.server'

async function backfill() {
  console.log('Starting backfill...')
  const db = await tables()
  const client = db._doc as unknown as DynamoDBDocument
  const TableName = db.name('circulars')
  const pages = paginateScan(
    { client },
    {
      TableName,
    }
  )
    page.Items?.forEach(async (record) => {
      const circular = record as unknown as Circular
      const validEvent =
        circular.eventId || parseEventFromSubject(circular.subject)
      if (!circular.eventId && validEvent) {
        const updateParams = {
          TableName,
          Key: {
          },
          UpdateExpression: 'SET eventId = :eventId',
          ExpressionAttributeValues: {
            ':eventId': validEvent,
          },
          ReturnValues: ReturnValue.UPDATED_NEW,
        }

        try {
          const result = await client.send(new UpdateCommand(updateParams))
          console.log(
            `Circular ${circular.circularId} updated successfully: `,
            result.Attributes
          )
        } catch (error) {
          console.error(
            `ERROR updating Circular ${circular.circularId}: `,
            error
          )
          throw error
        }
        await db.circulars.get({ circularId: circular.circularId })
      }
      if (validEvent && !(await synonymExists({ eventId: validEvent }))) {
        try {
          const uuid = await createSynonyms([validEvent])
          console.log(
            'Created synonym',
            JSON.stringify({ eventId: validEvent, synonymId: uuid })
          )
        } catch (error) {
          console.error(`ERROR creating synonym ${validEvent}: `, error)
          throw error
        }
      }
    })
  }
  console.log('... End backfill')
}
backfill()
```

Courey:
okay. I'll not make draft PRs for things I'm still working on in the future. This is not ready for review at all.

lpsinger:
> okay. I'll not make draft PRs for things I'm still working on in the future.

I wouldn't say that! Anyone is always welcome to make draft PRs.

> This is not ready for review at all.

I wasn't offering a review; I was just asking for the motivation for this PR, since I missed the Monday call.

lpsinger (changes requested):
Please use [TransactWriteItems](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html#transaction-apis-txwriteitems) to eliminate the race condition.

app/routes/synonyms/synonyms.server.ts
```diff
@@ -167,6 +154,25 @@ export async function createSynonyms(synonymousEventIds: string[]) {
...
+      Item: {
+        synonymId: { S: crypto.randomUUID() },
+        eventId: { S: eventId },
+      },
+      ConditionExpression: 'attribute_not_exists(eventId)',
+    }
+    await client.send(new PutItemCommand(params))
+  } catch (error) {
+    if ((error as Error).name !== 'ConditionalCheckFailedException') throw error
+  }
```
lpsinger, line 173: I think that you can use the built-in Architect convenience function:

```ts
await db.synonyms.update({...})
```

so that you don't have to look up the actual table name, and so that the document client marshals the types for you (`{ synonymId: crypto.randomUUID() }` rather than `{ synonymId: { S: crypto.randomUUID() } }`).

lpsinger (changes requested):

Courey:
I made a couple of updates here.
1. instead of a hard delete when a moderator removes a synonym from a group, it now will just assign the removed one a unique id to maintain the expectation that all events will have a synonym group.
2. Instead of deleting a synonym group when it is deleted, each member will be given a new uuid so they all have their own synonym group.
3. the initial creation of a synonym has different rules than one that is created by a moderator. Instead of using a transaction (which costs 2x), I could just get away with using a condition check on the put and swallowing the failure if it fails that check. That enabled me to remove any check for existing synonyms and avoid more expensive transactions.
4. I re-named the createSynonyms function to moderatorCreateSynonyms since the behavior differs depending on how the synonym is created. This ensures that developers will know to distinguish between moderator created and system initiated.

lpsinger (approved):

PR #2675, opened 2024-11-06
Enable OpenTelemetry tracing for all Lambdas except for web app

lpsinger:
OpenTelemetry aggregates logs across multiple AWS services to form distributed traces.

Enable it for all Lambdas except for the one that powers our web app, because the OpenTelemetry layer increases cold-start time.


lpsinger:
The UX of AWS X-Ray is so terrible that I'm not sure that this is all that useful without a third-party OTEL collector and trace visualizer.

PR #2736, opened 2024-11-26
DO NOT MERGE

### LINKED ISSUE #2647: Backfill eventID in Circulars
And in addition to this change, I am aware that there needs to be a backfill to get things to work. I need a backfill to populate the eventId in circulars that existed before I added in the eventId on circular creation. I then need a backfill to populate the synonyms with the existing eventIds. I have combined both backfills into one for efficiency sake. This is the backfill script I have currently:
```
import { ReturnValue } from '@aws-sdk/client-dynamodb'
import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
import { paginateScan, UpdateCommand } from '@aws-sdk/lib-dynamodb'

import {
  type Circular,
  parseEventFromSubject,
} from '~/routes/circulars/circulars.lib'
import { createSynonyms, synonymExists } from '~/routes/synonyms/synonyms.server'

export async function backfill() {
  console.log('Starting backfill...')
  const db = await tables()
  const client = db._doc as unknown as DynamoDBDocument
  const TableName = db.name('circulars')
  const pages = paginateScan(
    { client },
    {
      TableName,
    }
  )
    for (const record of page.Items || []) {
      const circular = record as unknown as Circular
      const validEvent =
        circular.eventId || parseEventFromSubject(circular.subject)
      const promises: Promise<any>[] = []
      if (!circular.eventId && validEvent) {
        const updateParams = {
          TableName,
          Key: {
          },
          UpdateExpression: 'SET eventId = :eventId',
          ExpressionAttributeValues: {
            ':eventId': validEvent,
          },
          ReturnValues: ReturnValue.UPDATED_NEW,
        }

        promises.push(client.send(new UpdateCommand(updateParams)))
      }
      if (validEvent && !(await synonymExists({ eventId: validEvent }))) promises.push(createSynonyms([validEvent]))
      if(promises.length >= 1){
        try {
          const results = await Promise.all(promises)
          const data = await Promise.all(results)
          data[0].Attributes ? console.log(data[0].Attributes) : console.log(data[0])
          if(data[1]) console.log(data[1])
          console.log("--------")
        } catch (error) {
          console.error(
            `ERROR updating Circular ${circular.circularId}: `,
            error
          )
          throw error
        }
      }
    }
    console.log('... End backfill')
  }
}
await backfill()
```

_Originally posted by @Courey in https://github.com/nasa-gcn/gcn.nasa.gov/issues/2642#issuecomment-2429931625_

Courey:
DO NOT MERGE. Please put approval in comment. ticket will be closed once approved.
Resolves #2647


app/backfill-eventId.js
```diff
@@ -0,0 +1,125 @@
+import { DynamoDBClient, paginateScan } from '@aws-sdk/client-dynamodb'
+import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm'
+import { BatchWriteCommand } from '@aws-sdk/lib-dynamodb'
+import { unmarshall } from '@aws-sdk/util-dynamodb'
+import chunk from 'lodash/chunk.js'
+
+// THIS CODE IS CUT AND PASTED FROM THE APP CODE IN /routes/circulars/circulars.lib.ts
+// running as a node script it would not import the module from the file
```
lpsinger, line 8: I'm not comfortable copying and pasting this. If you are not able to import it from a standalone Node.js script, then please convert this to a `.ts` source and use esbuild to compile it down to JS. You should be able to hack something into https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/esbuild.config.js to do that.

app/backfill-eventId.js
```diff
@@ -0,0 +1,125 @@
...
+          RequestItems: {
+            [TableName]: writes.map((circ) => ({
+              PutRequest: {
+                Item: {
+                  ...circ,
+                },
+              },
+            })),
+          },
+        })
```
lpsinger, line 113: This is quite complicated. I would be more comfortable with the correctness of a functional programming implementation using [TC39 Async Iterator Helpers](https://github.com/tc39/proposal-async-iterator-helpers) (see polyfills [available from core-js](https://github.com/zloirock/core-js?tab=readme-ov-file#asynciterator-helpers)).

lpsinger (changes requested):

Courey:
I wish these major changes had been requested when I first asked for in depth review _before_ I started running out of test data. I can still run this script on test, there is still data there to test on, but I am totally out of data pertaining to the synonym slug backfill, all that is left is production data. This will bloat the time for this ticket.

app/backfill-eventId.js
Courey: can you explain to me why? It's a literal cut and paste (though I will have to adapt the GRB matcher to include ones without a letter). Copying and pasting directly from the current app code and updating that subject matcher shouldn't be an issue. I'm not sure what compiling etc would buy us over a copy and paste. I also have the issue that I need to include eventIds on circulars that have GRBs without a letter, but they are not permissible anymore, so it shouldn't be added to the app subject matchers. I'm not sure what hacking around in the esbuild would get us that would be any better

jracusin: This script is standalone and will only be run once on prod.  It also requires changes compared to the regular subject matchers to account for changes in GRB naming over time.  If there are no problems with the logic of how this is being implemented, let's just go with how this is currently implemented.

lpsinger: > can you explain to me why? It's a literal cut and paste (though I will have to adapt the GRB matcher to include ones without a letter). Copying and pasting directly from the current app code and updating that subject matcher shouldn't be an issue. I'm not sure what compiling etc would buy us over a copy and paste. I also have the issue that I need to include eventIds on circulars that have GRBs without a letter, but they are not permissible anymore, so it shouldn't be added to the app subject matchers. I'm not sure what hacking around in the esbuild would get us that would be any better

For the same reasons why we try to avoid duplicate copies of all but the simplest code in general: there is the risk that the two copies will not be the same, that bugs in one won't be fixed in the other, that they will become incompatible, etc. Even though this PR contains code that we do not plan to reuse, there is still the very real risk that the new copy in the PR will not be the same as the original copy in the main source code.

We have added new regular expressions several times while this PR has been open. In order to review this properly, one would need to manually check that the two copies are the same in this PR, and then we would need to re-check it any time one makes changes to either this PR or to the original code on the main branch.

I do not think that this is an acceptable risk, but @jracusin should have the final say.

Courey: I think it's fine. I can just check the subject matchers before I run them. I have both in my current version. I also have to adapt the GRB matcher so it includes GRBs without a letter in them. Currently the subject matchers will only match on GRBs with a letter. So there is already going to be a difference than what would be imported.

app/backfill-eventId.js
lpsinger: There are a lot of loops and conditionals here, which makes the code too complicated to review quickly.

As for whether we have existing code like this, I don't know of any other examples where we have a paginated input, which is filtered, and then which must be paginated again for insertion. That is the source of the complexity.

app/backfill-eventId.js
lpsinger: Here's how to modify the esbuild script to build out the additional script.

```
diff --git a/esbuild.config.js b/esbuild.config.js
index 463a36ae..6130b007 100644
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -7,7 +7,7 @@ import { basename, dirname, join } from 'path'
 const args = process.argv.slice(2)
 const dev = args.includes('--dev')
 const entryPoints = await glob(
-  './app/{email-incoming,scheduled,table-streams}/*/index.ts'
+  './app/{email-incoming,scheduled,table-streams,migrations}/*/index.ts'
 )
 
 /**
@@ -19,8 +19,9 @@ const options = {
   logLevel: 'info',
   outdir: 'build',
   outbase: 'app',
-  outExtension: { '.js': '.cjs' },
+  outExtension: { '.js': '.mjs' },
   external: ['@aws-sdk/*', 'aws-sdk'],
+  format: "esm",
   platform: 'node',
   target: ['node20'],
   minify: !dev,
```

Then put your script in `app/migrations/event-id/index.ts`, run `npm run build:esbuild`, and then execute the script by running `node build/migrations/event-id/index.mjs`.

app/backfill-eventId.js
lpsinger: So, the core-js polyfills would work just fine, but they won't be necessary if we make the following changes:

- Overwrite all event IDs, regardless of whether or not they are already populated.
- Process all events at once. No need to do it in small batches spread over days.

I discussed both of these changes with @jracusin, and she agrees.

PR #2759, opened 2024-12-08
Report Lambda sizes in GitHub workflow summary

lpsinger:


lpsinger:
See, for example, https://github.com/nasa-gcn/gcn.nasa.gov/actions/runs/12224330380/attempts/3#summary-34097993995.

PR #2775, opened 2024-12-10
removing unused warning

Courey:
This change removes a noisy warning that is not very useful.


lpsinger (changes requested):
Please transform this:

```ts
 if (!result) {
     if (process.env.NODE_ENV === 'production') {
       ...
     }
   }
```

to this:

```ts
 if (!result && process.env.NODE_ENV === 'production') {
     dieForEnv(key)
   }
```

lpsinger (approved):

PR #2785, opened 2024-12-13
Backfill synonym slugs (DO NOT MERGE)

### LINKED ISSUE #2784: backfill existing synonyms to add slugs
This needs to go through all the existing synonyms and add a slug where there is none.

Courey:
DO NOT MERGE!!! for review only
closes #2784


app/backfill-synonym-slugs.ts
```diff
@@ -0,0 +1,55 @@
...
+import { paginateScan } from '@aws-sdk/lib-dynamodb'
+import { slug } from 'github-slugger'
+
+import type { Synonym } from '~/routes/synonyms/synonyms.lib'
+
+function* chunks<T>(arr: T[], n: number): Generator<T[], void> {
+  for (let i = 0; i < arr.length; i += n) {
+    yield arr.slice(i, i + n)
+  }
+}
```
lpsinger, line 12: You could use [lodash's chunk](https://lodash.com/docs/4.17.15) here.

app/backfill-synonym-slugs.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export async function backfillSynonyms() {
+  console.log('Starting SYNONYM SLUG backfill...')
+  const db = await tables()
+  const client = db._doc as unknown as DynamoDBDocument
+  const TableName = db.name('synonyms')
```
lpsinger, line 18: Architect's service discovery will not work in this context. You'll need to look up the name of the table using Amazon SSM.

lpsinger (changes requested):

app/backfill-synonym-slugs.js
```diff
@@ -0,0 +1,72 @@
...
+import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm'
+import { BatchWriteCommand } from '@aws-sdk/lib-dynamodb'
+import { unmarshall } from '@aws-sdk/util-dynamodb'
+import { slug } from 'github-slugger'
+
+function* chunks(arr, n) {
+  for (let i = 0; i < arr.length; i += n) {
+    yield arr.slice(i, i + n)
+  }
+}
```
lpsinger, line 11: This is probably correct, but I'd use [lodash's `chunk`](https://lodash.com/docs/#chunk).

app/backfill-synonym-slugs.js
```diff
@@ -0,0 +1,72 @@
...
+
+async function getTableNameFromSSM(dynamoTableName) {
+  const ssmClient = new SSMClient({ region: 'us-east-1' })
```
lpsinger, line 14: You shouldn't have to explicitly specify the region.

```suggestion
  const ssmClient = new SSMClient()
```

app/backfill-synonym-slugs.js
```diff
@@ -0,0 +1,72 @@
...
+    throw error
+  }
+}
+
+export async function backfillSlugsOnSynonyms() {
+  const startTime = new Date()
+  console.log('Starting SYNONYM SLUG backfill... ', startTime)
+  const dynamoTableName = '/RemixGcnProduction/tables/synonyms'
+  const TableName = await getTableNameFromSSM(dynamoTableName)
+  const client = new DynamoDBClient({ region: 'us-east-1' })
```
lpsinger, line 36: Again, you shouldn't have to specify the region explicitly.

lpsinger (changes requested):

Courey:
so I tried it without the region and it doesn't work without a region. Just for future reference. I added it back in and was able to run things fine.

lpsinger:
> so I tried it without the region and it doesn't work without a region. Just for future reference. I added it back in and was able to run things fine.

Interesting. I think I have `AWS_REGION` set in my environment; maybe that made a difference. Not important, though.

app/backfill-synonym-slugs.js
```diff
@@ -0,0 +1,67 @@
...
+
+    if (!response.Parameter?.Value) {
+      throw new Error('dynamoTableName not found in SSM')
+    }
+
+    return response.Parameter.Value
+  } catch (error) {
+    console.error('Error fetching table name from SSM:', error)
+    throw error
+  }
```
lpsinger, line 23: FWIW, you could safely remove this entire `try`/`except` block. You'll get a good stack trace either way.

app/backfill-synonym-slugs.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export async function backfillSlugsOnSynonyms() {
+  const startTime = new Date()
+  console.log('Starting SYNONYM SLUG backfill... ', startTime)
```
lpsinger, line 28: Consider [`console.time`](https://developer.mozilla.org/en-US/docs/Web/API/console/time_static) and [`console.timeEnd`](https://developer.mozilla.org/en-US/docs/Web/API/console/timeEnd_static).

Courey:
There is no more test data to test this script. The only env left with data is prod.

lpsinger:
> There is no more test data to test this script. The only env left with data is prod.

OK, what's your new test strategy? If we could only ever test this once, then perhaps we did not have an adequate test strategy.

Issue #2877, opened 2025-02-10
Event view sort in reverse chronological order


jracusin:
# Current behavior
Event view currently sorts randomly.

# Expected behavior
Events should be sorted in reverse chronological order using the date of the most recent circular in each synonym group.

# Steps to reproduce
Event view on dev/test.

jracusin:
Superseded by #2898

PR #2882, opened 2025-02-11
Synonym sort

dakota002:
# Description
Sorts the synonym view by reverse chronological order of circulars

image

# Related Issue(s)
Resolves: #2877

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -337,7 +337,10 @@ export async function groupMembersByEventId({
     const members = (await Promise.all(promises)).flat()
     return { ...group, members }
   })
-  const items = await Promise.all(groupedItems)
+  const items = (await Promise.all(groupedItems)).sort(
+    (a, b) => b.members[0].createdOn - a.members[0].createdOn
+  )
+
   return {
     items,
     totalItems: searchResults.totalItems,
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -347,14 +347,19 @@ export async function groupMembersByEventId({
 
 async function getSynonymMembers(eventId: string) {
   const db = await tables()
-  const { Items } = await db.circulars.query({
-    IndexName: 'circularsByEventId',
-    KeyConditionExpression: 'eventId = :eventId',
-    ExpressionAttributeValues: {
-      ':eventId': eventId,
-    },
-  })
-  return Items as Circular[]
+  const Items: Circular[] = (
+    await db.circulars.query({
+      IndexName: 'circularsByEventId',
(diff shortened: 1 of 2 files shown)


jracusin:
This looks good, but I'm wondering if the design of sorting by the date (or circ ID) of the most recent circular is going to be intuitive to users.  If we sorted by the date (or circ ID) of the first circular in the group, it would be a proxy for the date in the eventID. Thoughts?

dakota002:
> If we sorted by the date (or circ ID) of the first circular in the group, it would be a proxy for the date in the eventID. Thoughts?

Do you mean the first circular bring the oldest/first circular to be associated to that group? If so, then I believe the order of the groups would be static, whereas in this view, the top event will be the most recently updated.

I like this view option, since the descending order of the circulars is still preserved, but I can see how either would be useful

app/routes/synonyms/synonyms.server.ts
```diff
@@ -347,14 +350,19 @@ export async function groupMembersByEventId({
...
+      KeyConditionExpression: 'eventId = :eventId',
+      ExpressionAttributeValues: {
+        ':eventId': eventId,
+      },
+    })
+  ).Items
+
+  Items.sort((a, b) => b.createdOn - a.createdOn)
+
+  return Items
```
lpsinger, line 365: Could we just set the sort key on this index?

lpsinger (changes requested):

app/routes/synonyms/synonyms.server.ts
dakota002: On the circulars table?

lpsinger: On this index for the circulars table.

Also, I don't think that this addresses #2877. The problem that issue describes is that the event names are in arbitrary order.

I think that this PR puts the circular IDs in chronological order _within an event_, which is also important, but does not address the issue.

dakota002: It does both, assuming we want to show the group with the most recent circular at the top

jracusin: We want the circulars in descending order within a group as this does.  But I'd like to also see the groups/events also in reverse order of date.  I thought the way to do that was to sort the groups in reverse order of the first circular in that group, which should be a good proxy for date of the event.

lpsinger: > It does both, assuming we want to show the group with the most recent circular at the top

How does it sort the event names chronologically?

lpsinger (changes requested):
- Sort by circular ID, not date.
- Do the sort in `getAllSynonymMembers`, not `getSynonymMembers`.

lpsinger:
@Courey, did your date indexing already address this?

PR #2903, opened 2025-02-19
Adds svom voevent topics

dakota002:
Adds SVOM vo event behind feature flag

### DIFF
--- a/app/components/NoticeTypeCheckboxes/NoticeTypeCheckboxes.tsx
+++ b/app/components/NoticeTypeCheckboxes/NoticeTypeCheckboxes.tsx
@@ -229,7 +229,8 @@ export function NoticeTypeCheckboxes({
   if (useFeature('SVOM_QUICKSTART')) {
     JsonNoticeTypes.SVOM = ['gcn.notices.svom']
     JsonNoticeTypeLinks.SVOM = '/missions/svom'
-
+  }
+  if (useFeature('SVOM')) {
     if (selectedFormat == 'voevent') {
       NoticeTypes['SVOM'] = [
         'gcn.notices.svom.voevent.grm',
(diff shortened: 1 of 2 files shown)


app/components/NoticeTypeCheckboxes/NoticeTypeCheckboxes.tsx
```diff
@@ -229,6 +229,14 @@ export function NoticeTypeCheckboxes({
   if (useFeature('SVOM_QUICKSTART')) {
     JsonNoticeTypes.SVOM = ['gcn.notices.svom']
     JsonNoticeTypeLinks.SVOM = '/missions/svom'
```
jracusin, line 231: Since there currently aren't any SVOM JSON notices, does removing this feature flag also add SVOM to the JSON list?

app/lib/utils.ts
```diff
@@ -11,7 +11,7 @@ export function formatAndNoticeTypeToTopic(
   noticeFormat: string,
   noticeType: string
 ) {
-  return noticeFormat == 'json'
+  return noticeFormat == 'json' || noticeType.startsWith('gcn.notices')
```
jracusin, line 14: Does this affect IGWN?

dakota002: no, their notice type is already JSON

app/components/NoticeTypeCheckboxes/NoticeTypeCheckboxes.tsx
dakota002: Yea, is there a different feature flag this should be behind (if one at all?)

jracusin: Feature flag `SVOM` is used elsewhere, like on the mission page.  Let's put the JSON under the flag `SVOM_JSON`, so we can track it separately from the VOEvent, mission page, and announcement that we're very close to releasing.

jracusin (approved):

lpsinger (approved):

PR #2906, opened 2025-02-20
add SVOM source name subject matcher

jracusin:
# Description
SVOM has recently started using sbYYMMDDFF source designations in their Circulars.  We need to add this to the subject matcher to get the correct eventID.

e.g. https://gcn.nasa.gov/circulars/39363

# Related Issue(s)

### DIFF
--- a/__tests__/circulars.ts
+++ b/__tests__/circulars.ts
@@ -73,6 +73,7 @@ describe('parseEventFromSubject', () => {
   const ztfEvent = 'ZTF23aabmzlp'
   const epEvent = 'EP241119a'
   const frbEvent = 'FRB 20250206A'
+  const svomEvent = 'sb25021804'
 
   test('handles nonsense subject cases', () => {
     expect(parseEventFromSubject('zawxdrcftvgbhnjm')).toBe(undefined)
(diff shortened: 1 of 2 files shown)


lpsinger (changes requested):
Please add a test case to the unit tests.

Courey:
It should be not a big deal to add those tests. 
just add it [here](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/__tests__/circulars.ts#L76)
then just copy one of the blocks, maybe the FRB block that was just added [here](https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/__tests__/circulars.ts#L577) and then adapt those cases to this subject matcher.

__tests__/circulars.ts
```diff
@@ -623,6 +624,56 @@ describe('parseEventFromSubject', () => {
...
+      expect(parseEventFromSubject(svomSubjectWithUnderscore)).toBe(svomEvent)
+    })
+
+    test('handles SVOM event name with a hyphen', () => {
+      const svomSubjectWithHyphen =
+        'SVOM/sb-25021804: SVOM detection of a long X-ray transient'
+      expect(parseEventFromSubject(svomSubjectWithHyphen)).toBe(svomEvent)
+    })
+
+    test('handles FRB event name with a hyphen in misc positions', () => {
```
Courey, line 671: there is an FRB still here

Courey (approved):

lpsinger (approved):

PR #2930, opened 2025-03-05
Circulars: Implement query_string method with fallback

### LINKED ISSUE #2288: Circulars: add support for Lucene-style search queries in archive

tylerbarna:
# Description
This changes the search method to `query_string` rather than `simple_query_string`. The former addresses all of the use cases outlined in #2288, but there is the drawback of it not being as tolerant to incorrect syntax as the latter. As a result, I implemented a fallback where opensearch switches to `multi_match` if the query is invalid for `query_string`. 

Note: this is still behind the `CIRCULARS_LUCENE` feature flag

# Related Issue(s)
Part of #2288

# Testing
1. Add `CIRCULARS_LUCENE` flag to `.env` file
2. Navigate to the circulars archive
3. Submit a standard search for something like `LIGO`. This should return the results as expected.
4. Submit a lucene search for something like `subject:LIGO AND subject:Swift` . The only results should include circulars with both "LIGO" and "Swift" in the subject.
5. Submit a lucene search for something like `submitter:LIGO`. This should only return a couple of circulars where the submitter has their LIGO association included.
6. Submit a lucene search for something like `submitter:LIGO AND`. Rather than failing due to the incorrect lucene syntax, it should fall back onto the `multi_match` method and show results different from the previous test. There will also be a warning above the search box alerting users of their error


tylerbarna:
I've added a warning when users submit a query that run afoul of lucene syntax
Screenshot 2025-03-11 at 14 24 41

app/routes/circulars/circulars.server.ts
```diff
@@ -213,7 +210,34 @@ export async function search({
       size: limit,
       track_total_hits: true,
     },
-  })
+  }
+
+  let searchResult
+  try {
+    searchResult = await client.search(searchBody)
+  } catch (error) {
```
lpsinger, line 218: There are many other kinds of errors here that are not related to invalid input. For example, OpenSearch might just time out. How do we make sure that we are taking this code path only when the search fails because the input is invalid?

app/routes/circulars/circulars.server.ts
```diff
@@ -213,7 +210,34 @@ export async function search({
...
+    console.error(
+      'Search with queryObj failed, falling back to multi_match',
+      error
+    )
```
lpsinger, line 222: Although this is is a user error, it doesn't suggest anything wrong with our code, so we shouldn't log it.

```suggestion
```

app/routes/circulars._archive._index/route.tsx
```diff
@@ -250,6 +251,12 @@ export default function () {
           Synonym Moderation
         </Link>
       )}
+      {queryFallback && (
+        <Hint id="searchFailedHint" style={{ color: 'red' }}>
+          "{query}" does not adhere to lucene search syntax. Please refer to the
+          documentation and try again.
```
lpsinger, line 257: Link to the documentation.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -250,6 +251,12 @@ export default function () {
           Synonym Moderation
         </Link>
       )}
+      {queryFallback && (
+        <Hint id="searchFailedHint" style={{ color: 'red' }}>
+          "{query}" does not adhere to lucene search syntax. Please refer to the
```
lpsinger, line 256: Please make sure that we are using consistent terminology. Do we call this "Lucene search syntax" (note capitalization) or "advanced search syntax?"

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
tylerbarna: Would it work to parse the error string to check if it contains something like `Reason: Failed to parse query` and only do the fallback if that is true?

lpsinger: That would probably work, but first check if the OpenSearch client provides a type that you can recognize.

lpsinger: Also note that there is an updated version of the OpenSearch client library that may have stronger typing.

dakota002:
Since you have two functions returning the same object, you can probably just define it as a new type and return it from both `groupMembersByEventId` and `search`:
```ts
type SomeName{
  items: CircularMetadata[] | SynonymGroupWithMembers[]
  totalPages: number
  totalItems: number
  queryFallback: boolean
}
```

tylerbarna:
> Since you have two functions returning the same object, you can probably just define it as a new type and return it from both `groupMembersByEventId` and `search`:
> 
> ```ts
> type SomeName{
>   items: CircularMetadata[] | SynonymGroupWithMembers[]
>   totalPages: number
>   totalItems: number
>   queryFallback: boolean
> }
> ```

I agree that would be neater, but the current structure would require cross-imports between `circulars.lib.ts` and `synonyms.lib.ts` in a way that feels a bit messy. I think it would make sense to refactor the latter into the former, but that should probably be a separate PR.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -314,6 +326,7 @@ export default function () {
           />
         )}
       </ToolbarButtonGroup>
+
```
dakota002, line 329: You can remove this new line

app/routes/circulars/circulars.server.ts
```diff
@@ -153,9 +153,12 @@ export async function search({
   items: CircularMetadata[]
   totalPages: number
   totalItems: number
+  queryFallback: boolean
 }> {
   const client = await getSearch()
 
+  var queryFallback = false
```
Courey, line 160: just out of curiosity, why is this a var instead of a let?

tylerbarna: no particular reason, changing to a let now

lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
```diff
@@ -213,7 +210,34 @@ export async function search({
...
       track_total_hits: true,
     },
-  })
+  }
+
+  let searchResult
+  try {
+    searchResult = await client.search(searchBody)
+  } catch (error) {
+    if ((error as typeof Error).toString().includes('Failed to parse query')) {
```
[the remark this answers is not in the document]
tylerbarna: I've updated the most recent commit to make use of the opensearch typed error classes, but upon reviewing some other implementations, it seems the ResponseError type property 'search_phase_execution_exception' and the root_cause type property 'query_shard_exception" are not specific to cases where a query string parsing error is raised (see https://github.com/opensearch-project/OpenSearch/issues/1513). I've retained the usage of the `includes` method barring there being a more precise way of identifying the specific error we're checking for.

app/routes/circulars/circulars.server.ts
```diff
@@ -17,6 +17,7 @@ import {
   DynamoDBAutoIncrement,
   DynamoDBHistoryAutoIncrement,
 } from '@nasa-gcn/dynamodb-autoincrement'
+import type { errors } from '@opensearch-project/opensearch'
```
lpsinger, line 20:
```suggestion
import { errors } from '@opensearch-project/opensearch'
```

app/routes/circulars/circulars.server.ts
```diff
@@ -216,7 +217,11 @@ export async function search({
   try {
     searchResult = await client.search(searchBody)
   } catch (error) {
-    if ((error as typeof Error).toString().includes('Failed to parse query')) {
+    if (
+      (error as errors.ResponseError).body.error.root_cause[0].reason.includes(
+        'Failed to parse query'
+      )
```
lpsinger, line 223: Please avoid using [type assertions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions). Instead, actually test if the object has the required type.

```suggestion
      error instanceof errors.ResponseError && error.message.includes(
        'Failed to parse query'
      )
```

lpsinger (changes requested):

lpsinger (commented):
I am trying to test this out locally. Can you suggest a search string that will fail to parse and trigger the fallback code path?

tylerbarna:
> I am trying to test this out locally. Can you suggest a search string that will fail to parse and trigger the fallback code path?

anything with 'OR' or 'AND' without following it with another string. I've been testing with "subject:GRB OR"

app/routes/circulars._archive._index/route.tsx
```diff
@@ -250,6 +251,17 @@ export default function () {
           Synonym Moderation
         </Link>
       )}
+      {queryFallback && (
+        <Hint id="searchFailedHint" style={{ color: 'red' }}>
```
lpsinger, line 255: Please use the [USWDS error label element](https://trussworks.github.io/react-uswds/?path=/docs/components-form-elements-formgroup--docs#text-input-error-form-group) here, and make sure that you set the `htmlFor` element to link the label to the search box for accessibility purposes.

lpsinger (changes requested):

lpsinger (changes requested):
I don't think that we need the title. Just the error text. And it should appear below the search bar, not above it.

tylerbarna:
> I don't think that we need the title. Just the error text. And it should appear below the search bar, not above it.

like this?
image

lpsinger:
> > I don't think that we need the title. Just the error text. And it should appear below the search bar, not above it.
> 
> like this?

Yes, that's good. How about that error text is _immediately_ below the search bar, before even the hints?

tylerbarna:
image

app/routes/circulars._archive._index/route.tsx
```diff
@@ -351,6 +354,16 @@ export default function () {
         )}
       </Hint>
       {useFeature('CIRCULARS_LUCENE') && <LuceneAccordion />}
+      {queryFallback && (
+        <ErrorMessage id="searchFailedHint">
```
lpsinger: Please use the Label element and include the `htmlFor` attribute for accessibility purposes.

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
tylerbarna: Should I be wrapping the ErrorMessage element in Label element or should I just have it immediately before the ErrorMessage element? If the latter, is there specific text I should use in the label given that it's required?

tylerbarna: Alternatively, I can set the children as undefined

lpsinger: Sorry, I misread the documentation. Disregard the comment about the label.

app/routes/circulars._archive._index/route.tsx
```diff
@@ -335,6 +338,22 @@ export default function () {
...
+            the{' '}
+            <Link
+              className="usa-link"
+              to="/docs/circulars/archive#advanced-search"
+            >
+              documentation
+            </Link>{' '}
+            and try again.
+          </ErrorMessage>
+        </>
```
lpsinger, line 355:
```suggestion
          <ErrorMessage>
            "{query}" does not adhere to advanced search syntax. Please refer to
            the{' '}
            <Link
              className="usa-link"
              to="/docs/circulars/archive#advanced-search"
            >
              documentation
            </Link>{' '}
            and try again.
          </ErrorMessage>
```

and then rerun prettier.

lpsinger (changes requested):

lpsinger (approved):

PR #2939, opened 2025-03-13
Courey/backfills all

Courey:
DO NOT MERGE!

This is both backfills along with the esbuild config changes to support building them.


app/migrations/synonyms/index.ts
```diff
@@ -0,0 +1,101 @@
...
+            ':eventId': { S: circular.eventId },
+          },
+        })
+
+        const response = await client.send(command)
+        const items = (response.Items || []).map((item) => ({
+          circularId: item.circularId?.S,
+          eventId: item.eventId?.S,
+          createdOn: item.createdOn?.N,
+        }))
```
lpsinger, line 60: Use `docClient` so that you don't have to marshal and unmarshal types yourself.

app/migrations/synonyms/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        const initialDate = orderBy(items, ['circularId'], ['asc'])[0].createdOn
```
lpsinger, line 62: Use `minBy` which is O(N) instead of `orderBy` which is O(N log N).

app/migrations/synonyms/index.ts
```diff
@@ -0,0 +1,101 @@
...
+          throw Error
+        }
+        writes.push({
+          eventId: circular.eventId,
+          initialDate: parseInt(initialDate),
+        })
+      }
+    }
+    if (writes.length > 0 && TableName) {
+      const dedupedWrites = uniqBy(writes, 'eventId')
```
lpsinger, line 74: Why wouldn't this already be unique?

lpsinger (changes requested):

app/migrations/synonyms/index.ts
Courey: before I added the set to ensure that it only ran one event once, if in the same page there were multiples, it would break the write because the write can't have duplicates. So that was just vestigial.

lpsinger: So you can remove this now?

app/migrations/synonyms/index.ts
Courey: If I remove the type, you can see that it is now storing an AttributeValue
Screenshot 2025-03-24 at 3 43 07 PM

lpsinger: That's because you imported `QueryCommand` from `@aws-sdk/client-dynamodb`. You need to import it from `@aws-sdk/lib-dynamodb`.

lpsinger (approved):

lpsinger (approved):

PR #2982, opened 2025-04-07
Courey/fix synonym search

### LINKED ISSUE #2981: Synonym moderation search does not return results
# Current behavior
In the Synonym Moderator interface, search bar does not limit the results to less than the full eventId table.

# Expected behavior
Search conducts either a text search or an exact search of eventIds.

# Steps to reproduce
On prod, dev, or test, as a signed in moderator, click "Synonym Moderation" interface on Circulars page, type something into search bar, and find it does not perform a search.

# Environment
Safari 18.3.1 on Sequoia 15.3.2

Courey:
# Description
Adding sorting behavior wiped out the relevance search. When you are not searching for a match, the results should be sorted by the initialDate, but when you are using a search term, it should sort by relevance. 

This branch changes the behavior of the search so that it does NOT sort by initialDate when using a query. Instead it will sort by relevance. When a query is not provided, the results are sorted by initialDate (as there is no relevance to sort by).

I fine tuned the results for a query search because the fuzziness search was not returning them in what I would see as an appropriate order. by changing it to a wildcard regex match search, it is returning only results that fit the query and in an appropriate order of relevance. 

# Related Issue(s)

Resolves #2981

# NOTE:
the sort order when sorting by initialDate is still not exact. I am working on this and will include improvements to that in a different branch. I made minimal improvements by making my sort more explicit in its behavior. This did not resolve the issue fully (though did help some).
i added the following fields to the sort for that:
```          
missing: '_last',
unmapped_type: 'long',
```
So the order of the index is slightly off, but not grossly off.

I want to unblock Judy and Vidushi, so further improvements will be in a different branch.

# Testing
This has been tested locally.
Screenshot 2025-04-07 at 4 33 19 PM
Screenshot 2025-04-07 at 4 33 13 PM

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -61,54 +61,29 @@ export async function searchSynonymsByEventId({
 }> {
   const client = await getSearchClient()
   const body: any = {
-    query: {},
+    query: {
+      bool: {},
+    },
   }
 
   if (eventId) {
-    body.query = {
-      function_score: {
-        query: {
-          bool: {
-            should: [
-              {
-                wildcard: {
-                  'eventIds.keyword': {
-                    value: `*${eventId}*`,
-                  },
-                },
-              },
-            ],
-            minimum_should_match: 1,
+    body.query.bool.filter = [
+      {
+        regexp: {
+          'eventIds.keyword': {
+            value: `.*${eventId}.*`,
+            case_insensitive: true,
           },
         },
-        functions: [
-          {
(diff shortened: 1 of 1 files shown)


jracusin:
@Courey, you said "When you are not searching for a match, ...", but I was searching for a match and it didn't seem to provide any query. Was there any way in which what is currently deployed (prior to this PR) would provide a search result?

lpsinger (changes requested):
Looks good, but please add some code comments.

PR #2988, opened 2025-04-08
normalize Virgo in subject matcher

### LINKED ISSUE #2984: LIGO/Virgo event names are not normalized to a uniform case
# Current behavior

Image

# Expected behavior

The event name should be normalized to the following case: `LIGO/Virgo G298048`.

We should audit all of the regular expressions and make sure that they are normalizing to a consistent case.

# Steps to reproduce

See query in screen shot.

# Environment

macOS Sequoia 15.4 / Safari 18.4

Courey:
# Description
This normalizes:
 `LIGO/Virgo G298048` `LIGO/VIRGO G298048` `LIGO/VIRGO/KAGRA G298048` `LIGO/Virgo/KAGRA G298048`
So they all have virgo with a capital V and the rest lowercase:
`LIGO/Virgo G298048` `LIGO/Virgo/KAGRA G298048`

# Related Issue(s)
Resolves #2984

# Testing
This was tested locally by running the following variations through the subject matcher and examining the results:
 `LIGO/Virgo G298048` `LIGO/VIRGO G298048` `LIGO/VIRGO/KAGRA G298048` `LIGO/Virgo/KAGRA G298048`

Test cases were also added for the following cases:
`LIGO/VIRGO G298048` 
`LIGO/VIRGO/KAGRA G298048`

### DIFF
--- a/__tests__/circulars.ts
+++ b/__tests__/circulars.ts
@@ -515,6 +515,24 @@ describe('parseEventFromSubject', () => {
       expect(parseEventFromSubject(ligoVirgo)).toBe('LIGO/Virgo S231127cg')
     })
 
+    test('handles ligo/virgo normalization', () => {
+      const ligoVirgo =
+        'ligo/virgo S231127cg: Identification of a GW compact binary merger candidate'
+      expect(parseEventFromSubject(ligoVirgo)).toBe('LIGO/Virgo S231127cg')
+    })
+
+    test('handles ligo/virgo flag normalization', () => {
+      const ligoVirgo =
+        'ligo/virgo s231127cg: Identification of a GW compact binary merger candidate'
+      expect(parseEventFromSubject(ligoVirgo)).toBe('LIGO/Virgo S231127cg')
+    })
+
+    test('handles ligo/virgo id normalization', () => {
+      const ligoVirgo =
+        'ligo/virgo S231127CG: Identification of a GW compact binary merger candidate'
+      expect(parseEventFromSubject(ligoVirgo)).toBe('LIGO/Virgo S231127cg')
+    })
+
     test('handles LIGO alert', () => {
       const ligoVirgoKagra =
         'LIGO S231127cg: Identification of a GW compact binary merger candidate'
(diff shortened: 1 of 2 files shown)


jracusin (approved):

lpsinger (changes requested):
This doesn't enforce the correct case for `LIGO`, `KAGRA`, or the event ID string.

Courey:
> This doesn't enforce the correct case for `LIGO`, `KAGRA`, or the event ID string.

Could you expand a little more? The ticket said:
> The event name should be normalized to the following case: LIGO/Virgo G298048.

If there was additional information, it should be included in the issue. I thought the only issue was that LIGO should be Ligo. There is no mention of any issue with the other two.

lpsinger:
> > This doesn't enforce the correct case for `LIGO`, `KAGRA`, or the event ID string.
> 
> Could you expand a little more? The ticket said:
> 
> > The event name should be normalized to the following case: LIGO/Virgo G298048.
> 
> If there was additional information, it should be included in the issue.

In the meeting, I said that we need to normalize (i.e., translate to a single canonical form) the entire event ID, not just the word "Virgo."

We need to audit _all_ of the subject matchers, starting with this one. I definitely don't have the bandwidth to write a formal specification. Do we need to pair you with another team member on this?

> I thought the only issue was that LIGO should be Ligo.

No, the correct cases are LIGO, Virgo, and KAGRA, respectively. The event ID, G298048, should always be all caps.

Courey:
I have made and pushed all those changes for this matcher. As far as the others go, I have no reference for what is standard and what is not. 

all seem standardized except there is no upper or lower on these:
`Baksan Neutrino Observatory Alert ${id}`
`EP${id}`
`sb${id}`

do those need to be upper or lowered?

lpsinger:
> all seem standardized except there is no upper or lower on these: `Baksan Neutrino Observatory Alert ${id}` `EP${id}` `sb${id}`
> 
> do those need to be upper or lowered?

I don't know. I would only be able to tell by looking at the archive and trying to figure out which usage is more prevalent.

lpsinger (approved):

PR #3028, opened 2025-04-11
allow non-subject matcher eventIds in circular edit form

### LINKED ISSUE #2999: remove requirement for subject matcher match in circular moderation eventId field
# Description
Currently it is required that when a moderator changes or adds an event Id on a circular, that the id must match a subject matcher. We have a few that Judy wants to manually add event ids to that do not need a subject matcher. 

The request is that the eventId can be submitted even when it is NOT a valid subject match, but a visual indication that it isn't a subject match is displayed so moderators are aware that what they are submitting is not a subject match and can check for typos if that is not the intent.

# Acceptance criteria
- eventIds may be added to circulars that do not have a valid subject matcher
- a visual indication that the eventId is not a valid subject matcher is displayed on the circular edit page when a match is not found

Courey:
# Description
Judy would like to be able to add eventIds that are not in the subject matchers. She would like a visual indication when it doesn't match a subject matcher so moderators can be aware of typos and accidental use of non-matched eventIds. A non-matched eventId should not prevent save.

# Related Issue(s)
Resolves #2999 

# Testing
This was tested locally.

Hint shows up as you are typing:
Screenshot 2025-04-11 at 10 31 40 AM
Input group is highlighted in red when the moderator clicks outside of the input:
Screenshot 2025-04-11 at 10 31 46 AM

### DIFF
--- a/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.edit.$circularId/CircularEditForm.tsx
@@ -30,6 +30,7 @@ import {
 import { RichEditor } from './RichEditor'
 import { CircularsKeywords } from '~/components/CircularsKeywords'
 import CollapsableInfo from '~/components/CollapsableInfo'
+import Hint from '~/components/Hint'
 import Spinner from '~/components/Spinner'
 import { AstroDataContext } from '~/components/circularDisplay/AstroDataContext'
 import { MarkdownBody } from '~/components/circularDisplay/Body'
@@ -148,8 +149,7 @@ export function CircularEditForm({
   const bodyValid = bodyIsValid(body)
   const dateTimeValid = circularId ? dateTimeIsValid(dateTime) : true
   const sending = Boolean(useNavigation().formData)
-  const valid =
-    subjectValid && bodyValid && dateTimeValid && submitterValid && eventIdValid
+  const valid = subjectValid && bodyValid && dateTimeValid && submitterValid
 
   let headerText, saveButtonText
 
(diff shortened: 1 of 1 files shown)


tylerbarna (commented):
Would it be useful to add a secondary confirmation when confirming an edit in this case?

Courey:
> Would it be useful to add a secondary confirmation when confirming an edit in this case?

I asked Judy about that and she said some kind of modal that requires clicking would be overkill and she preferred having a visual notification instead.

app/routes/circulars.edit.$circularId/CircularEditForm.tsx
```diff
@@ -278,21 +278,33 @@ export function CircularEditForm({
...
+                defaultValue={defaultEventId}
+                onChange={({ target: { value } }) => {
+                  setEventId(value)
+                  setEventIdValid(!value || eventIdIsValid(value))
+                }}
+              />
+            </InputGroup>
+            <Hint className="text-secondary">
+              {eventIdValid
+                ? '\u00A0'
```
lpsinger, line 304: Is the purpose of this non-breaking space to reserve space on the page for the hint, even when it is not displayed?

Courey: yes. you don't like when anything jumps on the page

lpsinger: Would you please assign this to `const nonBreakingSpace = '\u00A0'` so that its purpose is more obvious?

lpsinger (approved):

PR #3042, opened 2025-04-15
normalizing subject matchers

Courey:
# Description
Some subject matchers were missing and others were not normalized. This change brings the subject matchers in line with what is expected.

### Subject matchers added:
'GW250206'
'SN2002ap'
'XRF 050509C'
'AT2023lcr'
'IceCube-Cascade 221223A'

### Subject matchers normalized:
'EP241119a'
'FRB 20250206A'
# Related Issue(s)
Resolves #2997

# Testing
Included unit tests to cover all changes

PR #3047, opened 2025-04-16
correct ligo and ep inconsistent names DO NOT MERGE!

### LINKED ISSUE #2989: Repair instances of all caps VIRGO in eventId in the database
There are instances of event ids where `LIGO/VIRGO` or `LIGO/VIRGO/KAGRA` are used. These should be `LIGO/Virgo` or `LIGO/Virgo/KAGRA` instead.

Repair the event id's on the circulars that have these and repair synonym records where necessary (ensuring the capitalization in the synonym eventId is consistent with the updated circular).

In some instances, where there are two of the same event name with the inconsistent capitalization, the removal of one of the synonyms will be necessary. In those cases, the initialDate should be recalculated so it is accurate given the new data.

Courey:
# Description
After https://github.com/nasa-gcn/gcn.nasa.gov/pull/3042 goes in, this branch needs to be rebased.

Repairing data for EP and LVK circulars after normalization of the subject matchers. 

# Related Issue(s)
This depends on https://github.com/nasa-gcn/gcn.nasa.gov/pull/3042 

Resolves #2989


lpsinger (commented):
After we completed the last database migration, I raised a concern that manual database migrations carry risks of data loss and cause a lot of code churn. I asked that we do some research on automated migration tools for DynamoDB. For relational databases, there are tools like [Alembic](https://alembic.sqlalchemy.org/en/latest/) in Python which encourage you to keep migration scripts under version control and make it possible to test and roll back migrations as many times as needed. I don't know whether or not there are similar tools for DynamoDB or NoSQL databases more generally, but I think we should do some research on that before we do another migration.

Courey:
> After we completed the last database migration, I raised a concern that manual database migrations carry risks of data loss and cause a lot of code churn. I asked that we do some research on automated migration tools for DynamoDB. For relational databases, there are tools like [Alembic](https://alembic.sqlalchemy.org/en/latest/) in Python which encourage you to keep migration scripts under version control and make it possible to test and roll back migrations as many times as needed. I don't know whether or not there are similar tools for DynamoDB or NoSQL databases more generally, but I think we should do some research on that before we do another migration.

There are 48 LVK records to repair and 76 EP repairs, so no great number. The downside to a document store that is a schema-less database is that schema migrations are not going to come naturally. 

Most packages that try to do this are many years unmaintained and have few weekly downloads. I wouldn't trust any of those. 

The best option I've found is [onetable-migrate](https://www.npmjs.com/package/onetable-migrate). it was last updated 10 months ago, and before that was around a 2 year gap. it has 1,468 weekly downloads.
It is dependent on [dynamodb-onetable](https://www.npmjs.com/package/dynamodb-onetable).

Since we already use ARC to manage our dynamo tables, I'm not sure how well `dynamodb-onetable` would play with that.

onetable works best with one table architecture. We have multiple tables. I'm not sure the risk there, I'd have to look into it more. 
We would continue to define tables and indexes in app.arc. 
We would have to write and maintain a OneTable schema that matches any changes to dynamo. There would be one for each table. We would have to remember to always update OneTable when we make any changes (indexes, fields, etc) to any of the tables managed by architect.
We would have to create a OneTable instance for each table.

This would allow us to run migrations, but not necessarily rollback migrations. If we change data like would happen in this (or any) backfill, we could not roll it back. dynamodb does not keep track of what changes were made to which records. So it would be a one way migration, so it would have no benefit of rollback.

Courey:
Why don't we just create a temporary table and instead of updating the circulars, put all the updates into the temporary table. Then they can be reviewed. once everyone is happy with the state of them, they could then be used to update the circular with the same circularId in the circulars table?

lpsinger:
The tables and relationships involved here are quite complex. I am not comfortable making direct changes to the data in the tables. The backend that you have developed has methods to make changes to event IDs and their synonyms in a way that maintains the integrity of the data. I would like you to make the event ID updates using those library functions.

I suggest that as a first step you expose them in a REST API. We already have a REST API for creating GCN Circulars. It may not be much work to add a REST API for submitting updates.

PR #3064, opened 2025-04-24
add appropriate eventId on update, correction, and creation

### LINKED ISSUE #3063: update eventId automatically on correction or edit
# Description
make it so that the eventId is updated when:
an eventId is submitted on create, update, or correction

# Acceptance criteria
- the eventId updates when the subject is updated.
- the eventId can be updated to be something that is not a valid subject matcher
- the eventId can be updated or corrected directly in the edit circular form.

Courey:
# Description
This update makes it so that the eventId is included automatically on update, create, and correction even if not included. The different options for the correction are complex, so I created an additional function that determines the desired eventId depending on input. 

That priority is:
- if the circular id and the eventId parsed from the subject AND the correction eventId is not the same as the circular's eventId, the correction eventId is the desired.
- if there is no eventId on the original circular and the eventId being passed in is different than what it would be if it was parsed from the subject, the eventId being passed in is the desired one. This allows moderators to put an eventId on the circular that doesn't come from the subject being parsed.
-   if there is not an eventId on the original circular and there is not an eventId being passed in, it will attempt to parse the subject for an eventId and will use that.
- if the original circular's eventId is the same as the eventId on the correction request (which will happen if there was an eventId originally and no change was entered), then we see if the parsed subject is different. If so, we update the eventId based on the parsed value. This is so that when there is a correction submitted that changes the subject to a valid subject matcher match, the eventId will automatically be added.
- and finally, as a default case, we side with the submitted correction's eventId

# Related Issue(s)
Resolves #3063 

# Testing
This was tested locally.
I also added unit tests for the cases I could think of.

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -105,7 +105,7 @@ export async function action({ request }: ActionFunctionArgs) {
   const eventId = getFormDataString(data, 'eventId') || undefined
   const intent = getFormDataString(data, 'intent')
   const format = getFormDataString(data, 'format') as CircularFormat | undefined
-  const eventId = getFormDataString(data, 'eventId') || undefined
+
   if (format && !circularFormats.includes(format)) {
     throw new Response('Invalid format', { status: 400 })
   }


lpsinger:
What is the relation of this PR to #3090?

Courey:
closing in favor of https://github.com/nasa-gcn/gcn.nasa.gov/pull/3198

PR #3078, opened 2025-05-01
Circulars: Search by author with clickable name

### LINKED ISSUE #3111: Circulars: Make Author names clickable
# Description
Add the ability to click on an author's name in a circular to direct to an advanced search for that author

# Acceptance criteria
1. Visible clickability for circular authors
2. directs to a search that returns all/almost all circulars by that author

tylerbarna:
# Description
This makes the submitter of a circular a hyperlink that directs to a search for all Circulars submitted by that user with their current affiliation.

# Related Issue(s)
resolves #3111 

# Testing
1. navigate to the circular archive
2. click on a circular
3. click on the name of the submitter
4. it should direct to a search of the archive for that user's circulars


app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -68,7 +68,11 @@ export function FrontMatter({
           </small>
         </FrontMatterItem>
       )}
-      <FrontMatterItem label="From">{submitter}</FrontMatterItem>
+      <FrontMatterItem label="From">
+        <Link href={`/circulars?query=submitter:"${submitter}"`}>
```
lpsinger, line 72: - Is there a function that you can call to correctly escape the submitter for Lucene?
- Please use the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) and [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to construct the URL to ensure that query string arguments are correctly encoded.

lpsinger (changes requested):
Please post a screenshot.

tylerbarna:
here is how the link looks
Screenshot 2025-05-19 at 14 45 42
When clicked, it directs to the archive with a search executed for the author 
Screenshot 2025-05-19 at 14 46 28

tylerbarna:
I also addressed a point Judy made about this feature; the search will execute with just the submitter's name rather than their name and association. This will account for cases where an author changes affiliations over the course of their work

app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -69,9 +77,7 @@ export function FrontMatter({
         </FrontMatterItem>
       )}
       <FrontMatterItem label="From">
-        <Link href={`/circulars?query=submitter:"${submitter}"`}>
-          {submitter}
-        </Link>
+        <Link href={submitterUrl.toString()}>{submitter}</Link>
```
lpsinger, line 80:
```suggestion
        <Link href={submitterUrl}>{submitter}</Link>
```

app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -51,6 +51,14 @@ export function FrontMatter({
...
   | 'editedOn'
 >) {
+  const authorName = submitter.includes(' at ')
+    ? submitter.split(' at ')[0]
+    : submitter
+  const params = new URLSearchParams({ query: `submitter: "${authorName}"` })
+  const submitterUrl = new URL(
+    `/circulars?${params.toString()}`,
+    window.location.origin
+  )
```
lpsinger, line 61:
```suggestion
  const submitterUrl = `/circulars?${params}`
```

app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -51,6 +51,14 @@ export function FrontMatter({
   | 'editedBy'
   | 'editedOn'
 >) {
+  const authorName = submitter.includes(' at ')
+    ? submitter.split(' at ')[0]
+    : submitter
+  const params = new URLSearchParams({ query: `submitter: "${authorName}"` })
```
lpsinger, line 57: How do we escape the author name for usage in Lucene syntax (e.g., escaping quotes)? There may be a third-party package to do that.

app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -51,6 +51,14 @@ export function FrontMatter({
   | 'editedBy'
   | 'editedOn'
 >) {
+  const authorName = submitter.includes(' at ')
+    ? submitter.split(' at ')[0]
+    : submitter
```
lpsinger, line 56: - Technically the author name could contain the string `' at '`. Please make sure that this code is correct for that case.
- Construct a query that searchers for submitter by their name OR by their email address.

lpsinger (changes requested):

app/components/circularDisplay/FrontMatter.tsx
tylerbarna: I ran a script checking the current collection of circulars in the archive for circulars that have more than one instance of ` ' at '` present in the submitter field; there are 58 instances where there is more than one ` ' at '` string. All of these are cases where there are institutions with "at" in their name (eg U. of Texas at Austin, ANU at CIT, SSL at UCB). The current method won't have any issue with isolating the submitter name since it's looking for the first chunk of text that precedes the first ` ' at '`.

tylerbarna: [601c7e9](https://github.com/nasa-gcn/gcn.nasa.gov/pull/3078/commits/601c7e9f919179dadd73510260664e36a550c1a4) adds the email to the search as well

lpsinger (changes requested):

app/components/circularDisplay/FrontMatter.tsx
```diff
@@ -51,6 +51,20 @@ export function FrontMatter({
   | 'editedBy'
   | 'editedOn'
 >) {
+  const escapeQuery = (str: string) =>
+    str.replace(/[+\-=&|><!(){}[\]^"~*?:/\\]/g, '\\$&')
```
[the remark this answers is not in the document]
tylerbarna: I did not realize these were already dependencies (`lucene` is used in the tooltip component). Just switched methods to use that. The only issue I'm noticing with it is that it escapes spaces, which is not really necessary with how our query string search works

lpsinger (commented):
This weird string formatting of the author field is a legacy of GCN Classic. Can we instead store these as separate fields so that we don't have to parse them apart?

tylerbarna:
Just to log the relevant information:

It should be possible to create a backfill that would split the existing submitter into 3 separate fields
- `Submitter`
- `Institution` (or something along those lines)
- `Email`

This would be fairly straightforward, as all submitter strings follow the format `[SUBMITTER] at [INSTITUTION] <EMAIL>`. There a small number of exceptions to this where some historical circulars either do not have an institution, email, or both.

However, because of the standardized format, institution and email can be made as optional fields that are generally populated, where `Submitter` is either the text prior to the first instance of ' <' if there are no instances of ' at ' in the string _or_ the portion of the string prior to the first instance of ' at ', `Institution` is either all text that follows the first instance of ' at ' if there is no email present _or_ all text prior to ' <' if there is an email present, and `Email` is all text enclosed inside '< >'. 

As mentioned in an earlier comment, this approach is very unlikely to run into any issues with submitter names containing ' at ' specifically, it won't be an issue if the institute contains ' at ' in the name, and it handles any cases where one or both of the optional fields are not present.

We can retain the current appearance by just populating the submitter line with these three fields separated by the existing `[SUBMITTER] at [INSTITUTION] <EMAIL>` structure

PR #3090, opened 2025-05-05
Maintain synonym state on version update

### LINKED ISSUE #3091: handle synonym management in circular version updates
Judy noticed that when a circular change request was approved with an eventId change in it, the change was successful but the synonym group was not created. This is because the put used in version control is different than the circular put so it bypasses the existing synonym management. Add handling to manage synonym state when changes are requested.

Courey:
we noticed that approval requests do not use the circular put. They would bypass the existing code to manage synonyms.
On edit or request change, synonym state should be managed appropriately when circular eventId's are changed:
- if no more circulars with the old eventId remain, the synonym's date is changed to the unix epoch. The synonym will exist but it's timestamp is set to the unix epoch
- if the eventId doesn't change, it shouldn't do anything
- if the eventId is a never used before eventId, it should create a new synonym
- if the eventId is one that already has a synonym, it doesn't initialize a synonym as one already exists
- if the eventId does not exist on the old or new version, it should skip synonym management.

# Issue
resolves #3091 

# Testing
I tested this locally by commenting out the zendesk and submitter checks in the app/routes/circulars._archive._index/route.tsx file. I added seed data for each different case.

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -314,7 +314,7 @@ export async function putSynonyms({
 export async function getOldestDate(eventId: string) {
   const circulars = await getSynonymMembers(eventId)
   const oldest = min(circulars.map(({ createdOn }) => createdOn))
-  return oldest ? oldest : -1
+  return oldest ?? -1
 }
 
 /*
--- a/app/table-streams/synonyms/index.ts
+++ b/app/table-streams/synonyms/index.ts
@@ -52,7 +52,7 @@ export const handler = createTriggerHandler(
             const oldestDate = min(
               synonyms.map(({ initialDate }) => initialDate)
             )!
-            const initialDate = oldestDate ? oldestDate : -1
+            const initialDate = oldestDate ?? -1
 
             await putIndex({
               synonymId,
--- a/seed.json
+++ b/seed.json
@@ -5680,6 +5680,12 @@
       "eventId": "GRB 230824R",
       "slug": "grb-230824r",
       "initialDate": 1693037110788
+    },
+    {
+      "synonymId": "d0ef15f2-a346-4b2c-ac9d-f7287194444",
+      "eventId": "GRB 230824X",
+      "slug": "grb-230824x",
+      "initialDate": 1693037110782
     }
   ],
   "auto_increment_metadata": [


lpsinger (changes requested):

app/routes/circulars/circulars.server.ts
```diff
@@ -379,9 +383,30 @@ export async function putVersion(
     editedBy: formatAuthor(user),
     editedOn: Date.now(),
   }
+
   validateCircular(newCircularVersion)
 
-  return await circularVersionsAutoIncrement.put(newCircularVersion)
+  const promises = [
+    circularVersionsAutoIncrement.put(newCircularVersion),
+  ] as any[]
```
lpsinger, line 391: Please try to avoid using type assertions, because they can mask bugs that type inference and type checking would reveal.

app/routes/circulars/circulars.server.ts
```diff
@@ -379,9 +383,30 @@ export async function putVersion(
...
+        newCircularVersion.createdOn
+      )
+    )
+  }
+
+  const results = await Promise.all(promises)
+
+  if (!newCircularVersion.eventId && oldCircular.eventId) {
+    await deleteIfGroupIsEmpty(oldCircular.eventId)
+  }
```
lpsinger, line 407: Is this necessary, given that `manageSynonymVersionUpdates` itself calls `deleteIfGroupIsEmpty`?

app/routes/synonyms/synonyms.server.ts
```diff
@@ -166,6 +166,55 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
...
+
+export async function manageSynonymVersionUpdates(
+  newEventId: string,
+  oldEventId?: string,
+  createdOn?: number
+) {
+  if (newEventId === oldEventId) return
+
+  const oldestDate = await getOldestDate(newEventId)
+  const dateParam = oldestDate ? oldestDate : createdOn
```
lpsinger, line 186:
```suggestion
  const dateParam = oldestDate || createdOn
```

app/routes/synonyms/synonyms.server.ts
```diff
@@ -166,6 +166,55 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
...
+  if (!dateParam) throw Error
+
+  const promises = []
+  if (newEventId) {
+    promises.push(tryInitSynonym(newEventId, dateParam))
+  }
+  if (oldEventId && oldEventId != newEventId) {
+    promises.push(deleteIfGroupIsEmpty(oldEventId))
+    promises.push(updateInitialDate(oldEventId))
```
lpsinger, line 195: You're doing more database work than you need to here: if the group is empty, then you need to delete it; otherwise, you need to update the date.

Also, where do you update the date on the group containing the _new_ event ID?

lpsinger (changes requested):

lpsinger:
> I mentioned that #3198 has a dependency on this branch. This branch IS 3090. So I guess it does have a dependency on itself as well.

Oops. I posted my question on the wrong PR.

Courey:
> @Courey: This morning you raised awareness that if we don't delete unused event IDs from the synonyms table, that the Circulars event view will contain event IDs that contain no Circulars.
> 
> What if we keep the "earliest date" up to date when we add or remove Circulars from a group, and set it to undefined when the group becomes empty? Can we configure our OpenSearch query to filter out groups that have a field that is equal to a certain value (in this case, undefined)?
> 
> Would that be performant? Or would it be too slow, because it would have to scan all of the results?

Without a date, the sort would break here:
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/synonyms/synonyms.server.ts#L85
and here:
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/table-streams/synonyms/index.ts#L52

it was designed to be a required field. What I have it as currently is if there isn't a circular to get a date from, it gets the unix epoch. So we could just exclude things with that datetime

Courey:
> > Without a date, the sort would break here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/routes/synonyms/synonyms.server.ts#L85
> 
> Does OpenSearch allow numeric fields to be undefined (or null)? If not, what is the failure mode?
> 
> If it does not support undefined or null, does it support NaN?
> 
> > and here: https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/table-streams/synonyms/index.ts#L52
> 
> That one is OK... we could supply any fallback value there that we like.

When you make it null, you get this error:
```
software.amazon.dynamodb.services.exceptions.DynamoDBLocalServiceException: ExpressionAttributeValues must not be empty

[
  ValidationException: ExpressionAttributeValues must not be empty
      at throwDefaultError (/Users/courey/dev/gcn.nasa.gov/node_modules/@smithy/smithy-client/dist-cjs/index.js:836:20)
      at /Users/courey/dev/gcn.nasa.gov/node_modules/@smithy/smithy-client/dist-cjs/index.js:845:5
      at de_CommandError (/Users/courey/dev/gcn.nasa.gov/node_modules/@aws-sdk/client-dynamodb/dist-cjs/index.js:2230:14)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async /Users/courey/dev/gcn.nasa.gov/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20
      at async /Users/courey/dev/gcn.nasa.gov/node_modules/@aws-sdk/lib-dynamodb/dist-cjs/index.js:164:30
      at async /Users/courey/dev/gcn.nasa.gov/node_modules/@smithy/core/dist-cjs/index.js:168:18
      at async /Users/courey/dev/gcn.nasa.gov/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38
      at async /Users/courey/dev/gcn.nasa.gov/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:34:22 {
    '$fault': 'client',
    '$metadata': {
      httpStatusCode: 400,
      requestId: '1f79e40b-dcb9-4154-8722-374a7f0fe545',
      extendedRequestId: undefined,
      cfId: undefined,
      attempts: 1,
      totalRetryDelay: 0
    },
    __type: 'com.amazon.coral.validate#ValidationException'
  }
]
```

app/routes/synonyms/synonyms.server.ts
```diff
@@ -166,11 +166,59 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
...
+ * @param createdOn - the Circular createdOn timestamp
+ * @param newEventId - the updated eventId
+ * @param oldEventId - the original eventId
+ */
+export async function manageSynonymOnVersionUpdates(
+  createdOn: number,
+  newEventId?: string,
+  oldEventId?: string
+) {
+  if (newEventId === oldEventId) return
```
lpsinger: What if the user changed the creation date but not the event ID?

jracusin: Why are we allowing users to modify the creation date?

lpsinger: We probably originally added that feature to fix up the dates of those first 100 or so Circulars ever that didn't have meaningful dates.

For this PR, It does not matter why users can modify dates. They can, and so this PR shouldn't break it. It's a reasonable question if we should continue supporting it, but it should be discussed in a separate issue.

jracusin: Added issue #3259

Courey:
wait on this for a minute. I'm reviewing errors I see in my terminal (they seem to happen on table-streams, which I didn't change in this branch). I want to make sure they have nothing to do with the branch.

Courey:
> > wait on this for a minute. I'm reviewing errors I see in my terminal (they seem to happen on table-streams, which I didn't change in this branch). I want to make sure they have nothing to do with the branch.
> 
> OK, ping me again when you are ready.

ok @lpsinger we are good to go. the issue was that I hadn't changed the table-stream to account for the -1 timestamp as well. [This was the fix](https://github.com/nasa-gcn/gcn.nasa.gov/pull/3090/files#diff-5786af0b72c8fd6515a8388a81ab734ff1bcf22af72dc7cbf91b5aebddf09d78R55)

app/table-streams/synonyms/index.ts
```diff
@@ -52,12 +52,13 @@ export const handler = createTriggerHandler(
             const oldestDate = min(
               synonyms.map(({ initialDate }) => initialDate)
             )!
+            const initialDate = oldestDate ? oldestDate : -1
```
lpsinger: If the list is empty, then `min()` will return undefined, so you can use the nullish coalescing operator here. This will eliminate the corner case where the date is valid but 0.

```suggestion
            const initialDate = oldestDate ?? -1
```

app/routes/synonyms/synonyms.server.ts
```diff
@@ -313,8 +313,8 @@ export async function putSynonyms({
 
 export async function getOldestDate(eventId: string) {
   const circulars = await getSynonymMembers(eventId)
-
-  return min(circulars.map(({ createdOn }) => createdOn)) || -1
+  const oldest = min(circulars.map(({ createdOn }) => createdOn))
+  return oldest ? oldest : -1
```
lpsinger:
```suggestion
  return oldest ?? -1
```

lpsinger (changes requested):

PR #3094, opened 2025-05-06
lazy loading circulars for each synonym group

### LINKED ISSUE #3093: lazy load circulars for synonyms on group view
# Description
Load times for the group view on the circulars route was too slow because it was loading 100 synonyms (default) PLUS all the circulars that belong to each of those 100 synonyms, so it was a large amount of data being pulled on page load. 

Lazy load the circular data for each synonym so that on page load only 100 items (default) are pulled at once. Additional items will be retrieved when the synonym group drop down arrow is clicked.

# Acceptance criteria
- synonyms are loaded on page load
- circulars for each synonym are lazy loaded

Courey:
# Description
Converts the group view on the circulars route to use lazy loading for the circulars belonging to each synonym group.

# Related Issue(s)
Resolves #3093 

# Testing
This has been tested locally.


app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,70 @@
...
+              <Spinner />
+            </span>
+          )}
+          {items.map(({ circularId, subject }) => {
+            return (
+              <li key={circularId} value={circularId}>
+                <Link to={`/circulars/${circularId}`}>{subject}</Link>
+              </li>
+            )
+          })}
```
lpsinger, line 58:
```suggestion
          {items.map(({ circularId, subject }) => (
              <li key={circularId} value={circularId}>
                <Link to={`/circulars/${circularId}`}>{subject}</Link>
              </li>
            )
          )}
```

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,70 @@
...
+
+  useEffect(() => {
+    setItems(fetcher.data ?? [])
+  }, [fetcher.data])
+
+  return (
+    <div>
+      <details
+        onClick={() => {
+          fetcher.submit(data, { method: 'POST', action: '/synonyms' })
```
lpsinger, line 38: Please use `GET` rather than `POST`, since this request does not mutate any data.

That might mean that you have to hang the loader off of a different or a new route.

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,70 @@
...
+import { Link, useFetcher } from '@remix-run/react'
+import type { action } from 'app/routes/synonyms/route'
+import { useEffect, useState } from 'react'
 
-import type { SynonymGroupWithMembers } from '../synonyms/synonyms.lib'
+import type { Circular } from '../circulars/circulars.lib'
+import type { SynonymGroup } from '../synonyms/synonyms.lib'
+import Spinner from '~/components/Spinner'
+
+function SynonymList({
```
lpsinger, line 16: Please propose a more descriptive name for this component. This is a list of Circulars belonging to an event ID, not a list of synonyms.

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,70 @@
...
+  searchString,
+}: {
+  eventIds: string[]
+  slugs: string[]
+  searchString: string
+}) {
+  const [items, setItems] = useState<Circular[]>([])
+  const fetcher = useFetcher<typeof action>()
+  const data = new FormData()
+  data.set('eventIds', eventIds.join(','))
```
lpsinger, line 28: I believe that you are allowed to have multiple form inputs with the same name. You don't need to concatenate them into a string (which would introduce a new requirement that the event names could not contain the delimiter).

app/routes/circulars._archive._index/route.tsx
```diff
@@ -373,7 +373,7 @@ export default function () {
         <>
           {isGroupView ? (
             <SynonymGroupIndex
-              allItems={items as SynonymGroupWithMembers[]}
+              allItems={items as SynonymGroup[]}
```
lpsinger, line 376: Why is a type assertion necessary here?

lpsinger (changes requested):

app/routes/circulars._archive._index/route.tsx
Courey: Screenshot 2025-05-09 at 12 55 59 PM

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
Courey: It's not. It's circulars belonging to an array of different eventIds which together are a synonym. I'm changing it to CircularsBelongingToASynonymGroup

lpsinger: How about CircularsForSynonymGroup?

app/routes/circulars._archive._index/route.tsx
lpsinger: I see.

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
Courey: I made it a post because we are posting data. You can't post data with a GET. It would then have to be in a query string, which doesn't support arrays. GETs don't have a body. So the options are to send it as a query string or to use POST and post the array of eventIds.

lpsinger: > I made it a post because we are posting data.

Yes, but _why_ are you posting the data?

> It would then have to be in a query string, which doesn't support arrays.

Both POST with URL form encoding and GET with query strings support multiple form key-value pairs with the same keys but different values. (After all, they are the same encoding!)

```
> let params = new URLSearchParams()
undefined
> params.append('foo', 'bar')
undefined
> params.append('foo', 'bat')
undefined
> params.toString()
'foo=bar&foo=bat'
```

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,70 @@
...
+
+  useEffect(() => {
+    setItems(fetcher.data ?? [])
+  }, [fetcher.data])
```
lpsinger, line 32: Please employ useFetcher instead of useEffect+useState. useFetcher is designed for non-navigation submission of forms.

lpsinger (changes requested):

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
lpsinger, line 32: Why do you need the state and effect hooks? Why can't you just use `fetcher.data` in your component code below?

lpsinger (changes requested):

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,64 @@
...
+  slugs,
+  searchString,
+}: {
+  eventIds: string[]
+  slugs: string[]
+  searchString: string
+}) {
+  const fetcher = useFetcher<typeof loader>()
+  const params = new URLSearchParams()
+  eventIds.forEach((eventId) => params.append('eventIds', eventId))
```
lpsinger, line 25:
```suggestion
  const params = new URLSearchParams(eventIds.map((eventId) => ['eventIds', eventId])
```

app/routes/circulars._archive._index/SynonymGroupIndex.tsx
```diff
@@ -5,17 +5,64 @@
...
+        <ol className="margin-left-3">
+          {fetcher.state === 'loading' &&
+            fetcher.data &&
+            fetcher.data.length === 0 && (
+              <span className="text-middle">
+                <Spinner />
+              </span>
+            )}
+          {fetcher.data &&
+            fetcher.data.map(({ circularId, subject }) => (
```
lpsinger, line 48: Does this work?

```suggestion
          {fetcher.data?.map(({ circularId, subject }) => (
```

app/routes/synonym.circulars.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { getAllSynonymMembers } from './synonyms/synonyms.server'
+
+export async function loader({
+  request: { headers, url },
+}: LoaderFunctionArgs) {
+  const isBrowserNavigation = headers.get('Sec-Fetch-Mode') !== 'cors'
+
+  if (isBrowserNavigation) {
+    throw new Response(null, { status: 404 })
+  }
```
lpsinger, line 12: What does this do? Please add comments.

app/routes/synonyms/route.tsx
```diff
@@ -27,6 +29,13 @@ export async function loader({ request }: LoaderFunctionArgs) {
   return null
 }
 
+export async function action({ request }: ActionFunctionArgs) {
+  const data = await request.formData()
+  const eventIds = getFormDataString(data, 'eventIds')?.split(',')
+
+  return eventIds?.length ? await getAllSynonymMembers(eventIds) : []
+}
+
```
lpsinger, line 38: Do you still need this action function, now that you are no longer making a POST request?

lpsinger (changes requested):

Courey:
You can still test the api route with
`curl "localhost:3333/api/synonym/circulars?eventIds=GRB%20230812B"` even if it's not available via the browser.

app/routes/api.synonym.circulars.ts
```diff
@@ -0,0 +1,21 @@
+import type { LoaderFunctionArgs } from '@remix-run/node'
```
lpsinger, line 1: Please add license header.

lpsinger (changes requested):

lpsinger (approved):

PR #3105, opened 2025-05-16
update circular api endpoint creates new version

### LINKED ISSUE #3060: create API endpoint for updating circulars
# Description
Add an API endpoint accessible by moderators only allowing an update to circulars.

# Acceptance criteria
An API endpoint exists that takes circular data and updates the circular.

Courey:
# Description
This adds a PUT path to the circulars api endpoint. 
It requires moderator permissions.
The subject, body, and/or eventId can be edited.
It creates a new version of the circular.

# Related Issue(s)
Resolves #3060 

# Testing
I tested locally by commenting out the cognito portion of the api endpoint and putting in a hard coded user.
I tested:
- being able to edit with an existing eventId to change it.
- editing a circular without an eventId to add one
- editing a circular without an eventId to change the subject and body without any addition of an eventId.
- POSTing a new circular without an eventId
- POSTing a new circular with an eventId

### DIFF
--- a/app/routes/api.circulars.ts
+++ b/app/routes/api.circulars.ts
@@ -196,7 +196,7 @@ export async function action({ request }: ActionFunctionArgs) {
       break
 
     default:
-      throw new Response(null, { status: 405 })
+      throw new Error('this code should not be reached')
   }
 
   const [{ existingIdp, ...attrs }, groups] = await Promise.all([
--- a/app/routes/api.circulars.ts
+++ b/app/routes/api.circulars.ts
@@ -172,7 +172,7 @@ async function getUserAttributes(Username: string) {
  */
 export async function action({ request }: ActionFunctionArgs) {
   if (!['PUT', 'POST'].includes(request.method))
-    throw new Response(null, { status: 405 })
+    throw new Response('unsupported method', { status: 405 })
 
   const bearer = getBearer(request)
   if (!bearer) throw new Response('Bearer missing', { status: 403 })
--- a/app/routes/api.circulars.ts
+++ b/app/routes/api.circulars.ts
@@ -172,7 +172,7 @@ async function getUserAttributes(Username: string) {
  */
 export async function action({ request }: ActionFunctionArgs) {
   if (!['PUT', 'POST'].includes(request.method))
-    throw new Response('unsupported method', { status: 405 })
+    throw new Response(null, { status: 405 })
 
   const bearer = getBearer(request)
   if (!bearer) throw new Response('Bearer missing', { status: 403 })
@@ -242,6 +242,6 @@ export async function action({ request }: ActionFunctionArgs) {
       return await putVersion(circularVersion, user)
 
     default:
-      throw new Response('unknown intent', { status: 400 })
+      throw new Error('this code should not be reached')
   }
 }
--- a/app/routes/api.circulars.ts
+++ b/app/routes/api.circulars.ts
@@ -15,7 +15,13 @@ import invariant from 'tiny-invariant'
 
 import { getOpenIDClient } from './_auth/auth.server'
 import { type User, parseGroups, parseIdp } from './_auth/user.server'
-import { put, submitterGroup } from './circulars/circulars.server'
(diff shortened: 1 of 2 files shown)


app/routes/api.circulars.ts
```diff
@@ -166,8 +183,15 @@ export async function action({ request }: ActionFunctionArgs) {
...
-    throw new Response('Invalid scope', { status: 403 })
+  switch (request.method) {
+    case 'POST':
+      if (!scope.split(' ').includes(submitterGroup))
+        throw new Response('Invalid scope', { status: 403 })
+
+    case 'PUT':
+      if (!scope.split(' ').includes(moderatorGroup))
+        throw new Response('Invalid scope', { status: 403 })
+  }
```
lpsinger, line 194: Add a default case here that throws an error.

app/routes/api.circulars.ts
```diff
@@ -192,12 +216,26 @@ export async function action({ request }: ActionFunctionArgs) {
...
+
+      return await put(circular, user)
+
+    case 'PUT':
+      const circularVersion = {
+        circularId,
+        subject,
+        body,
+        eventId,
+      } as Circular
```
lpsinger, line 237: Remove the type assertion. You may need to add some runtime checks to clear type checking, but that's a good thing, because it will force you to do the necessary validation.

lpsinger (changes requested):

app/routes/api.circulars.ts
```diff
@@ -171,7 +172,7 @@ async function getUserAttributes(Username: string) {
  */
 export async function action({ request }: ActionFunctionArgs) {
   if (!['PUT', 'POST'].includes(request.method))
-    throw new Response(null, { status: 405 })
+    throw new Response('unsupported method', { status: 405 })
```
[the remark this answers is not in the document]
Courey: yeah, just when you use something like curl where you aren't using a `-i` flag, you don't get any feedback about what's happening. Just trying to follow the campfire rule since I was here and changing related things when I noticed.

app/routes/api.circulars.ts
```diff
@@ -166,8 +184,15 @@ export async function action({ request }: ActionFunctionArgs) {
...
-    throw new Response('Invalid scope', { status: 403 })
+  switch (request.method) {
+    case 'POST':
+      if (!scope.split(' ').includes(submitterGroup))
+        throw new Response('Invalid scope', { status: 403 })
+
+    case 'PUT':
+      if (!scope.split(' ').includes(moderatorGroup))
+        throw new Response('Invalid scope', { status: 403 })
+  }
```
lpsinger, line 195: Add a default clause.

lpsinger (changes requested):

app/routes/api.circulars.ts
Courey: so we have two options here. Currently it does have a check (it's not a default case, but a check that the method is either put or post) on line 174 https://github.com/nasa-gcn/gcn.nasa.gov/pull/3105/files#diff-62f353c505e0f6ee6f4790fa20b2ccd3d9dce7ae11b05797df5b061ddf8a2d47R174
so a default case would never be hit.

if you want a default method, that's fine, i can remove line 174 and then put in a default case. I would have to break up the rest of the code because it would become unreachable if that section has a default case, so I would just have to rearrange things so that each case block calls the function to do the updating. 

would you like to keep line 174 checking for that, or do you want me to rearrange the code and add a default case?

app/routes/api.circulars.ts
```diff
@@ -166,8 +184,20 @@ export async function action({ request }: ActionFunctionArgs) {
...
+        throw new Response('Invalid scope', { status: 403 })
+      break
+
+    case 'PUT':
+      if (!scope.split(' ').includes(moderatorGroup))
+        throw new Response('Invalid scope', { status: 403 })
+      break
+
+    default:
+      throw new Response(null, { status: 405 })
```
lpsinger:
```suggestion
      throw new Error('this code should not be reached')
```

PR #3110, opened 2025-05-19
Truncate emails if they are over 262144 bytes

### LINKED ISSUE #2899: Check Circulars message size before accepting submission
# Current behavior

There is an effective limit on the size of GCN Circulars due to the [limit of 262144  bytes on the Amazon SES API's `TemplateData` argument](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_Template.html). Circulars that would result in `TemplateData` arguments larger than this value are accepted, but they are never successfully distributed as emails. Also, they are repeatedly sent as emails because the table stream Lambda can never succeed and is retried indefinitely.

# Expected behavior

- Circulars form should prevent user from submitting form if size is exceeded
- Action function should reject Circulars larger than this size

# Steps to reproduce

See https://gcn.nasa.gov/circulars/39259. Originally reported as #2872.

# Environment

N/A

Courey:
# Description
This adds a check for byte size of the circular email body. It ensures it's at or below 262144 bytes and truncates the email if it is not. If the email is truncated, a message is added to the email body noting that it has been truncated.

# Related Issue(s)
Resolves #2899 

# Testing
This has been tested locally by form submission. I directly called the function on circular creation and examined the resulting byte size.


app/lib/email.server.ts
```diff
@@ -89,6 +91,40 @@ function maybeThrowSES(e: any, warning: string) {
...
+  circularId?: number
+  origin?: string
+}) {
+  const textBody = getBody(body)
+  const inputBytes = Buffer.byteLength(textBody, 'utf-8')
+
+  if (inputBytes <= 262144) return textBody
+
+  const buffer = Buffer.from(textBody, 'utf-8')
+  const truncatedBuffer = buffer.subarray(0, 261900)
```
Courey, line 116: I truncated it generously here to allow a fair bit of margin. We can adjust this. I thought it wouldn't matter if it was truncated to be exactly 262144 bytes at the end. If we do want it to be at the exact maximum, I would get the byte size of the truncation warning message, then I would subtract that from 262144 and make the subarray that size, but I figured that just giving it a wide margin and not bothering with that complexity was likely fine.

app/lib/email.server.ts
```diff
@@ -89,6 +91,40 @@ function maybeThrowSES(e: any, warning: string) {
...
+ * Amazon SES has a byte limit of 262144 for TemplateData arguments.
+ * If the circular body is over this size limit, the message will be truncated.
+ * If the body is truncated, the circular website link will be removed.
+ * A warning that the email has been truncated is added and the circular website link is added back in.
+ * An additional buffer is allowed for for variation in url length and truncation warning.
+ */
+export function truncateEmailBodyIfNecessary({
+  body,
+  circularId,
+  origin,
```
lpsinger, line 104: This shouldn't be an argument. Import it from `env.server.ts`.

app/lib/email.server.ts
```diff
@@ -89,6 +91,40 @@ function maybeThrowSES(e: any, warning: string) {
...
+  origin?: string
+}) {
+  const textBody = getBody(body)
+  const inputBytes = Buffer.byteLength(textBody, 'utf-8')
+
+  if (inputBytes <= 262144) return textBody
+
+  const buffer = Buffer.from(textBody, 'utf-8')
+  const truncatedBuffer = buffer.subarray(0, 261900)
+  const truncatedBody = truncatedBuffer.toString('utf-8')
```
lpsinger, line 117: This code encodes the text to bytes as UTF-8, truncates, then decodes as UTF-8. This is unsafe because UTF-8 is a variable length encoding. I suggest searching for a JavaScript library to summarize or truncate text that safely handles variable-length encodings.

lpsinger (changes requested):

app/lib/email.server.ts
```diff
@@ -89,6 +91,35 @@ function maybeThrowSES(e: any, warning: string) {
...
+ * If the body is truncated, the circular website link will be removed.
+ * A warning that the email has been truncated is added and the circular website link is added back in.
+ * An additional buffer is allowed for for variation in url length and truncation warning.
+ */
+export function truncateEmailBodyIfNecessary({
+  body,
+  circularId,
+}: {
+  body: string
+  circularId?: number
```
lpsinger, line 106: In what situation would `circularId` be undefined?

lpsinger (changes requested):
The byte limit applies to all of `TemplateData`. Some more corner cases:

- `subject` may be arbitrarily long; it needs to be truncated as well.
- Actually what matters is the JSON string representation of the body, which may grow longer if it contains escape sequences. And extreme case: suppose that the body consists of a string of 262144 double quotes (`""""""""""...`) which would be truncated to an appropriate length by this code, but then encoded as `"\"\"\"\"\"\"\"\"\"\"..."` which is double the length, and would trigger the Amazon SES error.

Courey:
Instead of worrying about recursively truncating to avoid json encoding adding escape characters and putting us over the limit, Judy wanted it to just aggressively truncate to give a large margin so we don't have to worry about complexity. Should the 60,000 byte buffer not be enough, we can adjust the value to truncate even more aggressively. https://github.com/nasa-gcn/gcn.nasa.gov/issues/2899#issuecomment-2959486875

app/lib/email.server.ts
```diff
@@ -89,6 +91,42 @@ function maybeThrowSES(e: any, warning: string) {
...
+  subject: string
+  circularId?: number
+}) {
+  const textBody = getBody(body)
+  const jsonTemplateData = JSON.stringify({
+    subject,
+    body: textBody,
+  })
+
+  if (Buffer.byteLength(jsonTemplateData, 'utf-8') <= 262144) return textBody
```
lpsinger: The maximum length in the check here must match the maximum length in the call to `truncate`, otherwise the corner cases may still occur.

```suggestion
  if (Buffer.byteLength(jsonTemplateData, 'utf-8') <= 200000) return textBody
```

Even better would be to define a constant, so that you don't have to repeat the literal value.

lpsinger (changes requested):

lpsinger:
> Instead of worrying about recursively truncating to avoid json encoding adding escape characters and putting us over the limit, Judy wanted it to just aggressively truncate to give a large margin so we don't have to worry about complexity. Should the 60,000 byte buffer not be enough, we can adjust the value to truncate even more aggressively. [#2899 (comment)](https://github.com/nasa-gcn/gcn.nasa.gov/issues/2899#issuecomment-2959486875)

I believe that the worst possible case is that the body consists entirely of [Unicode control characters](https://en.wikipedia.org/wiki/Unicode_control_characters), each of which has a string length of 1 but in JSON is encoded to 6 characters. For example, the JavaScript string literal `'\0'` is encoded to the JSON content `"\u0000"`. That gives you the maximum string length (excluding the lengths of the JavaScript keys, quotes, commas, whitespace, etc.) of floor(262144/6) =43,690, which is a lot smaller than the value you currently have.

Also make sure that you truncate _any_ field over which the user has sufficient control that they could inject a string of unbounded length.

app/lib/email.server.ts
```diff
@@ -109,22 +105,70 @@ export function truncateEmailBodyIfNecessary({
...
-  const jsonTemplateData = JSON.stringify({
+  const templateData = {
     subject,
     body: textBody,
-  })
+  }
+  const jsonTemplateData = JSON.stringify(templateData)
+  const awsSESByteLimit = 262144
+
+  if (Buffer.byteLength(jsonTemplateData, 'utf-8') <= awsSESByteLimit)
```
lpsinger: `Buffer.byteLength(jsonTemplateData, 'utf-8')` is not an accurate estimate of the byte length of the JSON-encoded data because it does not include escaped strings.

The only way to really do this reliably is to actually encode it to JSON, encode to UTF-8, and measure the length.

You can avoid this kind of bug by factoring out the length check so that it is not duplicated. Or, you can restructure the code so that the check is only done once, in the condition of the `while` loop. (The body of the `while` loop would be entirely skipped if the body is already sufficiently short.)

lpsinger (changes requested):

app/lib/email.server.ts
Courey: can you explain this more? What I see is that on line 112, I am stringifying the data, which should be the json encoding which should include escaped strings, and then on line 115 I am taking the stringified JSON data and converting it to utf-8 and getting the buffer size.

app/lib/email.server.ts
lpsinger: What does this change have to do with https://github.com/nasa-gcn/gcn.nasa.gov/issues/2899?

lpsinger (changes requested):
- When there is a Node.js function that has an equivalent standard ECMAScript function that works in any JavaScript engine, please use the standard ECMAScript version for portability. It looks like Buffer.byteLength is a Node.js-specific feature, but there are several more portable replacements.
- Please factor out the code to truncate a string to a maximum byte length when JSON-encoded, into a standalone function without side effects, and add unit tests that exercise nicely behaved strings and also pathological strings that require escape sequences.
- Reuse that function in two places, to truncate the body and to truncate the subject.
- Drop the dependency on truncate-utf8-bytes. It would have been relevant if we were simply encoding JavaScript strings to UTF-8, but because we are encoding it to JSON, which requires lots of additional escaping that may inflate the byte length in unpredictable ways. I think it is acceptable to use `String.prototype.substring` to truncate strings; I don't think it can split Unicode code points.
- Please format the commit message, PR title, and PR description, to conform to the style guide.

app/lib/email.server.ts
Courey: I'm not sure what you are asking. when this issue was assigned to me, you said to look at this comment that you made: https://github.com/nasa-gcn/gcn.nasa.gov/issues/2899#issuecomment-2798899590 and to do that instead of just make a size check. that way the circular in the database would be complete and the email would be truncated to prevent issue.

app/lib/email.server.ts
lpsinger, line 1: I see now why this PR brought in some new imports (`origin`) and some new arguments (`circularId`) that seemed unrelated... it's because you added the Circulars-specific truncation code to `email.sever.ts` rather than `circulars.server.ts`. Would you please move the new truncation code over to `circulars.server.ts`? I think that will result in a smaller diff, and it will prevent concerns related to Circulars from leaking into code that deals with all kinds of emails.

Courey:
closing because https://github.com/nasa-gcn/gcn.nasa.gov/pull/3245 is faster to get in

PR #3122, opened 2025-05-27
Announce synonyms

jracusin:
# Description
Updated date and schema version in news announcement.


jracusin:
Trying to make minor updates to previous PR that was already merged. This didn't work right, will do another one branched properly off of main.

PR #3124, opened 2025-05-27
removing page return value from synonym search as it messes up pagina…

### LINKED ISSUE #3125: pagination looks weird on production group view
pagination looks really wrong on production

Image

fix pagination and make it function the same as the circulars view

Courey:
…tion

pagination for the group view was very wonky. 

Screenshot 2025-05-27 at 2 23 15 PM

I discovered that the event search is returning a page variable. This is not how the regular circulars search works.

page_overwrites

The page value that was returned from that search function was overwriting the page value that existed already within that function (which was correct).

Removing the page return fixed pagination.

Screenshot 2025-05-27 at 2 26 06 PM

Resolves #3125

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -53,7 +53,6 @@ export async function searchSynonymsByEventId({
   items: SynonymGroup[]
   totalItems: number
   totalPages: number
-  page: number
   queryFallback: boolean
 }> {
   const client = await getSearchClient()
@@ -128,7 +127,6 @@ export async function searchSynonymsByEventId({
     items: results,
     totalItems,
     totalPages,
-    page,
     queryFallback: false,
   }
 }


dakota002 (approved):
Looks good, which PR removed this?

app/routes/synonyms/synonyms.server.ts
```diff
@@ -53,7 +53,6 @@ export async function searchSynonymsByEventId({
   items: SynonymGroup[]
   totalItems: number
   totalPages: number
-  page: number
```
lpsinger, line 56: Please restore this argument. Removing the `page` argument will break pagination by causing page changes to have no effect.

lpsinger (changes requested):
A more succinct description of this bug fix would be: "Remove `page` return value from `searchSynonymsByEventId` which shadowed the loader's return value."

app/routes/synonyms/synonyms.server.ts
Courey: Can you explain this a little more? What that is doing is saying that it returns a promise with a page field. If I remove the page from the return, shouldn't I also remove it from where the shape of what I'm returning is defined? How does removing it from the promise affect pagination?

lpsinger: Sorry, I misread.

lpsinger (approved):

PR #3149, opened 2025-06-10
DO NOT MERGE/gather event ids to repair

### LINKED ISSUE #3061: create a script that would gather circular data to be updated via api
# Description
We need a script that can go through the archive and determine what circulars need to be updated and what the update should be.

# Acceptance criteria
A script exists that would determine what circulars need the eventId to be updated.
This script should iterate through changes and do an update to the circular via and API endpoint

Courey:
# Description
This is a script that uses the subject matchers in the app to go through the archive and find circulars that need to have their eventId's updated or corrected. Detailed instructions on how to use this script are found at the top of the script file.

# Related Issue(s)
Resolves #3061

# Testing
I have run this locally.

### DIFF
--- a/app/migrations/updates/index.ts
+++ b/app/migrations/updates/index.ts
@@ -7,6 +7,9 @@
  * 4. Run `npm run build`
  * 5. Run `node build/migrations/updates/index.mjs`
  * 6. Your file with circulars that need to be updated will be at the top level directory.
+ *
+ * If you are using visual studio code, you can open the `event_id_updates.json` and make it more readable
+ * by pushing command + shift + P and typing `Format Document`.
  */
 import * as fs from 'fs'
 import * as path from 'path'
--- a/app/migrations/updates/index.ts
+++ b/app/migrations/updates/index.ts
@@ -1,0 +1,57 @@
+/**
+ * STEPS:
+ * You do NOT need to have the app running.
+ * 1. Download a fresh copy of the archive
+ * 2. Unzip it
+ * 3. Move it into the `app` folder. it should be named archive.json
+ * 4. Run `npm run build`
+ * 5. Run `node build/migrations/updates/index.mjs`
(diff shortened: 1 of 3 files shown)


app/migrations/updates/index.ts
```diff
@@ -0,0 +1,67 @@
...
+interface UpdateData {
+  circularId: number
+  eventId: string
+  oldEventId: string
+}
+
+function processFiles() {
+  const folderPath = './app/archive.json'
+  const results = [] as UpdateData[]
+  fs.readdir(folderPath, (err, files) => {
```
lpsinger, line 26: Please use the `fs/promises` module and async/await syntax over the `fs` module and callback syntax.

app/migrations/updates/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (err) {
+      console.error('Error reading directory:', err)
+      return
+    }
+    files.forEach((file) => {
```
lpsinger: Please use:

```js
result = input.map((item) => ...)
```

rather than:

```js
const results = []
input.forEach((...) => {results.push(...)})
```

app/migrations/updates/index.ts
```diff
@@ -0,0 +1,67 @@
...
+import { parseEventFromSubject } from '~/routes/circulars/circulars.lib'
+
+interface UpdateData {
+  circularId: number
+  eventId: string
+  oldEventId: string
+}
+
+function processFiles() {
+  const folderPath = './app/archive.json'
```
lpsinger, line 25: Where does this file, archive.json, come from? Our downloadable archive is a tarball of JSON files, not one big JSON file.

Courey: you required the downloaded folder to be called archive.json
That is the name of the directory unpacked from the tarball.

Courey:
this has been run

Issue #2626, opened 2024-10-15
CHIME JSON Notices


jracusin:
# Description
Add CHIME notices

# Acceptance criteria

- [x] create [topics/ACLs](https://gcn.nasa.gov/docs/notices/producers) (GCN Team)
- [x] add producer POC to producers git team  (GCN Team)
- [x] finalize [JSON schema](https://gcn.nasa.gov/docs/notices/schema) (Producers)
- [x] add/update [producer mission page](https://gcn.nasa.gov/missions) (Producers)
- [x] update [mission page table](https://gcn.nasa.gov/missions) (GCN Team)
- [x] add to [quickstart](https://gcn.nasa.gov/quickstart) (GCN Team)
- [x] test and validate production notices (Producers + GCN Team)
- [x] remove [feature flag](https://gcn.nasa.gov/docs/contributing/feature-flags) (GCN Team)

dakota002:
Schema PR: https://github.com/nasa-gcn/gcn-schema/pull/217

dakota002:
Draft PR for quickstart and basic mission page: https://github.com/nasa-gcn/gcn.nasa.gov/pull/2702

Vidushi-GitHub:
@jracusin please review: https://github.com/nasa-gcn/gcn-schema/pull/226

jracusin:
CHIME schema [#217](https://github.com/nasa-gcn/gcn-schema/pull/217)

Vidushi-GitHub:
@jracusin please review https://github.com/nasa-gcn/gcn.nasa.gov/pull/3046

PR #3152, opened 2025-06-13
updated mission page and fixed quickstart for CHIME

jracusin:
# Description
In preparation for going live with CHIME notices, I've updated the mission table to point to them and made some corrections to the Quickstart with the right topic name, and making it link properly.

# Related Issue(s)
Addresses part of #2626 

# Testing
Local dev


app/routes/missions.chime/route.mdx
```diff
@@ -7,11 +7,6 @@ import { feature } from '~/lib/env.server'
 
 import logo from './logo.svg'
 
-export async function loader() {
-  if (!feature('CHIME')) throw new Response(null, { status: 404 })
```
dakota002, line 11: Can you also remove the import of `feature` if it is no longer being used on this page?

jracusin: Strange, I did that in my second commit and it's gone in my local version, which says it's up to date with the branch.

PR #3186, opened 2025-07-08
Adds lambda and cloudformation for a cloudtrail event bridge to log the last used times for app clients

### LINKED ISSUE #3182: Add usage tracking for app clients
# Description

We allow users to make as many app clients as they want. This has lead to many existing app clients that sit unused in Cognito. As a security measure, we should track the time of last use of app clients to determine which have been sitting for a while. 

# Acceptance criteria

- [ ] Event bridge rule to send `Token_POST` AWS Service Events to a Lambda
- [ ] Lambda function to update the `lastAccessed` field of an app client
- [ ] Add a display for the `lastAccessed` on the client credential card component

dakota002:
# Description
Adds a lambda and cloudformation template to send `Token_POST` events to a lambda

# Related Issue(s)

Resolves #3182


dakota002:
@lpsinger  how do I tell the cloudformation template the correct arn to use (line 28 in the deploy section)

app/routes/user.credentials/client_credentials.server.ts
```diff
@@ -136,6 +137,7 @@ export class ClientCredentialVendingMachine {
       client_id,
       scope,
       sub: this.#sub,
+      lastAccessed: Date.now(),
```
lpsinger: IMHO, we should not count simply creating the client credential as using it.

src/plugins/cloudtrail-bridge.js
lpsinger, line 1: This plugin can now handle _all_ EventBridge sources, not just CloudTrail events. Suggest renaming this to `eventbridge.js` and replacing related strings as well.

app/cloudtrail-bridge/app-clients/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  eventCategory: string
+  tlsDetails: TlsDetails
+}
+
+export async function handler(
+  event: EventBridgeEvent<'AWS Service Event via CloudTrail', EventDetail>
+) {
+  const db = await tables()
+  const eventName = event.detail.eventName
+  if (eventName !== 'Token_POST') return
```
lpsinger: I think that you can remove this because it is in the filter rule.

```suggestion
```

app.arc
```diff
@@ -13,6 +13,13 @@ circulars
 support
   src build/email-incoming/support
 
+@cloudtrail-bridge
```
lpsinger:
```suggestion
@eventbridge
```

And see related comments about the plugin itself.

app/routes/user.credentials/client_credentials.server.ts
dakota002: I'll remove this, that makes sense to me too.

lpsinger:
I am going to [deploy this manually](http://localhost:3333/docs/contributing/deployment#to-deploy-manually) to dev so that we can start observing its behavior.

PR #3201, opened 2025-07-30
Update package-lock.json

lpsinger:
When I run `npm i` on my machine, it makes these changes to the lock file. Evidently we had some unused dependencies.


dakota002:
Seems good to me

PR #3217, opened 2025-08-06
Add WithFeature fallback

### LINKED ISSUE #3216: WithFeature doesn't allow for fallback
# Current behavior
WithFeature function only allows for including content when the feature flag is on, it doesn't have a fallback option for when it is off like UseFeature.

# Expected behavior
WithFeature in .mdx pages should behave like UseFeature in .tsx pages.

# Steps to reproduce
Mission table (https://gcn.nasa.gov/missions) uses feature flags that are always in the off condition whether or not the feature flag is in your local .env.

# Environment
Not relevant

jracusin:
# Description
We were incorrectly using `WithFeature` in the mission table page with the syntax for `UseFeature` for when the feature flag was or was not set.  This PR modifies the WithFeature function such that it maintains its old usage and can handle scenarios of the feature being On or Off.  This PR modifies the function and corrects the usage in the mission table. @Courey put together the function

# Related Issue(s)
Resolves #3216

# Testing
local


lpsinger (changes requested):
This is a really weird implementation: the component inspects its children and decides which to include and which not to. This is very unusual compared to common place React patterns. Why did you do it this way?

Courey:
> This is a really weird implementation: the component inspects its children and decides which to include and which not to. This is very unusual compared to common place React patterns. Why did you do it this way?

I did it 3 different ways with different tradeoffs, gave them to Judy to play around with, and she liked this one because it takes markdown for the fallback and was backwards compatible.  

another option was this way, but it wasn't backwards compatible so all feature flags in .mdx files would need to be changed (this could be a fine option, i'm not particular). It uses a slotted design:
```
type WithFeatureProps = {
  children: ReactNode
} & Record<string, boolean>

export function WithFeature({ children, ...features }: WithFeatureProps) {
  const featureKey = Object.keys(features)[0]
  const isOn = useFeature(featureKey)

  const onChild = getSlot(children, WithFeatureOn)
  const offChild = getSlot(children, WithFeatureOff)

  return <>{isOn ? onChild : offChild}</>
}

function getSlot(children: ReactNode, slot: React.ElementType) {
  const array = React.Children.toArray(children) as ReactElement[]
  return array.find(child => child.type === slot) ?? null
}

export function WithFeatureOn({ children }: { children: ReactNode }) {
  return <>{children}</>
}

export function WithFeatureOff({ children }: { children: ReactNode }) {
  return <>{children}</>
}

WithFeature.On = WithFeatureOn
WithFeature.Off = WithFeatureOff
```

There was also this way, but the tradeoff here is that when you call it, you have to pass html into the fallback argument:
```
export function WithFeature({
  children,
  fallback = null,
  ...features
}: {
  children: ReactNode;
  fallback?: ReactNode;
} & Record<string, boolean>) {
  const featureKey = Object.keys(features)[0];
  const featureOn = useFeature(featureKey);
  return <>{featureOn ? children : fallback}</>;
}
```
so you would end up with this:
```
<WithFeature NOTREAL fallback={<p>Feature is OFF</p>}>
  Feature is ON
</WithFeature>
```

So the code in this branch is done the way that it is because it is backwards compatible, you can pass markdown instead of html. Use whichever you want, I'm impartial.

PR #3218, opened 2025-08-07
Add WithoutFeature component

### LINKED ISSUE #3216: WithFeature doesn't allow for fallback

lpsinger:
Fixes #3216.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -187,6 +187,15 @@ export function WithFeature({
   return <>{useFeature(Object.keys(features)[0]) && children}</>
 }
 
+export function WithoutFeature({
+  children,
+  ...features
+}: {
+  children: ReactNode
+} & Record<string, boolean>) {
+  return <>{useFeature(Object.keys(features)[0]) || children}</>
+}
+
 export function useOrigin() {
   return useLoaderDataRoot()?.origin
 }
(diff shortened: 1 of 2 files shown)


lpsinger:
@jracusin, @dakota002, do you want this? If not, I will close it.

dakota002:
I think the idea makes sense, but what if instead of the same component with an `off` flag it was just another component defined in the same file as `WithoutFeature`?

```html
<WithFeature FEATURE_FLAG>
    Render if present
</WithFeature>
<WithoutFeature FEATURE_FLAG>
    Otherwise render me
</WithoutFeature>
```

lpsinger:
> I think the idea makes sense, but what if instead of the same component with an `off` flag it was just another component defined in the same file, just called `WithoutFeature`?

Done.

dakota002 (approved):
Looks good to me

PR #3225, opened 2025-08-11
try setting archive tests to have longer timeout

### LINKED ISSUE #3163: playwright tests failing
# Current behavior
In CI/CD tests, playwright often fails (after 5 min timeout), potentially due to check in Circular archive component existing.  It is not failing when deployed or in local development environment.

# Expected behavior
Tests should pass

# Steps to reproduce
See PR #3147 for an example

# Environment

tylerbarna:
# Description
Attempting to solve the playwright test timeout. On first pass, it seems that the archive tests did not have the "slow" flag enabled like several of the other tests, which may be to blame.

# Related Issue(s)
Resolves #3163 

# Testing
1. Run Playwright Tests

### DIFF
--- a/__playwright__/circulars/archive.spec.ts
+++ b/__playwright__/circulars/archive.spec.ts
@@ -9,6 +9,7 @@ test.describe('Circulars archive page', () => {
   test('responds to changes in the number of results per page', async ({
     page,
   }) => {
+    test.slow()
     await page.goto('/circulars')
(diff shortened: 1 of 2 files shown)


lpsinger (commented):
Very nice! @dakota002, would you please review this?

dakota002 (approved):
Looks good to me!

PR #3232, opened 2025-08-11
cache playwright browsers when running tests

tylerbarna:
# Description
Attempts to cache playwright browsers so they don't need to be installed by each test, which _should_ speed github actions somewhat. Follows https://github.com/microsoft/playwright/issues/7249

# Related Issue(s)
Related to #3163 

# Testing
1. Run Playwright Tests
2. They should run faster


lpsinger:
Just lack of activity. @tylerbarna, if you'd like to pursue this, please reopen, rebase, and make sure that the pipeline passes.

PR #3245, opened 2025-08-12
Truncate Circular text to fit in SES request limits

### LINKED ISSUE #2899: Check Circulars message size before accepting submission

lpsinger:
Fixes #2899. Fixes #3110.

### DIFF
--- a/__tests__/utils.ts
+++ b/__tests__/utils.ts
@@ -5,7 +5,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import { joinListWithOxfordComma } from '~/lib/utils'
+import { joinListWithOxfordComma, truncateJsonMaxBytes } from '~/lib/utils'
 import { stripTags } from '~/lib/utils.server'
 
 describe('stripTags', () => {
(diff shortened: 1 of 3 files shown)


lpsinger:
> I've been waiting on your review on my branch that fixes this for over a week. If you needed to give feedback that would be the appropriate place to do that, not just refuse to review it and make your own branch.

I did review it, but you did not make the change requested in comment https://github.com/nasa-gcn/gcn.nasa.gov/pull/3110#discussion_r2178730783:

> You can avoid this kind of bug by factoring out the length check so that it is not duplicated. Or, you can restructure the code so that the check is only done once, in the condition of the while loop. (The body of the while loop would be entirely skipped if the body is already sufficiently short.)

lpsinger:
> When you look at that image, where I moved the size check into the conditional of the while loop, what about that change does not meet the request:
> 
> > Or, you can restructure the code so that the check is only done once, in the condition of the while loop. (The body of the while loop would be entirely skipped if the body is already sufficiently short.)

Alright, please address https://github.com/nasa-gcn/gcn.nasa.gov/pull/3110#pullrequestreview-3112044291.

dakota002 (approved):

PR #3240, opened 2025-08-11
Update messaging on GCN Kafka

jracusin:
# Description
Updates messaging graphic on home page for recommending GCN Kafka for available topics.

# Related Issue(s)
Resolves #2310

# Testing
local dev

PR #3246, opened 2025-08-12
Refactor two copies of notice formats ad into one component

lpsinger:


lpsinger:
Closing in favor of #3240.

PR #3250, opened 2025-08-14
Playwright: address failing tests

tylerbarna:
# Description
Experimenting with different means of resolving the playwright timeout errors. This boosts the number of retries before giving up on a test as well as the timeout on the webServer.

# Related Issue(s)
#3163

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -42,7 +42,7 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: Boolean(process.env.CI),
   /* Retry on CI only */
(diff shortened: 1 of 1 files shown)


tylerbarna:
Tests were successful for a778560 5x in a row (note: one test was skipped during the second run; this was due to the docker image crashing rather than the playwright test causing a timeout error. This is a known, rarer issue that is not specific to playwright tests).

tylerbarna:
@dakota002 I think it's at least somewhat more consistent now?

dakota002 (approved):
Awesome, this looks great.

We should keep an eye out for failures in the future and see if there are other places to keep making improvements.

PR #3260, opened 2025-08-19
Add team page

jracusin:
# Description

Add team page listing current and former members of the GCN team.

# Related Issue(s)

# Testing

local dev


app/routes/docs.team/route.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <TeamCard
+          name="Dakota Dutko"
+          affiliation="NASA/GSFC/ADNET"
+          photo={headshot}
+        >
+          Full-Stack Developer
+        </TeamCard>
+        <TeamCard
+          name="Courey Elliott"
+          affiliation="Lousiana State University/ NextSource"
```
tylerbarna:
```suggestion
          affiliation="Lousiana State University/NextSource"
```

app/routes/docs.team/route.tsx
```diff
@@ -0,0 +1,101 @@
...
+  photo,
+}: {
+  children: ReactNode
+  name: string
+  affiliation: string
+  photo: string
+}) {
+  return (
+    <>
+      <div className="usa-card__container margin-bottom-2 grid-col-4">
```
lpsinger: Use `<CardGroup>` here instead of a div.

lpsinger (changes requested):
- Add a link to this page.
- Add GitHub handles and links to GitHub profiles.
- Add headshots for each person. For me, you can use the image from my GitHub profile.

app/routes/docs.team/route.tsx
jracusin: This space is there in order to get the right line break.

jracusin:
Screenshot 2025-09-02 at 2 11 49 PM
Screenshot 2025-09-02 at 2 12 03 PM

app/routes/docs.team/route.tsx
```diff
@@ -0,0 +1,318 @@
...
+  href: string
+  tags: string[]
+  selectedTags: string[]
+}) {
+  const tagSet = new Set(tags)
+  return (
+    <>
+      {(selectedTags.length == 0 ||
+        selectedTags.every((tag) => tagSet.has(tag))) && (
+        <CardGroup className="usa-card__container margin-bottom-1 tablet:grid-col-4">
```
Courey: If we use Card instead of CardGroup here, some of the styling is automatically handled. See how these corners don't really line up if the tags are different:
Screenshot 2025-09-02 at 2 47 18 PM
If you use Card instead of CardGroup, you can remove the className usa-card__container and it should align better:
Screenshot 2025-09-02 at 2 48 30 PM

app/routes/docs.team/route.tsx
```diff
@@ -53,51 +42,34 @@ function TeamCard({
...
-      )}
-    </>
+    <Card className="grid-col-fill">
+      <CardMedia imageClass="bg-white margin-bottom-neg-2">
+        <img
+          src={photo}
+          width="50"
+          height="50"
+          loading="lazy"
+          alt="headshot of GCN team member"
```
lpsinger: A "headshot" has the unfortunate second meaning of a gunshot to the head.

```suggestion
          alt="Photograph of GCN team member"
```

app/routes/docs.team/route.tsx
```diff
@@ -53,51 +42,34 @@ function TeamCard({
...
-        </Card>
-      )}
-    </>
+    <Card className="grid-col-fill">
+      <CardMedia imageClass="bg-white margin-bottom-neg-2">
+        <img
+          src={photo}
+          width="50"
+          height="50"
+          loading="lazy"
```
lpsinger: Why lazy loaded?

```suggestion
```

app/routes/docs.team/route.tsx
```diff
@@ -53,51 +42,34 @@ function TeamCard({
...
-          </CardFooter>
-        </Card>
-      )}
-    </>
+    <Card className="grid-col-fill">
+      <CardMedia imageClass="bg-white margin-bottom-neg-2">
+        <img
+          src={photo}
+          width="50"
+          height="50"
```
lpsinger: Are these images actually square?

app/routes/docs.team/route.tsx
```diff
@@ -53,51 +42,34 @@ function TeamCard({
...
-              <Tag key={tag} className="font-body-4xs">
-                {tag}
-              </Tag>
-            ))}
-          </CardFooter>
-        </Card>
-      )}
-    </>
+    <Card className="grid-col-fill">
+      <CardMedia imageClass="bg-white margin-bottom-neg-2">
```
lpsinger: Why the custom margin?

app/routes/docs.team/route.tsx
jracusin: Yes, I cut them to square, otherwise the heights of the cards end up all different.

app/routes/docs.team/route.tsx
jracusin: Once we switched to `Grid`, it is no longer needed.  Removed.

lpsinger (commented):
I will have some typographical suggestions on this, but I can't do that today because I need to finish my grant proposals.

However, I would suggest removing the group photos. If you want them, I would pick and keep just _one_.

Vidushi-GitHub:
> > I will have some typographical suggestions on this, but I can't do that today because I need to finish my grant proposals.
> > However, I would suggest removing the group photos. If you want them, I would pick and keep just _one_.
> 
> I like the Team photos, and it's difficult to keep just one because we don't have only one with everyone in it. I'm open to suggestions to make them less prominent. I could move them all to the end and/or make them smaller.

I think two photos, the one at NASA and other at MN are sufficient. Otherwise, it looks like an album gallery.

jracusin:
I removed team photos after discussion of their use elsewhere and rebased.  I also add another TeamCard for when we don't want to link to GitHub (e.g. past team members).

jracusin:
I replaced the href CSS with the Link component and added the GitHub handle as part of the TeamCard component. I'm not sure what other incline CSS can be replaced with USWDS components.

I asked Teresa for a photo, but haven't heard back yet.  I'd like her permission.

PR #3266, opened 2025-08-21
Lint commit messages automatically

lpsinger:


tylerbarna:
Would it be beneficial to add our own whitelist? I see in the output that is an option

lpsinger:
> Would it be beneficial to add our own whitelist? I see in the output that is an option

Shouldn't be necessary with https://github.com/mristin/opinionated-commit-message/pull/144 merged.

However, notice that there is a downside to this GitHub Action: grammatically correct commit messages that start with an adverb will be rejected, like this:

> Automatically lint commit messages

PR #3272, opened 2025-08-22
Reverse order of circulars event view

### LINKED ISSUE #3271: Add option to view circulars for an event in reverse chronological order
# Description
Currently, the event view for circulars only shows in reverse chronological order. For ongoing novel events, such as those related to AT2025ulz, it would be beneficial to see the most recent circulars first.

# Acceptance criteria
- Current behaviour is default
- Have a UI element that allows for switching between chronological order directions
- Reload the page once it is set

tylerbarna:
# Description

Reverses the order of the circulars when grouped by event. This makes it so the most recent circular is shown first

# Related Issue(s)

Resolves #3271 

# Testing

1. Navigate to a circular with an event
2. Click on event
3. Circulars should be shown in reverse order


lpsinger (changes requested):
You need to also remove the `.reverse()` in `api.synonym.circular.ts`.

__playwright__/utils.ts
lpsinger, line 1: These changes look like they snuck in from an unrelated PR.

lpsinger (changes requested):

__playwright__/utils.ts
tylerbarna: yep, sorry about that

lpsinger (approved):

PR #3275, opened 2025-08-22
Consolidate Circulars button bar

lpsinger:
- Display Next and Previous buttons at top so that the rest of the page does not visibly move when you press them.
- Shorten button text.
- Fix incorrect vertical positioning of disclosure arrow.
- Replace nonbreaking spaces with CSS positioning adjustment.

# Before

Screenshot 2025-08-21 at 22 46 44

# After

Screenshot 2025-08-21 at 22 44 07

### DIFF
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -95,6 +95,48 @@ export default function () {
           Back
         </Link>
         <ButtonGroup type="segmented">
+          {Number.isFinite(previousCircular) ? (
+            <Link
+              to={`/circulars/${previousCircular}${searchString}`}
+              className="usa-button"
+            >
+              <Icon.NavigateBefore
+                role="presentation"
+                className="margin-y-neg-2px"
+              />
+              Previous
+            </Link>
+          ) : (
+            <Button type="button" className="usa-button" disabled aria-disabled>
+              <Icon.NavigateBefore
+                role="presentation"
+                className="margin-y-neg-2px"
+              />
+              Previous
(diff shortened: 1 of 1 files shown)


tylerbarna (changes requested):
Overall, I like the changes, but I think the next and previous buttons still make more sense at the bottom of the page. Adding them to the top makes it more cluttered, and "back" and "previous" being adjacent may be somewhat confusing for users

jracusin:
"Back", "Previous", and "Next" should not be in the same button group.  Correct and edit also only exist for moderators. Is "correct" as clear to users as "request correction"?

jracusin:
"Edit" is only visible to the 4 moderators.  Would it look better without "Edit" for the majority of users?

tylerbarna:
I worry that having right/left directional buttons directly adjacent to one another could be somewhat confusing visually, do you have any thoughts on using up/down directional indicators for next/previous? This would match the reverse chronological order we have on the main page

tylerbarna:
This is likely outside the scope of this PR but, given the increasing number of elements at the top of the page, would it be beneficial to alter the behavior on mobile so some of the buttons are hidden and can be expanded to be shown?

lpsinger:
> "Edit" is only visible to the 4 moderators. Would it look better without "Edit" for the majority of users?

You're kind of reading my mind. What I was going to experiment with next was making it just a singe "Edit" button and deferring the user's choice of whether to request a correction or (if permitted) save the change on the following form. Stay tuned.

jracusin:
This looks good.  We also need to update the documentation on corrections to match.

PR #3279, opened 2025-08-22
allow moderator to delete synonyms with no members

### LINKED ISSUE #3148: Delete synonym group with no members
# Description
Empty synonym groups can occur when an eventId is changed.  After all members of a group are removed, the moderator should have control to delete the empty group.

# Acceptance criteria
- connect to moderator remove functionality

Courey:
# Description
Add ability for moderator to completely delete a synonym group where there are no circulars that are related to it by eventId.

# Related Issue(s)
Resolves #3148 

# Testing
local

### DIFF
--- a/app/routes/synonyms/synonyms.server.ts
+++ b/app/routes/synonyms/synonyms.server.ts
@@ -176,7 +176,7 @@ export async function moderatorCreateSynonyms(synonymousEventIds: string[]) {
  *   - If a new synonym was not created, it updates the initialDate on the existing synonym.
  *
  * If there is an oldEventId:
- *   - It deletes the old eventId synonym if it's empty.
+ *   - It sets the old synonym initialDate to -1 if no circulars have that eventId.
  *   - If it's not empty, it updates the initialDate on the remaining synonym.
  *
  * @param newCreatedOn - the Circular createdOn timestamp for the new circular version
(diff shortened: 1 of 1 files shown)


app/routes/synonyms/synonyms.server.ts
```diff
@@ -334,16 +334,29 @@ export async function deleteSynonyms(synonymId: string) {
...
-      Item: {
-        synonymId: crypto.randomUUID(),
-        eventId,
-        slug: slug(eventId),
-        initialDate: await getOldestDate(eventId),
-      },
-    },
-  }))
+  const writePromises = results.Items.map(async ({ eventId }) => {
+    const initialDate = await getOldestDate(eventId)
```
lpsinger, line 338: This makes the deletion operation accidentally quadratic in the number of synonyms.

lpsinger (changes requested):
Can you think of a more direct way to test if a Circulars group is empty, rather than checking the earliest created date?

Please conform to the commit message style guide: the commit summary should be capitalized.

Courey:
> Can you think of a more direct way to test if a Circulars group is empty, rather than checking the earliest created date?

Absolutely. The getSynonymMembers function will return an array of circulars that belong to that eventId. It is used by the getOldestDate function. If I call the getSynonymMembers function directly to check for empty synonyms, which would need to happen for each synonym, in the case that the synonym wasn't empty (the much more frequent case) it would then have to call the getOldestDate which would again call the getSynonymMembers function. So to be explicit, you are adding in the same call twice in most cases. So i chose to use the oldest date to tell if it was empty so that in the case that it was not, I would already have the initialDate to use and wouldn't have to duplicate the getSynonymMembers call.

app/routes/synonyms/synonyms.server.ts
Courey: Could you explain this a bit more? I'm not sure what you mean. The results will only contain unique eventIds since that is the partition key for the synonyms table, so there will only ever be 1 per eventId. So there would be no duplicates in results. the getOldestDate function and the function it uses to accomplish its job, getSynonymMembers, are not table scans. So I'm unsure how this could be accidentally quadratic.

lpsinger: It's quadratic in the sense that it is O(number of synonyms x number of Circulars per synonym). Do you agree?

Courey: That does make sense, but I believe what you are saying is that by running this function when you won't use it on deletes makes it less scalable? All that getOldestDate function does is takes the date off the results of the call to getSynonymMembers. If I am going to find out if there are any circulars with this eventId, i'm going to need to count how many members there are for a synonym, which would be getSynonymMembers. So the only additional thing that happens by calling getOldestDate is that it finds the min result of the createdOn field for the results. If there aren't any results, it returns -1. 

how would you approach it instead?

Courey: This is how I was thinking about it. This is how I have it in this PR:
synonym_delete_as_is

If I query to see if there are circulars with that eventId and then only call the getOldestDate function if it's not empty, in most cases it duplicates the call to getSynonymMembers.
counting_circulars_first

lpsinger (approved):

PR #3297, opened 2025-08-28
Clean up Circulars previous/next button button style

lpsinger:
- Use 'next' and 'previous' icons that are distinct from 'back'
- Use standard USWDS left alignment for form controls
- Combine into a segmented button group

Screenshot 2025-08-28 at 15 25 53


tylerbarna (approved):
Still open to discussion about consolidation of the navigation buttons as part of one row, but it's good to adhere to USWDS standard in the meantime

PR #3304, opened 2025-09-02
Put expired creds only behind feature flag

### LINKED ISSUE #3303: Add feature flag for credential deletion rollout
# Description
We started tracking last used client credential dates around August 11, putting 30 days around Sept 11.  We want to make sure people get at least 1 warning email before deletion, but not delete prematurely.  So we need to move the feature flag on Sept 9 to only disable deletion.

# Acceptance criteria
- Move feature flag and deploy on Sept 9

dakota002:
This moves the feature flag such that only the warning credentials list will get populated. 

It also requires an update to the condition of what defines the warning so that we can send emails about creds that would otherwise just have been marked for deletion

Resolves #3303

### DIFF
--- a/app/scheduled/app-client-expire/index.ts
+++ b/app/scheduled/app-client-expire/index.ts
@@ -64,7 +64,7 @@ export async function handler() {
         creds,
         (cred) => (cred.lastUsed ?? cred.created) < deletionCutoff
       )
-      if (feature('APP_CLIENT_TRACKING'))  {
+      if (feature('APP_CLIENT_TRACKING')) {
         expiredCreds.push(...moreExpiredCreds)
         warningCreds.push(...moreWarningCreds)
       } else {
--- a/app/scheduled/app-client-expire/index.ts
+++ b/app/scheduled/app-client-expire/index.ts
@@ -29,7 +29,6 @@ type CredentialInfo = {
 }
 
 export async function handler() {
-  if (!feature('APP_CLIENT_TRACKING')) return
   const db = await tables()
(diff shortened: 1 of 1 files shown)


dakota002:
Which commit message needs to be changed? And i've approved #3306, i'll rebase once merged

app/scheduled/app-client-expire/index.ts
```diff
@@ -65,8 +64,9 @@ export async function handler() {
         creds,
         (cred) => (cred.lastUsed ?? cred.created) < deletionCutoff
       )
-      expiredCreds.push(...moreExpiredCreds)
-      warningCreds.push(...moreWarningCreds)
+      if (feature('APP_CLIENT_TRACKING')) expiredCreds.push(...moreExpiredCreds)
+      // Put both expired and warning together for the first few iterations of the scheduled task
+      warningCreds.push(...moreWarningCreds, ...moreExpiredCreds)
```
lpsinger:
```suggestion
      if (feature('APP_CLIENT_TRACKING'))  {
        expiredCreds.push(...moreExpiredCreds)
        warningCreds.push(...moreWarningCreds)
      } else {
        // Put both expired and warning together for the first few iterations of the scheduled task
        warningCreds.push(...moreWarningCreds, ...moreExpiredCreds)
      }
```

jracusin:
Yes, we should merge and deploy on Tuesday September 9, so it will then start sending out warning emails. I think we can them remove the feature flag on delete on or around Sept 18.

PR #3311, opened 2025-09-05
Archive and update MOA mission page

jracusin:
# Description

GCN Classic MOA alerts haven't been issued since 2021. As far as I can tell, the experiment is no longer operational.  This PR makes updates to their mission page and moves it into the archive list.

# Related Issue(s)

# Testing


app/routes/missions.tsx
```diff
@@ -97,6 +94,9 @@ export default function () {
                     <NavLink key="integral" to="archive/integral">
                       INTEGRAL
                     </NavLink>,
+                    <NavLink key="moa" to="moa">
```
lpsinger:
```suggestion
                    <NavLink key="moa" to="archive/moa">
```

PR #3325, opened 2025-09-09
Unify interface for Circulars corrections and edits

lpsinger:
In the view route for individual Circulars, consolidate the buttons for requesting a correction and editing a Circular into a single button. If the user is a moderator, then they get the option to either submit a correction or save as a new version on the Circulars edit form itself.

This reduces clutter in the Circulars button bar and places the decision over what kind of change to make in a more logical place.

Screenshot 2025-09-09 at 15 51 36
Screenshot 2025-09-09 at 15 51 45


app/routes/docs.circulars.corrections/route.mdx
```diff
@@ -28,7 +28,7 @@ Anyone with a [peer endorsement](/docs/circulars/submitting#to-request-a-peer-en
      className="height-auto"
```
jracusin: This needs to be removed or replaced with an updated screenshot.

app/routes/circulars.edit.$circularId/route.tsx
```diff
@@ -47,5 +66,17 @@ export async function loader({
...
+      <p className="usa-paragraph">
+        See{' '}
+        <Link to="/docs/circulars/corrections">
+          documentation on Circulars moderation
+        </Link>{' '}
+        for more information on corrections.
+      </p>
+      <CircularEditForm {...data} />
+    </>
+  )
```
jracusin, line 81: Is this above the edit box?  I don't see it in your screenshots.

lpsinger: Yes:

Screenshot 2025-09-09 at 16 01 15

app/routes/docs.circulars.corrections/route.mdx
lpsinger: Done:

Screenshot 2025-09-09 at 16 29 46

PR #3337, opened 2025-09-11
Always show Circulars edit button

lpsinger:
Always show the Circulars edit button, even when the user is not logged in. When the user navigates to either the new Circular page or the edit page, if they are not signed in then they are presented with the same dialog box to invite them to sign in and get a peer endorsement.

https://github.com/user-attachments/assets/63f387c4-8454-4393-8aff-c3529e14a921


lpsinger:
> We could change the text to "In order to submit or edit ..."

Done.

jracusin (approved):

PR #3349, opened 2025-09-18
Replace paginateQuery with paginateScan

dakota002:
Cloudwatch is throwing the following error:
```
"errorType": "ValidationException",
"errorMessage": "Either the KeyConditions or KeyConditionExpression parameter must be specified in the request.",
```
This is likely due to trying to run a paginateQuery without a key expression. This should fix the issue

### DIFF
--- a/app/scheduled/app-client-expire/index.ts
+++ b/app/scheduled/app-client-expire/index.ts
@@ -45,7 +45,7 @@ export async function handler() {
         'attribute_not_exists(expired) and (lastUsed < :warningCutoff or (attribute_not_exists(lastUsed) and created < :warningCutoff))',
       TableName,
       ExpressionAttributeValues: {
-        ':warningCutoff': { N: warningCutoff.toString() },
+        ':warningCutoff': warningCutoff,
       },
     }
   )


app/scheduled/app-client-expire/index.ts
```diff
@@ -46,7 +45,7 @@ export async function handler() {
         'attribute_not_exists(expired) and (lastUsed < :warningCutoff or (attribute_not_exists(lastUsed) and created < :warningCutoff))',
       TableName,
       ExpressionAttributeValues: {
-        ':warningCutoff': warningCutoff,
+        ':warningCutoff': { N: warningCutoff.toString() },
```
lpsinger: Are you sure that this is the correct form, that the library does not marshal JavaScript types to DynamoDB types automatically?

lpsinger (changes requested):

Courey (approved):

app/scheduled/app-client-expire/index.ts
lpsinger: I'm pretty sure that this should be:

```suggestion
        ':warningCutoff': warningCutoff,
```

dakota002: I think you're right. That is what I had at first, but there was an error, but I no longer get that error when I try this. It may have been just vscode bugging out

lpsinger (approved):

PR #3356, opened 2025-09-24
Fix breaking error when a account isn't found for a sub

dakota002:
Fixes multiple errors, including when a sub isn't found, bad dedent import, and an email isn't provided for the "To" address

### DIFF
--- a/app/scheduled/app-client-expire/index.ts
+++ b/app/scheduled/app-client-expire/index.ts
@@ -8,7 +8,7 @@
 import { tables } from '@architect/functions'
 import { type DynamoDBDocument, paginateScan } from '@aws-sdk/lib-dynamodb'
 import partition from 'lodash/partition'
-import dedent from 'ts-dedent'
+import { dedent } from 'ts-dedent'
 
 import { EXPIRATION_MILLIS, WARNING_MILLIS } from '~/lib/cognito'
 import {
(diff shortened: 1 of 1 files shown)


app/scheduled/app-client-expire/index.ts
```diff
@@ -71,33 +71,37 @@ export async function handler() {
...
-      ])
-    )
-  )
+  const entries = []
+  for (const sub of uniqueSubs) {
+    try {
+      const attributes = (await getCognitoUserFromSub(sub)).Attributes
+      entries.push([sub, extractAttributeRequired(attributes, 'email')])
+    } catch (e) {
+      console.log('Error, user may not exist: ', sub)
```
jracusin: How would a user not exist?

dakota002: I am not sure, but there was a case on dev where this is the case. Maybe an account that was deleted?

app/scheduled/app-client-expire/index.ts
```diff
@@ -71,33 +71,37 @@ export async function handler() {
...
-        ),
-      ])
-    )
-  )
+  const entries = []
+  for (const sub of uniqueSubs) {
+    try {
+      const attributes = (await getCognitoUserFromSub(sub)).Attributes
+      entries.push([sub, extractAttributeRequired(attributes, 'email')])
+    } catch (e) {
```
lpsinger: Never catch all errors indiscriminately. What errors are you expecting to occur here?

lpsinger: Your `for` loop will be slow because it will make all of the Cognito API calls serially, instead of asynchronously and in parallel. Please keep the original structure:

```ts
const userEmailMap: { [key: string]: string } = Object.fromEntries(
  await Promise.all(
    uniqueSubs.map(async (sub) => {
      let email
      try {
        const attributes = (await getCognitoUserFromSub(sub)).Attributes
        email = extractAttributeRequired(attributes, 'email')
      } catch (e) {
        /* appropriately handle the types of errors that you expect */
      }
      return [sub, email]
    })
  )
)
```

lpsinger (changes requested):
Please rebase.

app/scheduled/app-client-expire/index.ts
dakota002: The error is fairly generic, it usually looks like this:
```json
{
    "errorType": "Error",
    "errorMessage": "[object Response]",
    "stack": [
        "Error: [object Response]",
        "    at Object.intoError (file:///var/runtime/index.mjs:46:16)",
        "    at Object.textErrorLogger [as logError] (file:///var/runtime/index.mjs:699:56)",
        "    at postError (file:///var/runtime/index.mjs:823:27)",
        "    at done (file:///var/runtime/index.mjs:855:11)",
        "    at fail (file:///var/runtime/index.mjs:867:11)",
        "    at file:///var/runtime/index.mjs:894:20",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)"
    ]
}
```

It gets thrown if there is a sub that doesn't correspond to a user. 

This doesn't seem like it should be possible, but it is at least on dev

lpsinger: It's probably this error: https://github.com/nasa-gcn/gcn.nasa.gov/blob/9d5e8b22199199753fe5d34489277fb768a121b4/app/lib/cognito.server.ts#L87-L89

dakota002: Likely the case, is there an appropriate way that I should catch this? It doesn't like my current checks

lpsinger:
What _is_ the dedent error?

You can keep it as one PR, but that means that every time you update it, I will have to re-review all of it.

app/scheduled/app-client-expire/index.ts
lpsinger: Yes. Make sure that the `try` block encloses _only_ the call to `getCognitoUserFromSub`. Check that the error is an instance of `Response`. If it is not, then re-throw the error.

lpsinger:
> > What _is_ the dedent error?
> > You can keep it as one PR, but that means that every time you update it, I will have to re-review all of it.
> 
> Thats fine, it was an import error. It was imported incorrectly, without the curly brackets

It does make more work for the reviewer. In the future, please open separate PRs to fix separate bugs, even when there are multiple bugs in the same file.

dakota002:
Okay, tested and working now on dev

app/scheduled/app-client-expire/index.ts
```diff
@@ -75,23 +75,27 @@ export async function handler() {
...
           }
         }
-        return [sub, email]
+        if (user) {
+          const attributes = user.Attributes
+          const email = extractAttributeRequired(attributes, 'email')
+          return [sub, email]
+        } else {
+          return [sub, undefined]
+        }
```
lpsinger:
```suggestion
        return [sub, extractAttribute(user?.Attributes, 'email')]
```

app/scheduled/app-client-expire/index.ts
```diff
@@ -75,23 +75,27 @@ export async function handler() {
...
-        let email
+        let user
         try {
-          const attributes = (await getCognitoUserFromSub(sub)).Attributes
-          email = extractAttributeRequired(attributes, 'email')
+          user = await getCognitoUserFromSub(sub)
         } catch (e) {
-          if (e instanceof Error && e.message == '[object Response]') {
+          if (e instanceof Response) {
             console.log('Error, user may not exist: ', sub)
```
lpsinger:
```suggestion
            console.error('user does not exist: ', sub)
```

app/scheduled/app-client-expire/index.ts
dakota002: I'll add this in another commit, needs import update too

PR #3358, opened 2025-09-25
Set canonical URL for Circulars index

lpsinger:


tylerbarna (approved):

dakota002:
@lpsinger  This looks good, but I have been noticing in vscode that the `json` import from `remix-run/node` is deprecated. Is there something else we should be switching to? The messaging in the deprecation warning is not very descriptive:
image

PR #3393, opened 2025-10-15
Replace backslashed to fix path on Windows

dakota002:
This resolves an issue on Windows where the path joining causes the Schema Browser to error out locally.

### DIFF
--- a/app/lib/schema-data.server.ts
+++ b/app/lib/schema-data.server.ts
@@ -8,8 +8,7 @@
 import { RequestError } from '@octokit/request-error'
 import { Octokit } from '@octokit/rest'
 import memoizee from 'memoizee'
-import { basename, dirname, extname } from 'path'
-import { join, relative } from 'path/posix'
+import { basename, dirname, extname, join, relative } from 'path/posix'
 
 import { getEnvOrDieInProduction } from './env.server'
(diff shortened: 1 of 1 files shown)


app/lib/schema-data.server.ts
```diff
@@ -8,8 +8,8 @@
 import { RequestError } from '@octokit/request-error'
 import { Octokit } from '@octokit/rest'
 import memoizee from 'memoizee'
-import { basename, dirname, extname, join } from 'path'
-import { relative } from 'path/posix'
+import { basename, dirname, extname } from 'path'
```
lpsinger: Should we import all three of these other methods from `path/posix` as well?

dakota002: Maybe, I haven't had any other issues related to those functions, so I don't really think that changing that would make much of a difference. But it might just be cleaner

lpsinger: Would you please try it?

dakota002: yup

dakota002: Yea, they all work, updated to use them instead

lpsinger (approved):

PR #3424, opened 2025-11-21
Split the list user function to fix timeout errors

dakota002:
This fixes a timeout issue on the user admin page where the `listUsers` function times out on prod. 

Adding the filter directly to the `paginateListUsers` function improves the time, but it does not support any level of complexity other than "equal to" (=) or "starts with" (^=) so it has to run twice if we want to check both the name and email fields for a user

### DIFF
--- a/app/lib/cognito.server.ts
+++ b/app/lib/cognito.server.ts
@@ -103,9 +103,8 @@ export async function listUsers(
     name: extractAttribute(user.Attributes, 'name'),
     affiliation: extractAttribute(user.Attributes, 'custom:affiliation'),
   }))
-  return results.filter(
-    (item, index, self) => index === self.findIndex((t) => t.sub === item.sub)
-  )
+  const userMap = new Map(results.map((user) => [user.name, user])).values()
+  return Array.from(userMap)
 }
 
 async function searchForUsersByKey(


app/lib/cognito.server.ts
```diff
@@ -91,41 +91,52 @@ export async function getCognitoUserFromSub(sub: string) {
...
+  const results = []
+  for await (const page of email_pages) {
+    const nextUsers = page.Users
+    if (nextUsers)
+      results.push(
+        ...nextUsers.filter((user) =>
+          Boolean(extractAttribute(user.Attributes, 'email'))
+        )
+      )
+  }
```
lpsinger: There are two almost identical `for` loops. Refactor to avoid duplicated code.

app/lib/cognito.server.ts
```diff
@@ -91,41 +91,52 @@ export async function getCognitoUserFromSub(sub: string) {
...
+  return results
+    .map((user) => ({
+      sub: extractAttributeRequired(user.Attributes, 'sub'),
+      email: extractAttributeRequired(user.Attributes, 'email'),
+      name: extractAttribute(user.Attributes, 'name'),
+      affiliation: extractAttribute(user.Attributes, 'custom:affiliation'),
+    }))
+    .filter(
+      (item, index, self) => index === self.findIndex((t) => t.sub === item.sub)
+    )
```
lpsinger: This method of removing duplicates is O(N^2). Use an O(N log N) method instead (for example, using [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)).

dakota002: Is the method you are referring to the `[...new Set(thisList)]` way?  I had that originally and it wasn't removing the duplicates, but I can try it again, maybe something just didn't rebuild correctly when I tried the change

lpsinger: That's one way, yes.

dakota002: Is there an alternative? ~~I'd use [union](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union) but using this function gives me the "Try changing the 'lib' compiler option to 'esnext' or later.ts(2550)" message~~

Edit: Tried union and it doesn't work as expected

dakota002: @lpsinger what method do you have in mind? `[...new Set(thisList)]` doesn't reduce the array as expected

lpsinger: It doesn't work as you expect because [`Set` uses equality by reference, not by value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#value_equality):

```
$ node
Welcome to Node.js v24.11.1.
Type ".help" for more information.
> let s = new Set();
undefined
> s.add({a: 42})
Set(1) { { a: 42 } }
> s.add({a: 42})
Set(2) { { a: 42 }, { a: 42 } }
```

lpsinger: I would expect that the `Map` and `Object` constructors are O(N log N), and so you could do it like this:

```
> const users = [{name: 'Alice', flavor: 'chocolate'}, {name: 'Bob', flavor: 'strawberry'}, {name: 'Alice', flavor: 'chocolate'}]
undefined
> new Map(users.map((user) => [user.name, user])).values()
[Map Iterator] {
  { name: 'Alice', flavor: 'chocolate' },
  { name: 'Bob', flavor: 'strawberry' }
}
```

app/lib/cognito.server.ts
```diff
@@ -91,41 +91,45 @@ export async function getCognitoUserFromSub(sub: string) {
...
+    ...(await searchForUsersByKey(filterString, 'name')),
+  ].map((user) => ({
+    sub: extractAttributeRequired(user.Attributes, 'sub'),
+    email: extractAttributeRequired(user.Attributes, 'email'),
+    name: extractAttribute(user.Attributes, 'name'),
+    affiliation: extractAttribute(user.Attributes, 'custom:affiliation'),
+  }))
+  return results.filter(
+    (item, index, self) => index === self.findIndex((t) => t.sub === item.sub)
+  )
```
lpsinger: Still O(n^2). Please use an O(N log N) algorithm.

lpsinger (approved):

PR #3198, opened 2025-07-29
Add eventId autofill on circular edit form

Courey:
# Description
Add eventId autofill on circular edit form
# Related Issue(s)
resolves #3183

PR #3433, opened 2025-12-01
Fix bug with automatic subject-eventId linking

### LINKED ISSUE #3427: Circulars EventId checkbox is not correcting the EventId
# Current behavior
When one edits a Circular that includes an update to the event name in the subject and leaves the checkbox checked, it shows the new name, but says "The eventID does not match".  Then when the edit request is received, it shows the old EventId.

# Expected behavior
The checkbox should show the updated EventId and implement it in the correction.  Showing the EventId matching and the message not matching is also very confused and should be revised.

# Steps to reproduce
I reproduced this on test by editing a Circular, adding a letter or changing the letter in the GRB name.  The eventID field appeared to change, but with the warning message above. When I looked at the moderator interface, the EventId remained as the original one.

# Environment
N/A

Courey:
# Description
This corrects an issue where it looks like the eventId was being updated in the circular edit form and it wasn't.

# Related Issue(s)
This was introduced when we made these changes [PR 3](https://github.com/Courey/gcn.nasa.gov/pull/3/files) I believe at [this line](https://github.com/Courey/gcn.nasa.gov/pull/3/files#diff-be6bb4a0f73274ad470d7b05e6594f345b3578ddff27702fddba5465fe34b20dR289) which was merged in in this [PR 3198](https://github.com/nasa-gcn/gcn.nasa.gov/pull/3198)
Resolves #3427 

# Testing
Tested locally.
- changed subject with eventId linked (in edit and request correction)
- unlinked them and changed the subject (in edit and request correction)
- re-linked them and submitted (in edit and request correction)
- unlinked them and changed the eventId (in edit and request correction)
- re-linked them and submitted (in edit and request correction)

### DIFF
--- a/app/routes/circulars.new.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.new.$circularId/CircularEditForm.tsx
@@ -234,8 +234,8 @@ export function CircularEditForm({
                 <InputPrefix className="wide-input-prefix">
                   Event ID
                 </InputPrefix>
-                <span className="padding-1">{defaultEventId}</span>
-                <input type="hidden" name="eventId" value={defaultEventId} />
+                <span className="padding-1">{derivedEventId}</span>
+                <input type="hidden" name="eventId" value={derivedEventId} />
               </InputGroup>
             ) : (
               <InputGroup className="maxw-full">
@@ -261,7 +261,8 @@ export function CircularEditForm({
               label={
                 <>
                   Automatically fill event ID from subject
-                  {defaultEventId !== derivedEventId &&
+                  {!linkEventId &&
+                    eventId !== derivedEventId &&
                     '. The event ID does not match.'}
                 </>
               }


app/routes/circulars.new.$circularId/CircularEditForm.tsx
```diff
@@ -134,13 +134,11 @@ export function CircularEditForm({
   const [submitter, setSubmitter] = useState(defaultSubmitter)
   const subjectValid = subjectIsValid(subject)
   const derivedEventId = parseEventFromSubject(subject)
-  const [linkEventId, setLinkEventId] = useState(
-    originalEventId === derivedEventId
+  const [linkEventId, setLinkEventId] = useState(true)
```
lpsinger: This default value is not correct: if, upon navigating to the form, the event ID derived from the subject does not match the actual event ID, then the default should be false.

Courey: I set it to false.

lpsinger: That's not correct either; the correct initial value when the page loads depends on both the event ID derived from the subject and the actual event ID.

app/routes/circulars.new.$circularId/CircularEditForm.tsx
```diff
@@ -137,10 +137,10 @@ export function CircularEditForm({
   const [linkEventId, setLinkEventId] = useState(
     originalEventId === derivedEventId
   )
-  const [defaultEventId, setDefaultEventId] = useState(
-    originalEventId || derivedEventId
+  const [eventId, setEventId] = useState(
+    linkEventId ? derivedEventId : originalEventId
```
lpsinger, line 141: I think that this change may not be necessary.

PR #3462, opened 2026-01-13
Users migration

dakota002:
- Add 3 new tables and seed data examples for managing Users:
  - `users`: where minor metadata is maintained
  - `teams`: future replacement for group membership
  - `team_members`: link between previous two
- Add `user.server.ts` and `teams.server.ts` to do DynamoDB based interactions with the new `users` and `teams` tables
- Update client credentials server to add new functions that check team membership before creating new app clients
- Update the New Credential Form behind a feature flag to use the new functions 
- Update the users api behind feature flag to query DynamoDB by name and email instead of checking cognito
- Update the user page to update their rows in DynamoDB

### DIFF
--- a/app.arc
+++ b/app.arc
@@ -104,6 +104,19 @@ legacy_users
   email *String
   PointInTimeRecovery true
 
+users
+  sub *String
+  PointInTimeRecovery true
+
+teams
+  teamId *String
+  PointInTimeRecovery true
+
+team_members
+  sub *String
+  teamId **String
+  PointInTimeRecovery true
+
 @tables-indexes
 email_notification_subscription
   topic *String
@@ -160,6 +173,18 @@ synonyms
 client_credentials
   client_id *String
   name credentialsByClientId
+
+users
+  username *String
+  name usersByUserName
+
+users
+  email *String
+  name usersByEmail
+
+team_members
+  teamId *String
+  name usersByTeam
 
 @aws
 runtime nodejs24.x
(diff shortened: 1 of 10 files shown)


app.arc
```diff
@@ -104,6 +104,14 @@ legacy_users
   email *String
   PointInTimeRecovery true
 
+user_data
```
lpsinger:
```suggestion
users
```

(Any table holds data, so 'data' is not meaningful in a table name)

app.arc
```diff
@@ -104,6 +104,14 @@ legacy_users
...
+  sub *String
+  PointInTimeRecovery true
+
+kafka_teams
```
lpsinger:
```suggestion
teams
```

Teams might include non-Kafka roles like Circulars submitters and moderators.

lpsinger (changes requested):
Would you please add to this PR the business logic code changes to use DynamoDB instead of Cognito to store group/team membership and user name/email/affiliation?

app.arc
```diff
@@ -104,6 +104,19 @@ legacy_users
...
 
+users
+  sub *String
+  PointInTimeRecovery true
+
+teams
+  teamId *String
+  PointInTimeRecovery true
+
+user_permissions
```
lpsinger: How about `teamMembers`?

lpsinger (changes requested):
This table design looks fine but I wanted to see in this PR it replacing Cognito for group management and user metadata storage.

app.arc
dakota002: I like it, much more clear

dakota002:
Splitting out into smaller PRs, starting with #3574

Issue #2941, opened 2025-03-14
IceCube Gold/Bronze Track Alerts


jracusin:
# Description
IceCube is adding a new notice topic for single neutrinos (Gold, Bronze Track).  There are schema updates including a new base neutrino schema.

# Acceptance criteria

- [x] create [topics/ACLs](https://gcn.nasa.gov/docs/notices/producers) (GCN Team)
- [x] finalize [JSON schema](https://gcn.nasa.gov/docs/notices/schema) (Producers)
- [x] add/update [producer mission page](https://gcn.nasa.gov/missions) (Producers)
- [x] update [mission page table](https://gcn.nasa.gov/missions) (GCN Team)
- [x] add to [quickstart](https://gcn.nasa.gov/quickstart) (GCN Team)
- [x] remove [feature flag](https://gcn.nasa.gov/docs/contributing/feature-flags) (GCN Team)
- [x] post and distribute [announcement](https://gcn.nasa.gov/news) (GCN Team)

# Mission contact people
@blaufuss @azegarelli

jracusin:
JSON schema in https://github.com/nasa-gcn/gcn-schema/pull/236

dakota002:
@jracusin should this get added behind a feature flag?

PR #3487, opened 2026-01-30
Correct new IceCube topic name

jracusin:
# Description
This PR fixes an inconsistency in the new IceCube topic name between the new mission page documentation (still behind a feature flag) and the schema.

# Related Issue(s)
Related to #2941

# Testing
local dev


jracusin:
We have clarified that the [example filename](https://github.com/nasa-gcn/gcn-schema/pull/333) needs to change not the topic name in the documentation.

PR #3499, opened 2026-02-25
Fix userGroupString

dakota002:
Resolves #2268 

Use correct input for `listGroupsForUser`, correct use in admin page

### DIFF
--- a/app/routes/admin.users.$userId.tsx
+++ b/app/routes/admin.users.$userId.tsx
@@ -42,7 +42,8 @@ export async function loader({
     throw new Response(null, { status: 403 })
   if (!userId) throw new Response(null, { status: 404 })
   const user = await getCognitoUserFromSub(userId)
-  const userGroups = (await listGroupsForUser(userId)).map(
+  if (!user || !user.Username) throw new Response(null, { status: 404 })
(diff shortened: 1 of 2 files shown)


app/routes/admin.users.$userId.tsx
```diff
@@ -41,7 +41,8 @@ export async function loader({
   if (!currentUser?.groups.includes(adminGroup))
     throw new Response(null, { status: 403 })
   if (!userId) throw new Response(null, { status: 404 })
-  const user = await getCognitoUserFromSub(userId)
+  const users = await listUsers(userId)
```
lpsinger: This function does not search for users by sub. Try this instead:

```suggestion
  const user = await getCognitoUserFromSub(sub)
```

dakota002: I'm pretty sure we _don't_ want to get user by sub, we want to use their user ID, no?

This loader function expects a cognito id, not a sub

lpsinger: > This loader function expects a cognito id, not a sub

Are you sure? The URL params are subs, not Cognito usernames.

dakota002: You were right, i had to double check what was going on but yes, it should be resolved now

PR #3534, opened 2026-03-10
Alter back buttons in circulars to use react hook instead of hard-coded links

tylerbarna:
# Description

There has been a known inconsistency in the on-screen navigation buttons and whether or not they preserve the user's search parameters in the archive. This was partially addressed in solving https://github.com/nasa-gcn/gcn.nasa.gov/pull/2391, but we have continued to have some cases where the link to the previous page does not include the relevant search parameters. However, as @dakota002 noted, there is a [react hook](https://reactnavigation.org/docs/navigating/) to initiate a return to the previous page, which simplifies this button. This PR replaces the hard-coded link with the react hook.

# Testing

1. Navigate to the circulars archive
2. Initiate a search
3. Select any result 
4. Press the on-screen back button
5. The page should still show the search being executed.
6. Repeat the above steps in Event view
7. Repeat steps 1 and 2, but select the button to submit a new circular. Pressing the back button at the bottom of the page should have the same behavior as step 5.

### DIFF
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -98,9 +98,6 @@ export default function () {
         <Link
           onClick={(event) => {
             event.preventDefault()
-            console.log(
-              'Navigating back to previous page with javaScript navigation'
-            )
             navigate(-1)
           }}
           className="usa-button"
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -101,7 +101,7 @@ export default function () {
             navigate(-1)
           }}
           className="usa-button"
-          to={`/circulars/${searchString}`}
+          to={`/circulars${searchString}`}
         >
           <Icon.ArrowBack role="presentation" className="margin-y-neg-2px" />
           Back
--- a/app/routes/circulars.$circularId.($version)/route.tsx
+++ b/app/routes/circulars.$circularId.($version)/route.tsx
@@ -7,7 +7,12 @@
  */
 import type { HeadersFunction, LoaderFunctionArgs } from '@remix-run/node'
 import { json } from '@remix-run/node'
-import { Link, useLoaderData, useRouteLoaderData } from '@remix-run/react'
+import {
+  Link,
+  useLoaderData,
+  useNavigate,
+  useRouteLoaderData,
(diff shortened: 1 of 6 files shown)


app/routes/circulars.$circularId.($version)/route.tsx
```diff
@@ -85,11 +90,19 @@ export default function () {
...
     <>
       <ToolbarButtonGroup className="flex-wrap">
-        <Link to={`/circulars${searchString}`} className="usa-button">
+        <Link
+          onClick={(event) => {
+            event.preventDefault()
+            navigate(-1)
+          }}
+          className="usa-button"
+          to={`/circulars/${searchString}`}
```
tylerbarna:
```suggestion
          to={`/circulars${searchString}`}
```

lpsinger (commented):
This will only work as expected if the user navigates to a given Circular. If the user has pasted a URL into the search bar, or followed a link from another site, then the last page visited will not be the search results index.

tylerbarna:
> This will only work as expected if the user navigates to a given Circular. If the user has pasted a URL into the search bar, or followed a link from another site, then the last page visited will not be the search results index.

Do you have any ideas as to how that edge case might be obviated?

lpsinger:
No. I don't think it's an edge case. I think it indicates that this method is unsuitable.

app/routes/circulars.$circularId.($version)/route.tsx
```diff
@@ -85,11 +86,17 @@ export default function () {
   const linkString = `/circulars/${circularId}${
     !latest && version ? `/${version}` : ''
   }`
+  // const navigate = useNavigate()
```
tylerbarna:
```suggestion
```

lpsinger (commented):
Hmm, interesting. This might work.

lpsinger (changes requested):
In practice, this does not work (at least on Safari). The problem, I think, is that this check is always false, because none of our locations have a location key value of `default`:

https://github.com/evelynhathaway/use-back/blob/main/src/use-back.ts#L16

tylerbarna:
> In practice, this does not work (at least on Safari). The problem, I think, is that this check is always false, because none of our locations have a location key value of `default`:
> 
> https://github.com/evelynhathaway/use-back/blob/main/src/use-back.ts#L16

Interesting, it was working for me with chrome. Will investigate further.

PR #3535, opened 2026-03-10
Move new circular button to top of page

tylerbarna:
# Description

As per sprint planning discussion, several team members agreed that the "new circular" button wasn't particularly useful to have as part of the search bar and created some level of confusion. This PR moves the new circular button to the top of the page but does not change the behavior otherwise.

# Testing

1. Navigate to Circulars archive
2. The new circulars page should function the same, just in a different place


lpsinger:
I think it looks better and makes more sense as is.

PR #3568, opened 2026-04-21
Adds moderator check to synonym moderation action

Courey:
# Description
I wanted to add a moderator check to this action to ensure the actions are only available with the proper permissions.

# Related Issue(s)
None

# Testing
curled local endpoint to check that a 200 is not returned if user doesn't have appropriate permissions and throws a 403.

Locally a 500 is returned instead due to a bug in the 403 error handling. That will be patched following this PR.

the line that throws the 500:
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/root.tsx#L214
it does this because origin is undefined when hit from curl.
 
in the error for unauthorized:
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/root.tsx#L308
 
that calls useUrl, which fails on the invariant line since the origin is null
https://github.com/nasa-gcn/gcn.nasa.gov/blob/main/app/root.tsx#L212
Screenshot 2026-04-21 at 10 56 46 AM
mod_perms

### DIFF
--- a/app/routes/synonyms.$synonymId.tsx
+++ b/app/routes/synonyms.$synonymId.tsx
@@ -82,7 +82,7 @@ export async function action({
 
   const user = await getUser(request)
 
-  if (!user || !user.groups.includes(moderatorGroup))
+  if (!user?.groups.includes(moderatorGroup))
     throw new Response(null, { status: 403 })
 
   const data = await request.formData()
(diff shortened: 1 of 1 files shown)


app/routes/synonyms.$synonymId.tsx
```diff
@@ -79,6 +79,12 @@ export async function action({
   params: { synonymId },
 }: ActionFunctionArgs) {
   invariant(synonymId)
+
+  const user = await getUser(request)
+
+  if (!user || !user.groups.includes(moderatorGroup))
```
lpsinger:
```suggestion
  if (!user?.groups.includes(moderatorGroup))
```

lpsinger (changes requested):

lpsinger (approved):

PR #3571, opened 2026-04-22
Add userpool client resource for producer credential

dakota002:
@lpsinger Should this update the environment variable, or should the circulars producer code be updated to retrieve the clientid/secret from cognito on demand?

### DIFF
--- a/src/plugins/circulars-producer-client.js
+++ b/src/plugins/circulars-producer-client.js
@@ -1,0 +1,22 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const deploy = {
+  start({ cloudformation }) {
+    cloudformation.Resources.GcnCircularsProducerUserPoolClient = {
+      Type: 'AWS::Cognito::UserPoolClient',
+      Properties: {
+        AllowedOAuthFlows: ['client_credentials'],
+        AllowedOAuthFlowsUserPoolClient: true,
+        AllowedOAuthScopes: ['gcn.nasa.gov/kafka-circulars-producer'],
+        GenerateSecret: true,
+        UserPoolId: process.env.COGNITO_USER_POOL_ID,
+      },
+    }
+  },
+}
--- a/src/plugins/circulars-producer-client.js
+++ b/src/plugins/circulars-producer-client.js
@@ -1,0 +1,31 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const deploy = {
+  start({ cloudformation }) {
+    const user_pool_id =
+      cloudformation.Resources.AnyCatchallHTTPLambda.Properties.Environment
+        .Variables.COGNITO_USER_POOL_ID
+    if (!user_pool_id)
+      throw new Error('Environment variable COGNITO_USER_POOL_ID must be set')
+    cloudformation.Resources.GcnCircularsProducerUserPoolClient = {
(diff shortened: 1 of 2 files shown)


dakota002:
Resolves https://github.com/nasa-gcn/gcn-infrastructure/issues/107

lpsinger (changes requested):
This is incomplete. Please add CloudFormation and Architect plugin code to provide the client ID and secret to the Lambda.

dakota002:
> This is incomplete. Please add CloudFormation and Architect plugin code to provide the client ID and secret to the Lambda.

This relates to my question at the top. By "provide the client ID and secret to the lambda" I assume you mean to set the environment variable?

lpsinger:
> @lpsinger Should this update the environment variable, or should the circulars producer code be updated to retrieve the clientid/secret from cognito on demand?

It should update the environment variable.

You are going to encounter https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1855 and may need to write a custom CloudFormation resource.

dakota002:
Yea, I see our custom plugin in the infrastructure repo and how we do it there. Should I just do the same here?

lpsinger:
> Yea, I see our custom plugin in the infrastructure repo and how we do it there. Should I just do the same here?

It would be better to do something that is reusable across both CloudFormation templates. See https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-types.html. Is developing and publishing a CloudFormation extension feasible?

dakota002:
Have we tested that CloudFormation doesn't return the values as that issue suggests? I would think that the correct way would be to use something like `!GetAtt MyAppClient.[ClientId | ClientSecret]` 
The properties are at least defined in their own schema https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1#/registry/public-extensions/details/schema?arn=arn%3Aaws%3Acloudformation%3Aus-east-1%3A%3Atype%2Fresource%2FAWS-Cognito-UserPoolClient

lpsinger:
I have not tried it recently. That's a good idea.

dakota002:
I guess my latest commit is what I would expect the usage to be, using `GetAtt`

lpsinger:
> I guess my latest commit is what I would expect the usage to be, using `GetAtt`

Yes, looks great. Does it work?

dakota002:
I can try deploying it to dev and let you know, its currently untested other than locally

lpsinger:
> I can try deploying it to dev and let you know, its currently untested other than locally

Instead, can you create a small toy CloudFormation template that has an app client resource and an Output that comes from the client secret? Then you can attempt to deploy that standalone stack through the AWS console to test the approach.

dakota002:
I made a test on dev, check out DemoClientStack, it looks right in the Outputs to me

src/plugins/circulars-producer-client.js
```diff
@@ -0,0 +1,26 @@
...
+export const deploy = {
+  start({ cloudformation }) {
+    cloudformation.Resources.GcnCircularsProducerUserPoolClient = {
+      Type: 'AWS::Cognito::UserPoolClient',
+      Properties: {
+        AllowedOAuthFlows: ['client_credentials'],
+        AllowedOAuthFlowsUserPoolClient: true,
+        AllowedOAuthScopes: ['gcn.nasa.gov/kafka-circulars-producer'],
+        GenerateSecret: true,
+        UserPoolId: process.env.COGNITO_USER_POOL_ID,
```
lpsinger: Although this is an environment variable within the context of the Lambda, it's probably not within the environment that the Architect CLI runs. Do we have another example of a CloudFormation resource that depends on the value of an environment variable? You might need to have a CloudFormation dependency on the corresponding SSM parameter.

lpsinger (changes requested):

src/plugins/circulars-producer-client.js
dakota002: Ah yea, lambda cognito permissions, I can copy how that does it

lpsinger (approved):

lpsinger:
Please squash your changes when you merge.

PR #3576, opened 2026-04-28
Fixes 403 handling when there is no origin

### LINKED ISSUE #3575: fix 500 error in 403 handling when origin is missing
# Current behavior
When you hit a 403 error in the synonyms moderation routes (edit, delete), it gathers data to create the error page. The origin is used in the error page. If the user has curled the endpoint and there is no error page, the error handling of the 403 throws a 500 because origin is undefined and invariant can't handle that.

# Expected behavior
even if you curl so there is no origin, it should pass a 403

# Steps to reproduce
locally have your instance running and then curl:
`curl -s -o /dev/null -w  -X POST "http://localhost:3333/synonyms/958b3345-60fa-4d9a-8190-0d5c7175ff52" -H "Content-Type: application/x-www-form-urlencoded" -d "intent=delete"`

it throws a 500 because of an error with invariant since origin is undefined

# Environment
any/all

Courey:
# Description
When curling routes, when unauthorized and there is no origin as the endpoint was curled, it was throwing a 500 error within the 403 error handling as it was trying to use the origin in the unauthorized page.

# Related Issue(s)
closes #3575

# Testing
I defaulted the origin to our main production domain so that it returns a 403.

Screenshot 2026-04-28 at 11 14 32 AM

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -210,9 +210,10 @@ export function useOrigin() {
 }
 
 export function useUrl() {
-  const origin = useOrigin()
-  invariant(origin)
+  const origin = useOrigin() || 'https://gcn.nasa.gov'
   const { pathname, search, hash } = useLocation()
+
+  invariant(origin)
   const url = new URL(origin)
   url.pathname = pathname
   url.search = search


lpsinger (changes requested):
This will be incorrect for the dev and test stages.

As far as I can recall, this custom 403 page is only supposed to be rendered for view routes, not pure resource routes. The problem is that it shouldn't be rendering this view at all.

Courey:
Since this issue only happens when you try to curl a view so origin is undefined, we can just leave it as it's okay because they've accessed the page inappropriately.

On the UI you still get the correct response.
steps to confirm:
1. log in
2. open a second gcn tab
3. navigate to a valid synonym in the UI
4. log out with the other tab
5. refresh the original tab that had the valid synonym loaded and it should show you the 403 error page with a link to log in.

Image

PR #3577, opened 2026-04-28
Teams Update: Add teams and topics related DB functions

dakota002:
# Description

- Makes `Permission` a type to be more reusable
- Adds new type and table: `TeamInvite`,`team_invites` for tracking status of users to be added to teams
- Adds functions for:
  - Create/Retrieve/Update/Delete teams 
  - Create/Update/Delete Topics
  - Inviting users to join team
  - Managing invite status for users
  - Managing team permission level

# Related Issue(s)
This is a follow-up to https://github.com/nasa-gcn/gcn.nasa.gov/pull/3574

### DIFF
--- a/app/lib/teams.server.ts
+++ b/app/lib/teams.server.ts
@@ -5,12 +5,23 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+import { tables } from '@architect/functions'
+import crypto from 'crypto'
+import { dedent } from 'ts-dedent'
+
+import { sendEmail } from './email.server'
+import { origin } from './env.server'
+import type { User } from '~/routes/_auth/user.server'
+
+const fromName = 'GCN Teams'
 
 export type Team = {
   teamId: string
   teamName: string
   description: String
 }
+
+export type Permission = 'admin' | 'write' | 'read'
 
 /**
  * Maps a User to a Team and their respective permission level
@@ -27,7 +38,14 @@ export type TeamMember = {
   sub: string
   teamId: string
   topicId: string
-  permission: 'admin' | 'write' | 'read'
+  permission: Permission
+}
+
+export type TeamInvite = {
+  teamId: string
+  email: string
+  topicId: string
+  permission: Permission
 }
 
 export type Topic = {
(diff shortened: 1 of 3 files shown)


app/lib/teams.server.ts
```diff
@@ -36,3 +54,308 @@ export type Topic = {
...
+ *
+ */
+export async function createTeam(
+  user: User,
+  teamName: string,
+  description: string,
+  pocEmail: string,
+  topicName: string
+) {
+  if (!user.groups.includes('gcn.nasa.gov/gcn-admin'))
```
lpsinger: I thought that Teams would replace Cogito groups.

app/lib/teams.server.ts
```diff
@@ -36,3 +54,308 @@ export type Topic = {
...
+  })
+
+  await sendEmail({
+    to: [newUserEmail],
+    fromName,
+    subject: `GCN Teams Invite: ${team.teamName}`,
+    body: dedent`You have been invited to join ${team.teamName} by ${user.name}. 
+    To accept, go to ${origin}/teams and accept the invite. Once complete, you will be able to create client credentials to 
+    produce and/or consume Kafka messages, as determined by your team admin.`,
+  })
```
lpsinger: Wrap in `Promise.all([])`.

app/lib/teams.server.ts
```diff
@@ -27,7 +38,14 @@ export type TeamMember = {
   sub: string
   teamId: string
   topicId: string
-  permission: 'admin' | 'write' | 'read'
+  permission: Permission
+}
+
+export type TeamInvite = {
+  teamId: string
+  email: string
```
lpsinger: Not `sub`?

app/lib/teams.server.ts
```diff
@@ -36,3 +54,308 @@ export type Topic = {
...
+  })
+
+  await sendEmail({
+    fromName,
+    to: [pocEmail],
+    subject: 'GCN Team Admin Invite',
+    body: dedent`You have been added as a team admin to ${teamName}. 
+    
+    To continue, go to ${origin}/teams and accept the invite. Once complete, you will be able to invite other users to join your team.`,
+  })
```
lpsinger: Wrap in `Promise.all([])`.

app/lib/teams.server.ts
```diff
@@ -36,3 +54,308 @@ export type Topic = {
...
+      ExpressionAttributeValues: {
+        ':topicId': topicId,
+      },
+    })
+  ).Items
+  await Promise.all(
+    memberships.map((x) =>
+      db.team_members.delete({ sub: x.sub, teamId: x.teamId })
+    )
+  )
```
lpsinger: Use bulk operations instead of individual deletes.

lpsinger (changes requested):
High-level comments:

- Perform requests concurrently with `Promise.all([])` where possible.
- Use `sub`, not email address, as a primary key.
- Use DynamoDB bulk operations wherever possible.

app/lib/teams.server.ts
dakota002: Originally I thought that this would be useful if the user being invited didn't yet had an account. But I now think this isn't for the best. I will rework this

app/lib/teams.server.ts
dakota002: I wasn't sure if I should replace all groups, but I can make that so

dakota002: Actually are we doing ALL groups? If so, this would change how the `createTeam` function works, since it also expects a topic as part of the team configuration. 

Unless we would have a separate definition for non-Kafka related teams for gcn-admin (website admins, not kafka related admin permission), circular-moderator, circular-submitters

What do you think?

app/lib/teams.server.ts
```diff
@@ -27,12 +40,390 @@ export type TeamMember = {
...
+    permission,
+  })
+}
+
+export async function removeUserFromTeam(sub: string, teamId: string) {
+  const db = await tables()
+  await db.team_members.delete({ sub, teamId })
+}
+
+// #region Topic Functions
```
Courey, line 329: what is the region this is talking about?

Courey: ah, I see the end of the region, never mind!

app/lib/teams.server.ts
```diff
@@ -27,12 +40,390 @@ export type TeamMember = {
...
+  sub: string
+  topicId: string
+  permission: Permission
 }
 
 export type Topic = {
   topicId: string
   topicName: string
-  public: boolean
+  isPublic: boolean
```
Courey, line 56: I like this change. It is more consistent this way.

app/lib/teams.server.ts
```diff
@@ -27,12 +40,390 @@ export type TeamMember = {
...
+  const membership = (
+    await db.team_members.query({
+      IndexName: 'membersByTopicId',
+      KeyConditionExpression: 'topicId = :topicId',
+      FilterExpression: '#sub = :sub AND permission = :permission',
+      ExpressionAttributeNames: {
+        '#sub': 'sub',
+      },
+      ExpressionAttributeValues: {
+        ':sub': sub,
```
Courey, line 421: does it perform okay without an index on sub? I looked and didn't see one, but could have missed or it could be irrelevant.

Courey (approved):

app/lib/teams.server.ts
dakota002: Thanks, public is also a reserved word I believe

PR #3595, opened 2026-05-07
Update the endorsement server to be functional, not class based

dakota002:
Convert from class based to just exported functions in server file


lpsinger (approved):

dakota002:
@lpsinger Can I merge? I see some of the node builds are failing, but that seems like a timeout or something in the `npm install` step

lpsinger:
> @lpsinger Can I merge? I see some of the node builds are failing, but that seems like a timeout or something in the `npm install` step

Yes.

Issue #3517, opened 2026-03-09
Create backend for filtering Circulars by Event type


tylerbarna:
# Description
Add the required tools to determine the event type associated with a given circular and associate it with said circular

# Acceptance criteria

- [ ] Fulfill all sub-issues

Issue #3521, opened 2026-03-09
Update circulars archive search backend to filter by eventType


tylerbarna:
# Description
Similar to https://github.com/nasa-gcn/gcn.nasa.gov/issues/878, add a filter to the circulars to filter circulars based on the eventType parameter

# Acceptance criteria

- [ ] Create filter such that users can select one, several, or all eventTypes
- [ ] Only accept supported eventTypes; discard any values not included in canonical list
- [ ] Filter should show all circulars by default
- [ ] EventType filter should function similar to date filtering in that a filter can be applied in conjunction with a keyword search 
- [ ] (optional) Allow for users to make boolean-style filters (i.e. show circulars tagged with "GRB" AND "GW," Circulars tagged "GRB" OR "GW," and Circulars tagged with "GRB" but NOT "GW")

PR #3596, opened 2026-05-07
Add matchers for event types 

### LINKED ISSUE #3518: Add regex to associate a submitted circular with specific types of events
# Description
This is the first step of the Event type milestone to be completed, which involves establishing the event types to be used in a regular expression

# Acceptance criteria

- [ ] Define the list of desired Event types (e.g. GRBs, GWs, other, etc.) as well as the strings that should be associated with a given event type (i.e. "LIGO/Virgo/Kagra" and/or "GW" in the subject line should be associated with "Gravitational Wave")
- [ ] Create a regular expression that can parse the subject of a circular and determine the associated event types. For reference to an existing regex with a similar purpose, see https://github.com/nasa-gcn/gcn.nasa.gov/blob/47ff16281e4e09addc347c6dbb8b01c7d849a4b9/app/routes/circulars/circulars.lib.ts#L52 
- [ ] The regex should be able to return a list of associated event types if a given circular refers to multiple
- [ ] Create unit tests that validate all event types are correctly identified from example circulars (as well as a check to ensure it does not associate an event type with an incorrect circular)

tylerbarna:
# Description

This PR adds the regular expression matcher function that is required for #3517. There are a number of regular expressions for each of the event types. Generally, it written in such a way that subjects can be tagged with multiple event types, but the "Retraction" tag is special in that it supersedes all others. The "Misc" tag is not explicitly named but the function will return `undefined` if it matches none of the regular expressions. 

This PR also adds testing to validate that the function works with subjects from the archive.

# Related Issue(s)

Resolves #3518

# Testing

1. Review new code in circulars.lib.ts
2. Run tests in circulars.ts

### DIFF
--- a/app.arc
+++ b/app.arc
@@ -32,8 +32,13 @@ app-client-expire
   src build/scheduled/app-client-expire
 
 @tables-streams
-circulars
-  src build/table-streams/circulars
+circulars-kafka
+  table circulars
+  src build/table-streams/circulars-kafka
+
+circulars-email
+  table circulars
+  src build/table-streams/circulars-email
 
 synonyms
   src build/table-streams/synonyms
--- a/app/table-streams/circulars-email/index.ts
+++ b/app/table-streams/circulars-email/index.ts
@@ -10,12 +10,9 @@ import { errors } from '@opensearch-project/opensearch'
 import type { DynamoDBRecord } from 'aws-lambda'
 
 import { unmarshallTrigger } from '../utils'
-import { send as sendKafka } from '~/lib/kafka.server'
 import { createTriggerHandler } from '~/lib/lambdaTrigger.server'
 import type { Circular } from '~/routes/circulars/circulars.lib'
 import { send } from '~/routes/circulars/circulars.server'
-
-import { $id as circularsJsonSchemaId } from '@nasa-gcn/schema/gcn/circulars.schema.json'
 
 const index = 'circulars'
 
@@ -42,28 +39,14 @@ async function putIndex(circular: Circular) {
 export const handler = createTriggerHandler(
   async ({ eventName, dynamodb }: DynamoDBRecord) => {
     const id = unmarshallTrigger(dynamodb!.Keys).circularId as number
-    const promises = []
-
     if (eventName === 'REMOVE') {
-      promises.push(removeIndex(id))
+      await removeIndex(id)
     } /* (eventName === 'INSERT' || eventName === 'MODIFY') */ else {
       const circular = unmarshallTrigger(dynamodb!.NewImage) as Circular
-      const { sub, ...cleanedCircular } = circular
-      promises.push(
-        putIndex(circular),
-        sendKafka(
-          'gcn.circulars',
-          JSON.stringify({
-            $schema: circularsJsonSchemaId,
-            ...cleanedCircular,
-          })
-        )
-      )
+      await putIndex(circular)
       if (eventName === 'INSERT') {
-        promises.push(send(circular))
+        await send(circular)
       }
     }
-
-    await Promise.all(promises)
   }
 )
--- a/app/table-streams/circulars-kafka/index.ts
+++ b/app/table-streams/circulars-kafka/index.ts
@@ -1,0 +1,31 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { DynamoDBRecord } from 'aws-lambda'
+
+import { unmarshallTrigger } from '../utils'
+import { send } from '~/lib/kafka.server'
+import { createTriggerHandler } from '~/lib/lambdaTrigger.server'
+import type { Circular } from '~/routes/circulars/circulars.lib'
+
+import { $id as circularsJsonSchemaId } from '@nasa-gcn/schema/gcn/circulars.schema.json'
+
+export const handler = createTriggerHandler(
+  async ({ eventName, dynamodb }: DynamoDBRecord) => {
+    if (eventName === 'INSERT' || eventName === 'MODIFY') {
+      const circular = unmarshallTrigger(dynamodb!.NewImage) as Circular
+      const { sub, ...cleanedCircular } = circular
+      await send(
+        'gcn.circulars',
+        JSON.stringify({
+          $schema: circularsJsonSchemaId,
+          ...cleanedCircular,
+        })
+      )
+    }
+  }
+)


tylerbarna:
As I'm thinking about how the search filter would be implemented for #3521, I'm wondering if using undefined as the category equivalent for Misc/other makes it more complicated to implement in opensearch. The docs aren't particularly clear on if this would be an intended use of a null value https://docs.opensearch.org/latest/mappings/mapping-parameters/null-value, especially in a case where multiple event types are selected in addition to Misc/other (which would end up being passed as something like ["event1", "event2", undefined] (see also: https://docs.opensearch.org/latest/query-dsl/query-filter-context/). This gets even more complicated when considering how one might implement additional logic with this filter (ie options to select AND/OR when selecting multiple filters, the ability to exclude certain event types)

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bis\s+not\b/i,
+  ],
+  GRB: [
+    /\bGRB\d{6}[A-Z]?\b/i,
+    /\bGRBs?\b/i,
+    /\bgamma[-\s]?ray[-\s]?bursts?\b/i,
+    /\bFermi(?:\s?(?:GBM|LAT)|\d{9})?\b/i,
+    /\bSwift(?:[/-](?:BAT|XRT|UVOT))?\b/i,
+    /\bSVOM\b/i,
+    /\bINTEGRAL\b/i,
```
jracusin: These are all GRB detectors who detect more than GRBs

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bXRF\d{6}[A-Z]?\b/i,
+    /\bMASTER\b/i,
```
jracusin: optical follow-up telescope network

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bHAWC\b/i,
+    /\bHETE\b/i,
+    /\bKONUS\b/i,
+    /\bAstroSat\b/i,
```
jracusin: More gamma-ray detectors who detect more than GRBs

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bGW\d+\b/i,
+    /\bGWs?\b/i,
+    /\bgravitational[-\s]?waves?\b/i,
+    /\bLIGO\b/i,
+    /\bVirgo\b/i,
+    /\bKAGRA\b/i,
+    /\bS\d{6}[a-z]+\b/i,
+  ],
+  SGR: [/\bSGR\S*/i, /\bsoft[-\s]?gamma[-\s]?repeaters?\b/i],
+  FRB: [
```
jracusin: you could add CHIME & DSA-110

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bFRBs?\b/i,
+    /\bfast[-\s]?radio[-\s]?bursts?\b/i,
+  ],
+  SN: [/\bSN\d{4}[A-Za-z]*\b/i, /\bSNe?\b/i, /\bsuper[-\s]?novae?\b/i],
+  Neutrino: [
+    /\bneutrinos?\b/i,
+    /\bIceCube(?:-HAWC|-\d+)?\b/i,
+    /\bANTARES\b/i,
+    /\bKM3NeT\b/i,
+  ],
```
jracusin: Super-Kamiokande, SNEWS2

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+  'X-Ray Transient': [
+    /(?<!\S)EP(?=\s|:|$)/,
+    /\bEPW?[-\s]?\d{6,8}[A-Z]{0,2}\b/i,
+    /\bEP-XRT\b/i,
```
jracusin: Not a telescope

jracusin: EP has a WXT

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bfast[-\s]?radio[-\s]?bursts?\b/i,
+  ],
+  SN: [/\bSN\d{4}[A-Za-z]*\b/i, /\bSNe?\b/i, /\bsuper[-\s]?novae?\b/i],
+  Neutrino: [
+    /\bneutrinos?\b/i,
+    /\bIceCube(?:-HAWC|-\d+)?\b/i,
+    /\bANTARES\b/i,
+    /\bKM3NeT\b/i,
+  ],
+  'X-Ray Transient': [
```
jracusin: Chandra, XMM, NICER, NuSTAR, BlackCAT, others

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,95 @@ export function parseEventFromSubject(value: string) {
...
+    /\bEP-FXT\b/i,
+    /\bX[-\s]?ray(?:\s+transient)?\b/i,
+    /\bEinstein\s+Probe\b/i,
+    /\bMAXI\s?J\d{4}[+-]\d+/i,
+    /\bXRT\b/i,
+    /\bXRF\b/i,
+    /\bNuEM\b/i,
+  ],
+  Afterglow: [/\bafterglows?\b/i],
+  'Optical Transient': [
```
jracusin: Rubin/LSST, so many other optical telescopes. It's not going to be possible to make a comprehensive list.

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+    {
+      name: 'GRB pattern: \\bSwift(?:[/-](?:BAT|XRT|UVOT))?\\b',
+      subject:
+        'LIGO/Virgo S190930t: Swift J221951-484240 optical photometry of Chilescope observatory',
+      expected: ['GRB', 'GW', 'Optical Transient'],
+    },
+    {
+      name: 'GRB pattern: \\bSVOM\\b',
+      subject: 'EP260214b: SVOM/VT optical observation',
+      expected: ['GRB', 'X-Ray Transient', 'Optical Transient'],
```
jracusin: SVOM/VT is not one of SVOM's X-ray telescopes

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'GRB pattern: \\bINTEGRAL\\b',
+      subject: 'INTEGRAL observations of the events in the GWTC-1 catalog',
```
jracusin: GWTC is a GW catalog

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+      expected: ['GRB'],
+    },
+    {
+      name: 'GRB pattern: \\bXRF\\d{6}[A-Z]?\\b',
+      subject: 'XRF040812: possible afterglow second Chandra observation',
```
jracusin: An XRF (X-ray Flash) is also an X-ray transient

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'GRB pattern: \\bMASTER\\b',
+      subject: 'GRB 181202A: Global MASTER-Net optical observations',
+      expected: ['GRB', 'Optical Transient'],
+    },
+    {
+      name: 'GRB pattern: \\bHAWC\\b',
+      subject: 'IceCube-220306A: No significant detection in HAWC',
+      expected: ['GRB', 'Neutrino'],
```
jracusin: Nothing about this relates to GRBs

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'GRB pattern: \\bAstroSat\\b',
+      subject: 'GRB 201104A: AstroSat CZTI detection',
+      expected: ['GRB'],
+    },
+    {
+      name: 'GW pattern: \\bGW\\d+\\b',
+      subject:
+        'LIGO/Virgo GW170817: Further Hubble Space Telescope observations',
```
jracusin: Also optical and afterglow

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'FRB pattern: \\bFRB\\s?\\d{6,8}[A-Za-z]?\\b',
+      subject:
+        'FRB 20250206A: Nondetection of Repeating Bursts from FRB 20250206A with FAST',
+      expected: ['FRB'],
+    },
+    {
+      name: 'FRB pattern: \\bFRBs?\\b',
+      subject: 'FRB 181228: Global MASTER Net  optical inspection',
```
jracusin: Not a GRB

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+      expected: ['GRB', 'X-Ray Transient'],
+    },
+    {
+      name: 'X-Ray Transient pattern: \\bXRT\\b',
+      subject: 'GRB 160912A: Swift-XRT refined Analysis',
+      expected: ['GRB', 'X-Ray Transient'],
+    },
+    {
+      name: 'X-Ray Transient pattern: \\bXRF\\b',
+      subject: 'Early OT detection of XRF in NGC 2770 in Asiago frames',
```
jracusin: OT is an optical transient

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,283 @@ describe('subjectIsValid', () => {
...
+      expected: ['X-Ray Transient'],
+    },
+    {
+      name: 'X-Ray Transient pattern: \\bNuEM\\b',
+      subject:
+        'AMON Coincidence Alert from the sub-threshold IceCube-HAWC search NuEm-240118A: correction on the localization',
```
jracusin: NuEM is neutrino + EM, not necessarily a GRB or X-ray

__tests__/circulars.ts
tylerbarna: I think to make sure it matches afterglow, I'll have to add GW170817 specifically

__tests__/circulars.ts
tylerbarna: I can switch to a more appropriate test subject for this pattern, but I think this specific subject is being tagged as an XRT because of the EP pattern

app/routes/circulars/circulars.lib.ts
tylerbarna: We may need to split out gamma-ray event types into two categories: (1) GRBs (2) gamma-ray transients, which would also necessitate deciding which detectors/patterns are more appropriate for each category. I guess pretty much everything outside of "GRB" would more fall under Gamma-ray transient? 

I could also add some logic similar to the Retraction type that removes Gamma-ray transient from the list of types if GRB is present

app/routes/circulars/circulars.lib.ts
tylerbarna: @jracusin I split up the GRB event type and the gamma-ray transient event type into separate categories and made it so anything tagged as a GRB is not also tagged as a gamma-ray transient

see [f9aa1dd](https://github.com/nasa-gcn/gcn.nasa.gov/pull/3596/commits/f9aa1dd56c003a3c474e0d8a1c87cabe5af4f4c6)

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,121 @@ export function parseEventFromSubject(value: string) {
...
+  ],
+  GRB: [
+    /\bGRB\d{6}[A-Z]?\b/i,
+    /\bGRBs?\b/i,
+    /\bgamma[-\s]?ray[-\s]?bursts?\b/i,
+    /\bXRF\d{6}[A-Z]?\b/i,
+  ],
+  'Gamma-ray Transient': [
+    /\bFermi(?:\s?(?:GBM|LAT)|\d{9})?\b/i,
+    /\bSwift(?![:?/-](?:XRT|UVOT))\b/i,
```
tylerbarna: note: this pattern will match "Swift" but will *not* match "Swift-UVOT," "Swift/XRT," and the like

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,128 @@ export function parseEventFromSubject(value: string) {
     if (match) return normalize(match)
   }
 }
+
+type EventType =
+  | 'Retraction'
+  | 'GRB'
+  | 'Gamma-ray Transient'
```
tylerbarna: should we simplify the eventType internal names so they don't use spaces? The exposed naming can be more readable (and I anticipated we would do that for stuff like GRBs), but it might be simpler internally/for the url. Just a thought.

app/routes/circulars/circulars.lib.ts
```diff
@@ -288,3 +288,128 @@ export function parseEventFromSubject(value: string) {
...
+  | 'GW'
+  | 'SGR'
+  | 'FRB'
+  | 'SN'
+  | 'Neutrino'
+  | 'X-ray Transient'
+  | 'Afterglow'
+  | 'Optical Transient'
+  | 'Kilonova'
```
Vidushi-GitHub: 'TDE'
'Radio'
'Supernova'
'FXT'?

tylerbarna: - TDE: I'm only seeing a single circular with "TDE" in the subject, not sure if there are many that we could find easily using a subject matcher
- Radio: I think "Radio Transient" might be better, though I do wonder how much we're straying into wavelength rather than type of event with this. Most I'm seeing in the search results are radio counterparts to GRBs
- Supernova: They have already been implemented as SN 
- FXT: I think the X-ray transient class subsumes this; if we exclude matches for the EP FXT instrument, there's <50 matches for FXT or fast x-ray transient

app/routes/circulars/circulars.lib.ts
jracusin: Swift-BAT or Swift/BAT could indicate a gamma-ray transient or a GRB

app/routes/circulars/circulars.lib.ts
jracusin: The most famous TDE in the GCN Circ archive is https://gcn.nasa.gov/circulars/events/grb-110328?view=group&query=110328 Though it only came out through the observations reported in the Circulars that it was a TDE. I wonder if we need the ability to manually add eventTypes.

I agree that FXT is a X-ray transient. We shouldn't get more specific.

app/routes/circulars/circulars.lib.ts
jracusin: keep the spaces

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'Retraction pattern: \\bnot\\s+a\\s+(?:GRB|GW|FRB|SN|SGR|neutrino)\\b',
+      subject: 'Swift Trigger 931484 is not a GRB',
+      expected: ['Retraction'],
+    },
+    {
+      name: 'Retraction pattern: \\bprobably\\s+not\\s+a\\b',
+      subject: 'BAT GRB 060204C is probably not a GRB',
+      expected: ['Retraction'],
```
Vidushi-GitHub:
```suggestion
      expected: ['GRB', 'Retraction'],
```

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'Retraction pattern: \\bis\\s+not\\b',
+      subject:
+        'Fermi Gamma-ray Burst Monitor trigger 757464861/250101954 is not a GRB',
+      expected: ['Retraction'],
+    },
+    {
+      name: 'Retraction pattern: \\bDisregard\\b',
+      subject: 'Disregard the Swift-XRT Notices today',
```
Vidushi-GitHub: Is this a valid keyword match?

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    {
+      name: 'Gamma-ray Transient pattern: \\bFermi(?:\\s?(?:GBM|LAT)|\\d{9})?\\b',
+      subject: 'Fermi-LAT Gamma-ray Observations of IceCube-260315A',
+      expected: ['Gamma-ray Transient', 'Neutrino'],
+    },
+    {
+      name: 'Gamma-ray Transient pattern: \\bSwift[:?/-]BAT\\b',
+      subject:
+        'LIGO/Virgo S190930t: Swift J221951-484240 optical photometry of Chilescope observatory',
+      expected: ['Gamma-ray Transient', 'GW', 'Optical Transient'],
```
Vidushi-GitHub:
```suggestion
      expected: ['Gamma-ray Transient', 'GW', 'Optical Counterpart'],
```
eg see GCN Circular 21644]

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    {
+      name: 'Gamma-ray Transient pattern: \\bAstroSat\\b',
+      subject: 'LIGO/Virgo G211117: Astrosat CZTI upper limits',
+      expected: ['Gamma-ray Transient', 'GW'],
+    },
+    {
+      name: 'GW pattern: \\bGW\\d+\\b',
+      subject:
+        'LIGO/Virgo GW170817: Further Hubble Space Telescope observations',
+      expected: ['GW', 'Afterglow', 'Optical Transient', 'Kilonova'],
```
Vidushi-GitHub: I'm doubtful about using optical transient word for GRBs afterglow or GW counterparts.

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    {
+      name: 'SGR pattern: \\bsoft[-\\s]?gamma[-\\s]?repeaters?\\b',
+      subject: 'New Soft Gamma Repeater 0501+4516 was GRB 080822',
+      expected: ['GRB', 'SGR'],
+    },
+    {
+      name: 'FRB pattern: \\bFRB\\s?\\d{6,8}[A-Za-z]?\\b',
+      subject:
+        'FRB 20250206A: Nondetection of Repeating Bursts from FRB 20250206A with FAST',
+      expected: ['FRB'],
```
Vidushi-GitHub:
```suggestion
      expected: ['FRB', 'Radio Transient'],
```

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'FRB pattern: \\bFRBs?\\b',
+      subject: 'FRB 181228: Global MASTER Net  optical inspection',
```
Vidushi-GitHub:
```suggestion
      subject: 'FRB 181228: Global MASTER Net optical inspection',
```

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    },
+    {
+      name: 'Neutrino pattern: \\bSNEWS2\\b',
+      subject: 'Hypothetical SNEWS2 Circular',
+      expected: ['Neutrino'],
+    },
+    {
+      name: 'X-ray Transient pattern: (?<!\\S)EP(?=\\s|:|$)',
+      subject:
+        'FRB 20250316A: Kinder optical upper limits of the Einstein Probe candidate X-ray source  EP J120944.2+585060 ',
```
Vidushi-GitHub:
```suggestion
        'FRB 20250316A: Kinder optical upper limits of the Einstein Probe candidate X-ray source EP J120944.2+585060 ',
```

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+      expected: ['FRB', 'X-ray Transient', 'Optical Transient'],
```
Vidushi-GitHub: It's optical observations not optical transient itself.

__tests__/circulars.ts
tylerbarna: The matcher is set up so all retractions are only tagged as retractions

__tests__/circulars.ts
tylerbarna: As currently implemented, 'Optical Transient' is the category we have

__tests__/circulars.ts
tylerbarna: We haven't implemented a Radio Transient category; after our mid-sprint meeting, we determined it would not be feasible to add one

__tests__/circulars.ts
tylerbarna: The extra space is part of the original subject

__tests__/circulars.ts
tylerbarna: The extra space is part of the original subject

__tests__/circulars.ts
tylerbarna: there isn't a straightforward way of making it so it doesn't flag things that include the word optical as optical transients since this is just a basic regex

__tests__/circulars.ts
tylerbarna: Yeah 'Disregard' is one of the Retraction matchers

app/routes/circulars/circulars.lib.ts
tylerbarna: > Swift-BAT or Swift/BAT could indicate a gamma-ray transient or a GRB

I believe it's set up such that BAT matches are generically gamma-ray transients but will be retagged as GRBs if they match one of the GRB patterns, so that should be covered (ie, a subject mentioning Swift-BAT will initially be tagged as a gamma-ray transient, but if the subject also mentions a named GRB, the gamma-ray transient tag will be swapped out with the GRB tag)

__tests__/circulars.ts
tylerbarna: I could add logic to create a separate category of "Optical Counterpart" if something is tagged as both a GW/GRB and an Optical Transient, but that doesn't add any utility compared to a user just filtering for Circulars with those tags

__tests__/circulars.ts
tylerbarna: Update: after looking through the collection of subjects with "optical upper limit(s)" I've modified the matcher so it's only tagged as an optical transient if the subject includes "optical" without "upper limit(s)" 

@jracusin do you think this is the right strategy?

PR #3599, opened 2026-05-14
Update IceCube doc

azegarelli:
Following a recent discussion within the IceCube collaboration regarding the way the circular error radius is reported in the GCN schema for high-energy track alerts, we would like to update the corresponding IceCube documentation on the GCN website. We also plan to include a few additional minor improvements.

In addition, we would like to point out that the links to the GCN schema and the JSON structure for high-energy track alerts currently provided in the documentation are no longer functional, although they were working previously.


app/routes/missions.icecube/route.mdx
```diff
@@ -145,11 +145,19 @@ Key fields are described below:
...
 
+Multiplying the RA uncertainty by `cos(dec)` correctly accounts for the fact that lines of constant RA converge toward the celestial poles. This converts the RA uncertainty into a true angular distance on the sky.
+
+The adopted formula provides an effective circularized uncertainty radius derived from the localization ellipse, using the mean uncertainties around the best-fit position as the ellipse axes. The resulting `err_combined` approximates the radius of a circle with an area comparable to the estimated 90% containment region around the best-fit position.
+
+It is worth highlighting that his quantity is only an approximate representation of the localization uncertainty. By construction, if `err_combined` is used as the radius of a circular error region, it always defines a circular region centered on the best-fit coordinates. However,  the true localization morphology of an alert can be highly irregular, asymmetric, multi-modal, or offset from the best-fit position. In these situations, a simple circular approximation may exclude regions that still contain significant probability density, including high-probability pixels belonging to the true 90% containment region.
```
jracusin:
```suggestion
It is worth highlighting that this quantity is only an approximate representation of the localization uncertainty. By construction, if `err_combined` is used as the radius of a circular error region, it always defines a circular region centered on the best-fit coordinates. However, the true localization morphology of an alert can be highly irregular, asymmetric, multi-modal, or offset from the best-fit position. In these situations, a simple circular approximation may exclude regions that still contain significant probability density, including high-probability pixels belonging to the true 90% containment region.
```

app/routes/missions.icecube/route.mdx
jracusin:
```suggestion
example [JSON message](/schema/stable/gcn/notices/icecube/single_neutrino_alerts.gold_bronze_track_alerts.example.json) files
```

jracusin: Were there other links that were broken? I think we simply didn't update this once we changed the example filename in the schema repo.

PR #3600, opened 2026-05-20
Don't filter based on email and mask

dakota002:
Removes filtering from the email attribute and applies a mask to the email field for non-admin users

### DIFF
--- a/app/routes/api.users.ts
+++ b/app/routes/api.users.ts
@@ -35,17 +35,18 @@ export async function action({ request }: ActionFunctionArgs) {
         .map((x) => {
           return {
             sub: x.Attributes?.find((x) => x.Name == 'sub')?.Value,
-            email: x.Attributes?.find((x) => x.Name == 'email')?.Value ?? '',
+            email:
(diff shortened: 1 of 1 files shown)


app/routes/api.users.ts
```diff
@@ -35,17 +35,18 @@ export async function action({ request }: ActionFunctionArgs) {
         .map((x) => {
           return {
             sub: x.Attributes?.find((x) => x.Name == 'sub')?.Value,
-            email: x.Attributes?.find((x) => x.Name == 'email')?.Value ?? '',
+            email:
+              x.Attributes?.find((x) => x.Name == 'email')?.Value?.replace(
+                /(.)(.*)@(.*)/,
+                '$1*****@$3'
```
lpsinger: Replace the ignored text with a non-capturing group:

```suggestion
                /(.)(?:.*)@(.*)/,
                '$1*****@$2'
```

lpsinger (changes requested):

lpsinger (approved):

PR #3601, opened 2026-05-20
Update the circulars route

dakota002:
Check for submitter group after check for moderator check

### DIFF
--- a/app/routes/circulars._archive._index/route.tsx
+++ b/app/routes/circulars._archive._index/route.tsx
@@ -37,6 +37,7 @@ import {
   put,
   putVersion,
   search,
+  submitterGroup,
 } from '../circulars/circulars.server'
 import CircularsHeader from './CircularsHeader'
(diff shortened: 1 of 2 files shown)


app/routes/circulars._archive._index/route.tsx
```diff
@@ -128,6 +129,8 @@ export async function action({ request }: ActionFunctionArgs) {
       if (user.groups.includes(moderatorGroup)) {
         submitter = getFormDataString(data, 'submitter')
         if (!submitter) throw new Response(null, { status: 400 })
+      } else if (!user.groups.includes(submitterGroup)) {
+        throw new Response(null, { status: 403 })
```
lpsinger: For consistency with how other server-side functions like `putVersion` are defined, this check should go inside the method `createChangeRequest`.

lpsinger (changes requested):

lpsinger (approved):

PR #3602, opened 2026-05-20
Remove chromium install in prepare script

dakota002:
# Description
Reduces the `prepare` script to just be the husky command 

# Related Issue(s)
The build steps in PRs are hanging on this. The dedicated playwright actions already install the browsers to run the tests, so this is just bloating the build process


tylerbarna (approved):

lpsinger (commented):
This would break the build, because Playwright and a headless browser are required at build time to render Mermaid diagrams.

dakota002:
Ahh interesting, okay. How required is it for mermaid?

lpsinger:
> Ahh interesting, okay. How required is it for mermaid?

Mermaid diagrams won't render at all without it.

dakota002:
What is the intended case here of rendering the mermaid diagrams? Is this because of the package `mermaid-isomorphic`? [Their page](https://github.com/remcohaszing/mermaid-isomorphic) says: 

> If you want to render Mermaid diagrams in the browser directly, use the [mermaid](https://www.npmjs.com/package/mermaid) package directly.

I don't know if this would affect our usage, just trying to get an understanding of the structure

lpsinger:
```
$ npm ls mermaid
@nasa-gcn/gcn.nasa.gov@ /Users/lpsinger/src/gcn.nasa.gov
└─┬ rehype-mermaid@3.0.0
  └─┬ mermaid-isomorphic@3.0.4
    └── mermaid@11.15.0
```

So refer to the retype-mermaid documentation for renderer options.

dakota002:
> ```
> $ npm ls mermaid
> @nasa-gcn/gcn.nasa.gov@ /Users/lpsinger/src/gcn.nasa.gov
> └─┬ rehype-mermaid@3.0.0
>   └─┬ mermaid-isomorphic@3.0.4
>     └── mermaid@11.15.0
> ```
> 
> So refer to the retype-mermaid documentation for renderer options.

No I am asking about OUR usage, are we intending this to render outside of a browser context? And I mean excluding any of our playwright tests

lpsinger:
I don't understand your question. Our usage of mermaid is via the rehype-mermaid plugin. It is part of the build process, not part of the tests. What is the distinction between rehype-mermaid and OUR usage?

dakota002:
> I don't understand your question. Our usage of mermaid is via the rehype-mermaid plugin. It is part of the build process, not part of the tests. What is the distinction between rehype-mermaid and OUR usage?

My question is: How are we using this package and why is the playwright install part of our build process, if the instructions say that chromium is only used for running outside of browsers?

lpsinger:
We are using this package to render Mermaid diagrams in MDX files. rehype-mermaid requires playwright because the build does not occur inside a browser.

lpsinger:
Although we need Playwright browsers for _both_ the tests _and_ the build, this step does not increase the build time because Playwright is smart enough not to re-download Chromium if it has already done so.

This was an interesting idea, but it is not suitable for or our build process.

dakota002:
> Although we need Playwright browsers for _both_ the tests _and_ the build, this step does not increase the build time because Playwright is smart enough not to re-download Chromium if it has already done so.
> 

Agreed, though this issue is not stemming from a local problem with build times, but the GitHub actions pipelines taking forever on Node 26 builds 

> This was an interesting idea, but it is not suitable for or our build process.

I see. Our use of mermaid is for one small diagram in the auth docs, so I would instead change my argument, we should instead just use mermaid as a component and do client side rendering. This latest commit is a working example, however it is going to break the GitHub linting checks from the dynamic import.

I updated the `tsc` command in the lint script to target the same module value we have in the tsconfig, and this does work correctly locally. So updating the one for GitHub should also see this passing, but I think that is maintained in another repo now, right?

What are your thoughts on an implementation like this instead?

lpsinger:
No, we should not render the Mermaid diagrams on the client side.

lpsinger:
> Agreed, though this issue is not stemming from a local problem with build times, but the GitHub actions pipelines taking forever on Node 26 builds

Which pipelines? Why is it timing out under Node 26 specifically?

lpsinger:
Ah, this was reported to Playwright as https://github.com/microsoft/playwright/issues/40724 and should be fixed in 1.60.0.

PR #2713, opened 2024-11-12
Circulars: Add In-Page Lucene Documentation

tylerbarna:
# Description
This adds a brief explanation of Lucene queries to the Circulars Archive page, hidden behind a feature flag. 

image

Note: This is a simplified version of #2589 that removes the quick action buttons until the proper implementation can be identified.
# Related Issue(s)
Tracked by https://github.com/nasa-gcn/gcn.nasa.gov/issues/2288

# Testing
1. Navigate to Circulars Archive
2. Click on the "Advanced Search" (Tentative Label) Accordion Menu
3. The menu should expand

PR #3605, opened 2026-05-20
Add table of contents to documentation

### LINKED ISSUE #3604: Add Table of Contents to documentation
# Description
Add tables of contents to documentation pages that link to each subheader. This is useful as some documentation pages grow in length and number of items covered.

# Acceptance criteria

- [ ] Table of contents lists relevant subheadings
- [ ] Add to all relevant documentation pages

tylerbarna:
# Description

Per our discussion in our most recent sprint planning, some of the documentation pages are becoming long enough to benefit from a table of contents. Rather than using markdown for each page and updating as new sections are added, it would be useful to have a component to generate the table of contents dynamically.

This adds a new TableOfContents component that finds all subheadings on a page and compiles them into a list of links to each. When clicked, each item will scroll the page to that subheading. By default, the table of contents is collapsed in a manner similar to #2713 

image

As of opening this PR, this is just implemented for the Circulars archive documentation page, but the component should be usable for any documentation page that uses subheaders to organize the content.

# Related Issue(s)

Resolves #3604 

# Testing

1. Navigate to Circulars documentation on the archive
2. Expand the table of contents
3. Click on any of the links
4. They should direct you to the relevant subheading

### DIFF
--- a/app/components/TableOfContents.tsx
+++ b/app/components/TableOfContents.tsx
@@ -1,0 +1,58 @@
+import classNames from 'classnames'
+import { useEffect, useState } from 'react'
+
+type TocItem = {
+  id: string
+  text: string
+}
+
+function compileTOC(container: ParentNode): TocItem[] {
+  return Array.from(container.querySelectorAll('h2, h3, h4, h5, h6'))
+    .map((heading) => ({
+      id: heading.id,
+      text: heading.textContent?.trim() ?? '',
+    }))
+    .filter((heading) => heading.id && heading.text)
+}
+
+export function TableOfContents({
+  className,
+  selector = '[data-mdx-content]',
+}: {
+  className?: string
+  selector?: string
+}) {
+  const [items, setItems] = useState<TocItem[]>([])
+
+  useEffect(() => {
+    const container = document.querySelector(selector)
+
+    if (container) {
+      setItems(compileTOC(container))
+    }
+  }, [selector])
+
+  if (items.length === 0) {
+    return null
+  }
+
+  return (
+    <nav
+      aria-label="Table of contents"
+      className={classNames('margin-y-4 padding-2', className)}
+    >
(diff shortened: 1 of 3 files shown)


lpsinger (changes requested):
This will only be rendered on the client side, so it may cause layout jumps.

Did you check if the is an existing remark/rehype plugin that will do this? I bet there are several.

tylerbarna:
> This will only be rendered on the client side, so it may cause layout jumps.
> 
> Did you check if the is an existing remark/rehype plugin that will do this? I bet there are several.

do you mean in regards to the table of contents generally or the movement of the page from the table of contents to the relevant header?

jracusin:
This looks great and very useful. I also wondered if you'd checked if there were existing components.

lpsinger:
See https://github.com/remarkjs/remark/blob/main/doc/plugins.md, https://github.com/rehypejs/rehype/blob/main/doc/plugins.md.

Also I am not entirely convinced of creating a per-page TOC. What about splitting this long page into multiple pages?

tylerbarna:
> This looks great and very useful. I also wondered if you'd checked if there were existing components.

I had looked through a couple, but I didn't see any that matched our design language/would automatically grab all the sections, but I can dive a bit deeper

lpsinger:
This implementation would not be acceptable because it is not compatible with server-side rendering.

jracusin:
What is the resolution here? In other documentation pages, when we have multiple sections, we have either split them into separate pages (e.g. https://gcn.nasa.gov/docs/sample) or separate links to different parts of the same page (https://gcn.nasa.gov/docs/client). We could do either for the Circulars archive page, but it would require going one more level deep in the menu. I agree that this page needs something because it is long and contains a bunch of different information.

jracusin:
Let's make a simplified intro paragraph and move the Circulars migration into a migration top menu

jracusin:
Superseded by #3686 and #3687.

PR #3609, opened 2026-05-22
Update playwright to 1.60.0

lpsinger:
This fixes hangs when installing Playwright browsers on Node 26. See https://github.com/microsoft/playwright/issues/40724, https://github.com/microsoft/playwright/pull/40747.


dakota002 (commented):
Shouldn't the package.json also be part of this?

PR #3578, opened 2026-05-01
Teams Update: Kafka Acl Functions

dakota002:
# Description

- Provision app client for admin actions
- Add functions for:
  - Building `AclEntry` objects based on Client Type ('producer'/'consumer')
  - Creating ACLs on brokers
  - Deleting ACLs on brokers
  - Retrieving ACLs from brokers
  - Updating topic as Public/Private

Also I am open to changing the use of producer/consumer wording in code to write/read

# Related Issue(s)
Relates to https://github.com/nasa-gcn/gcn.nasa.gov/pull/3577, referenced in TODO comments

# Testing

Working on tests

PR #3623, opened 2026-05-27
Rework app client plugin to allow for reuse

dakota002:
Will be useful for https://github.com/nasa-gcn/gcn.nasa.gov/pull/3578 to reduce duplicated code


src/plugins/app-clients.js
```diff
@@ -0,0 +1,47 @@
...
+          GenerateSecret: true,
+          UserPoolId: user_pool_id,
+        },
+      }
+      env[envClientIdName] = {
+        'Fn::GetAtt': `${clientKey}.ClientId`,
+      }
+      env[envClientSecretName] = {
+        'Fn::GetAtt': `${clientKey}.ClientSecret`,
+      }
```
lpsinger, line 47: Exposing these through the [`deploy.services`](https://arc.codes/docs/en/guides/plugins/deploy#deploy.services) plugin entry point will be cleaner here. That way you can return a structure of all of the app clients that is guaranteed not to collide with any other environment variables.

src/plugins/app-clients.js
```diff
@@ -0,0 +1,47 @@
...
+          },
+        ],
+      ] = Object.entries(item)
+      const env =
+        cloudformation.Resources[envResourceName].Properties.Environment
+          .Variables
+      const user_pool_id = env.COGNITO_USER_POOL_ID
+      if (!user_pool_id)
+        throw new Error('Environment variable COGNITO_USER_POOL_ID must be set')
+      cloudformation.Resources[clientKey] = {
```
lpsinger: Give the resource some prefix to ensure that it is unique, such as `AppClient${clientKey}`.

lpsinger (changes requested):

src/plugins/app-clients.js
dakota002: Should this whole plugin be moved into services? Or are you suggesting just the variable names?

lpsinger: Not the whole plugin, no, just exporting these values to the application environment. The `deploy.start` entry point is for customizing the CloudFormation resources whereas the `deploy.services` entry point is for customizing values transmitted to the application through Architect's service discovery mechanism.

dakota002:
Is this more what you have in mind?

src/plugins/app-clients.js
lpsinger: That's not how this entry point works. Please read the documentation and study the example code at https://arc.codes/docs/en/guides/plugins/deploy#deploy.services.

lpsinger (changes requested):

src/plugins/app-clients.js
dakota002: The example only shows this one:
```js
return {
      bucketName,
      accessKey: isLocal ? 'local' : { Ref: 'MyS3BucketCreds' },
      secretKey: isLocal ? 'local' : { 'Fn::GetAtt': [ 'MyS3BucketCreds', 'SecretKey' ] }
}
```
but `arc.services()` looks like it can be any type of supported resource:
```js
{
  // a plugin named 'imagebucket' exposing some service discovery variables
  imagebucket: {
    accessKey: 'someAccessKey',
    name: 'arc-plugin-s3-image-bucket-example-image-buket',
    secretKey: 'someSecretKey'
  },
  // built-in @static service discovery variables
  static: {
    bucket: 'arcplugins3imagebucketexamplestaging-staticbucket-g8rsuk82ancj',
    fingerprint: 'false'
  },
  // built-in @events service discovery variables
  events: {
    myevent: 'https://some-sns-url.amazon.us-east-2.com'
  }
}
```
But this looks like if we have multiple app client definitions in one plugin, it will get returned likely as something like:

```js
appclient: {
   {
     // Stuff from first one, 
   },
   {
     // Stuff from second one, 
   },
   ... etc
}
```
So what would the structure in the returned object look like to just set env variables? would each one return an object like
```js
{
  resourceName,
  properties: {
    environment: {
      variables: {
        envClientIdName: {
          'Fn::GetAtt': `${clientKey}.ClientId`,
        },
        envClientSecretName:{
        'Fn::GetAtt': `${clientKey}.ClientSecret`,
        }
      }
    }
  }
}
```

lpsinger: It's not supposed to return CloudFormation resources at all. Those are just arbitrary JavaScript key-value objects whose values may be CloudFormation expressions.

dakota002: Oh, why would we return it this way then? What do you want to do with the returned JSON?

app.arc
```diff
@@ -13,6 +13,14 @@ circulars
 support
   src build/email-incoming/support
 
+@appclients
+circulars-producer
+  clientKey GcnCircularsProducerUserPoolClient
+  envResourceName CircularsTableStreamLambda
+  scope gcn.nasa.gov/kafka-circulars-producer
+  envClientIdName KAFKA_CLIENT_ID
+  envClientSecretName KAFKA_CLIENT_SECRET
```
lpsinger, line 22:
```suggestion
circulars-producer
  scope gcn.nasa.gov/kafka-circulars-producer
```

All of these surplus properties can be derived from the name (in this case, `circulars-producer`).

src/plugins/app-clients.js
lpsinger: The returned object is made available to the application through the return value of `arc.services()`.

lpsinger (changes requested):

src/plugins/app-clients.js
dakota002: Wait so it the point so that by defining some json in the service function, I can call `arc.services()` in the start function? Or is it that I should use `arc.services()` in `kafka.server.ts` instead of `getEnvOrDieInProduction('KAFKA_CLIENT_ID')`?

lpsinger: > Or is it that I should use arc.services() in kafka.server.ts instead of getEnvOrDieInProduction('KAFKA_CLIENT_ID')?

That one.

dakota002:
I think this is now the correct implementation, but I am not seeing any entries in the services object when I log it locally, so I may be returning them incorrectly,

src/plugins/app-clients.js
```diff
@@ -0,0 +1,47 @@
...
+      if (!user_pool_id)
+        throw new Error('Environment variable COGNITO_USER_POOL_ID must be set')
+      cloudformation.Resources[`AppClient${key}`] = {
+        Type: 'AWS::Cognito::UserPoolClient',
+        Properties: {
+          AllowedOAuthFlows: ['client_credentials'],
+          AllowedOAuthFlowsUserPoolClient: true,
+          AllowedOAuthScopes: [scope],
+          GenerateSecret: true,
+          UserPoolId: user_pool_id,
```
lpsinger:
```suggestion
      const UserPoolId = process.env.COGNITO_USER_POOL_ID
      if (!UserPoolId)
        throw new Error('Environment variable COGNITO_USER_POOL_ID must be set')
      cloudformation.Resources[`AppClient${key}`] = {
        Type: 'AWS::Cognito::UserPoolClient',
        Properties: {
          AllowedOAuthFlows: ['client_credentials'],
          AllowedOAuthFlowsUserPoolClient: true,
          AllowedOAuthScopes: [scope],
          GenerateSecret: true,
          UserPoolId,
```

src/plugins/app-clients.js
```diff
@@ -0,0 +1,47 @@
...
+          : {
+              'Fn::GetAtt': `${key}.ClientId`,
+            },
+        clientSecret: isLocal
+          ? 'localClientSecret'
+          : {
+              'Fn::GetAtt': `${key}.ClientSecret`,
+            },
+      }
+    })
```
lpsinger: - Lift the definition of `isLocal` out of the loop.
- As written, this entry point returns `undefined` because `Array.prototype.forEach` returns `undefined`.

```suggestion
    const isLocal = stage === 'testing'
    return Object.fromEntries(appclients.map((item) => {
      const [[key]] = Object.entries(item)
      return {
        key,
        clientId: isLocal
          ? 'localClientId'
          : {
              'Fn::GetAtt': `${key}.ClientId`,
            },
        clientSecret: isLocal
          ? 'localClientSecret'
          : {
              'Fn::GetAtt': `${key}.ClientSecret`,
            },
      })
    })
```

lpsinger (changes requested):

src/plugins/app-clients.js
dakota002: Yea, I realized that.
It also doesn't like a nested response, it needs to be just key value(string) pairs so I am reworking it a bit

dakota002:
What are your thoughts on how to call the services function? The top level await like this causes problems

```js
const service_defs = await services()

const client_id = service_defs['circulars-producer-clientId'] ?? ''
const client_secret = service_defs['circulars-producer-clientSecret'] ?? ''
const kafka = new Kafka({
  client_id,
  client_secret,
  domain,
})

```

```
 error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', 'node18', 'node20', 'nodenext', or 'preserve', and the 'target' option is set to 'es2017' or higher.
```

lpsinger:
> What are your thoughts on how to call the services function? The top level await like this causes problems

Ideally, have esbuild emit ESM rather than CJS modules.

dakota002:
> > What are your thoughts on how to call the services function? The top level await like this causes problems
> 
> Ideally, have esbuild emit ESM rather than CJS modules.

How about this implementation though, instead of reworking esbuild for now? I think thats a bigger lift for another day

lpsinger:
Alright, let’s have it set environment variables for now, but please add a FIXME comment.

dakota002:
Reverted back to just set the variables

Courey (approved):

PR #3639, opened 2026-06-05
Add no store cache header to users api

dakota002:


tylerbarna (approved):

jracusin (approved):

lpsinger (changes requested):
What is this supposed to do? Our application-wide middleware disables caching on all responses that do not explicitly contain cache control headers:

https://github.com/nasa-gcn/gcn.nasa.gov/blob/473c7b4b6c310e5d9b8736fa317112c6cb95659d/server.ts#L29

PR #3666, opened 2026-06-17
Add matchers for event types 

tylerbarna:
# Description

This PR adds the regular expression matcher function that is required for https://github.com/nasa-gcn/gcn.nasa.gov/issues/3517. There are a number of regular expressions for each of the event types. Generally, it written in such a way that subjects can be tagged with multiple event types, but the "Retraction" tag is special in that it supersedes all others. The "Misc" tag is not explicitly named but the function will return undefined if it matches none of the regular expressions.

This PR also adds testing to validate that the function works with subjects from the archive.

# Related Issue(s)

Resolves https://github.com/nasa-gcn/gcn.nasa.gov/issues/3518

# Testing

1. Review new code in circulars.lib.ts
2. Run tests in circulars.ts

Note: This is a fresh PR due to squashing issues that broke #3596

PR #3667, opened 2026-06-17
Add backend to filter by event type in archive search

### LINKED ISSUE #3521: Update circulars archive search backend to filter by eventType

tylerbarna:
# Description

Adds the backend for the circulars archive to filter circulars by tagged event types. 

Additionally, this PR includes boolean logic that will allow for users to select whether to filter circulars with all (AND) or any (OR) of the selected event types. This logic is set as an argument in the search function. There is also the option to exclude specified event types (NOT), which are contained within a separate argument for excluded event types.

This PR refactors the filter aspect of the search function to be more modular, with event type filtering being separate from the date filtering and being hidden behind a feature flag 'EVENTTYPE' for now.

# Related Issue(s)

Resolves #3521 

# Testing

1. Enable EVENTTYPE in .env
2. Ensure seed.json is the version with eventType tags included
3. Start container
4. Navigate to Circulars archive
6. In circulars.server.ts, change the default values for eventTypes, eventTypesLogic, and evenTypesExclude (example: eventTypes=["GRB", "Afterglow"], eventTypesLogic="AND", eventTypesExclude=["GW"]) and verify that the listed circulars change.


tylerbarna:
Note: based this branch off the one from #3666

__tests__/circulars.ts
```diff
@@ -1346,3 +1347,384 @@ describe('subjectIsValid', () => {
...
+    {
+      name: 'X-ray Transient pattern: \\bSwift[:?/-]XRT\\b',
+      subject: 'Enhanced Swift-XRT position of SGR 0501+4516',
+      expected: ['SGR', 'X-ray Transient'],
+    },
+    {
+      name: 'Afterglow pattern: \\bafterglows?\\b',
+      subject: 'GRB 041006 Optical afterglow observations',
+      expected: ['GRB', 'Afterglow', 'Optical Transient'],
+    },
```
jracusin, line 1654: afterglows are not distinctly optical

jracusin:
I've been thinking about the "gamma-ray transient" distinction from a GRB. The way the logic works, I think any X-ray afterglow will be an X-ray transient, any optical afterglow will be an optical transient, any radio afterglow will be a radio transient, so I think we should be consistent.  If something is distinctly an SGR or other type of gamma-ray transient, that should be separate.

tylerbarna:
> I've been thinking about the "gamma-ray transient" distinction from a GRB. The way the logic works, I think any X-ray afterglow will be an X-ray transient, any optical afterglow will be an optical transient, any radio afterglow will be a radio transient, so I think we should be consistent. If something is distinctly an SGR or other type of gamma-ray transient, that should be separate.

@jracusin  I'll see about trying to implement this, but that would probably be a part of #3666. This PR adds the search backend, it's just showing the other things as changes because the other PR hasn't been merged into main yet

__tests__/circulars.ts
tylerbarna: This is due to matching for "optical" in the subject; I could change the matcher so it matches "optical" but not "optical afterglow"

jracusin: Ah, it's just triggering on the word "afterglow"

tylerbarna: yeah "GRB" is triggering "GRB," "afterglow" is triggering "afterglow," and "Optical" is triggering "Optical Transient"

I wrote the tests up to use actual circular subjects and have a test for each pattern in the matcher, so there are a bunch of subjects in the tests that match multiple event types even if the test name is referring to check for one type. It's just making sure that specific pattern works with the side effect of testing a couple other patterns. I thought this would be a bit more robust than using fake subjects that I write to match the pattern I want them to match

PR #3688, opened 2026-07-06
Update producer documentation for remaining connected

### LINKED ISSUE #3581: Modify producer documentation to advise constant Kafka connectivity
# Description
We don't want to miss ultra-rare transients because of expired producer credentials, so we want producers to never expire because they are always connected.

# Acceptance criteria
- Inventory when producers were last connected to see if this is even an issue
- update documentation

jracusin:
# Description
We are concerned that producer credentials can expire when disused due to rare alert topics.  A cursory look at the time since last producer connection shows that most producers are only running the producer code (including connecting to Kafka) when an alert is generated, rather than remaining connected at all times.  This PR adds information to the producer documentation that explains where their pipeline code should loop, such that they are connected at all times to Kafka, and which steps to run when an alert is generate.

Once this is merged, we need to reach out to all of the producers and ask them to modify their pipelines.

# Related Issue(s)
Resolves #3581

# Testing
local dev


jracusin:
Totally messed up a merge conflict. Closing and opening a new PR with these minor changes.

jracusin:
Replaced by #3800

PR #3691, opened 2026-07-07
Add Nova to subject keywords

jracusin:
# Description
This PR adds "Nova" to the Circular subject keywords list.

# Related Issue(s)

# Testing

### DIFF
--- a/app/routes/circulars/circulars.lib.ts
+++ b/app/routes/circulars/circulars.lib.ts
@@ -207,6 +207,7 @@ export const validSubjectKeywords = [
   'MAGIC',
   'MASTER',
   'MAXI',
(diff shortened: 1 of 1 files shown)


Vidushi-GitHub (approved):

Vidushi-GitHub (commented):
is this case sensitive, if so then should be 'nova'

Refer ATel: https://www.astronomerstelegram.org/?read=17218

jracusin:
> is this case sensitive, if so then should be 'nova'
> 
> Refer ATel: https://www.astronomerstelegram.org/?read=17218

It is not case sensitive

dakota002 (approved):

PR #3702, opened 2026-07-16
Add users opensearch index functions

dakota002:
Adds:
- Function that converts from a cognito user to the format opensearch expects
- function to put users into a new user index 
- search users function, which can be optionally filtered by group 

Updates:
- Seed data for local testing

### DIFF
--- a/app/lib/opensearch.server.ts
+++ b/app/lib/opensearch.server.ts
@@ -1,0 +1,86 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { search as getSearchClient } from '@nasa-gcn/architect-functions-search'
+
+import {
+  extractAttribute,
+  extractAttributeRequired,
+  getCognitoUserFromSub,
+  getUserGroupStrings,
+} from './cognito.server'
+
+// Gets user from Cognito by sub to be indexed into OpenSearch
+export async function getUserForOpenSearch(sub: string) {
+  const user = await getCognitoUserFromSub(sub)
+  if (!user) throw new Response(null, { status: 404 })
+  const groups = user.Username
+    ? ((await getUserGroupStrings(user.Username)) ?? [])
+    : []
+  return {
+    sub: extractAttributeRequired(user.Attributes, 'sub'),
+    email: extractAttributeRequired(user.Attributes, 'email'),
+    name: extractAttribute(user.Attributes, 'name'),
+    affiliation: extractAttribute(user.Attributes, 'custom:affiliation'),
+    groups,
+  }
+}
+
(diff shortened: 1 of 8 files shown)


app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,65 @@
+/*!
```
Courey, line 1: Is this really a cognito server or should we call it something more specific to what it's being used for? it seems more like its a user index instead of a cognito server since it's all related to the user index. or is there an intent to expand it beyond users?

package.json
```diff
@@ -104,7 +104,7 @@
     "@babel/preset-typescript": "^7.27.0",
     "@gitlab/svgs": "^3.83.0",
     "@nasa-gcn/architect-plugin-dynamodb-local": "^2.0.0",
-    "@nasa-gcn/architect-plugin-search": "^1.9.0",
+    "@nasa-gcn/architect-plugin-search": "github:dakota002/architect-plugin-search#66f1c1100853e7cd9f0bc30d025eee97f81531dc",
```
Courey: is this just until https://github.com/nasa-gcn/architect-plugin-search/pull/346/changes goes in? should we make this one blocked by that one and then we would just have a version bump instead?

app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function searchUsersIndex(name: string, group?: string) {
+  const client = await getSearchClient()
+
+  const queryObject = {
+    query: {
+      bool: {
+        must: [
+          {
+            query_string: {
+              query: `*${name}*`,
```
Courey, line 52: is there acceptable variability in the name we will be passing in or will we always be using the exact group name? I'm not sure if this is going to be something that would ever be queried from a query string or if it would always be us passing in the full group name. If it's always us, I think it should be an exact match.

Courey (changes requested):
Nothing is a deal breaker or anything, just figured we could talk through the things I pointed out to see if they even need to change.

app/lib/opensearch.server.ts
dakota002: I think you're referring to the `getUserForOpenSearch` function I added to the cognito server. I can move that function here.
As for this file, I see it being expanded on further for other opensearch related functionality, to reduce code duplication

app/lib/opensearch.server.ts
dakota002: The group name should (probably) always be an exact match if provided, as it would be jsut filtering the results down to specific groups of users

package.json
dakota002: Yea, I mean this works, but it may be best to see why the tests are failing in that repo first

dakota002: Update: we already have a solution and I have reverted this

tylerbarna (approved):
I concur with Courey about [fuzzy search](https://docs.opensearch.org/latest/query-dsl/term/fuzzy/) if this is ever a query string input, but if it's as you describe, everything looks correct to me

PR #3712, opened 2026-07-20
Update Kafka client instructions for @confluentinc/kafka-javascript

lpsinger:
The new client requires an explicit call to `connect()`.


mcoughlin (approved):
Time to use this in SkyPortal:)

lpsinger:
> Time to use this in SkyPortal:)

@confluentinc/kafka-javascript would also be the best tool for interacting with SCiMMA's Kafka cluster from JavaScript.

dakota002 (approved):

PR #3719, opened 2026-07-21
Adds DMARC validation check to incoming email handler

Courey:
# Description
This adds a check for incoming emails that fails unless DMARC passes.

# Testing
I have not tested this. I would love some input on how to go about testing this prior to putting it on dev.

### DIFF
--- a/app/email-incoming/handler.ts
+++ b/app/email-incoming/handler.ts
@@ -30,6 +30,8 @@ export function createEmailIncomingMessageHandler(
       throw new Error('Message failed virus check')
     if (message.receipt.action.type !== 'S3')
       throw new Error('Action type must be S3')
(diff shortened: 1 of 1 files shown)


jracusin:
Implementation of this should wait for outreach to users whose Circular emails would be rejected due to email server misconfiguration or pipeline/web application usage that would not have proper dmarc approvals.

Courey:
In our weekly meeting we decided to initially only throw on a DMARC fail result

dakota002 (approved):

lpsinger:
What's the status of this PR?

dakota002:
The PR is fine, but I thought we were waiting on some groups to make updates so they didn't get blocked from submission, @jracusin is that still the case or can we merge now?

Issue #3520, opened 2026-03-09
Add eventType association to Circular posting pipeline


tylerbarna:
# Description
Have the regular expression tool run at the time a circular is submitted to associate it with specific event types prior to being posted. This should be similar to the manner in which event IDs are associated with a circular at time of submission

# Acceptance criteria

- [ ] Call regular expression after a circular is submitted
- [ ] Assign the returned event types as a list in the eventType parameter for the circular

PR #3723, opened 2026-07-21
Add backend to search for event types

### LINKED ISSUE #3521: Update circulars archive search backend to filter by eventType

tylerbarna:
# Description

Adds the backend for the circulars archive to filter circulars by tagged event types. 

Additionally, this PR includes boolean logic that will allow for users to select whether to filter circulars with all (AND) or any (OR) of the selected event types. This logic is set as an argument in the search function. There is also the option to exclude specified event types (NOT), which are contained within a separate argument for excluded event types.

This PR refactors the filter aspect of the search function to be more modular, with event type filtering being separate from the date filtering and being hidden behind a feature flag 'EVENTTYPE' for now. 

When adding future search/filter features, this modularity should make it easier to modify

In order to test this, you will have to temporarily replace your seed.json, as we don't have #3520 implemented yet.

Note: this is the same as #3667, which was closed as it was based on an out of date fork.

# Related Issue(s)

Resolves #3521 

# Testing

1. Enable EVENTTYPE in .env
2. Replace seed.json with the version with eventType tags included (located [here](https://github.com/tylerbarna/gcn.nasa.gov/blob/db9819e55ecdd75fbe0b285115c751a68ca252cc/seed.json))
3. Start container
4. Navigate to Circulars archive
6. In circulars.server.ts, change the default values for eventTypes, eventTypesLogic, and evenTypesExclude (example: eventTypes=["GRB", "Afterglow"], eventTypesLogic="AND", eventTypesExclude=["GW"]) and verify that the listed circulars change.

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -203,7 +203,7 @@ export async function search({
   ]
 
   if (feature('EVENTTYPE')) {
-    if (eventTypes && eventTypes.length > 0) {
+    if (eventTypes?.length > 0) {
       if (eventTypesLogic === 'AND') {
         filterConditions.push({
           bool: {
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -147,7 +147,7 @@ export async function search({
   startDate,
   endDate,
   eventTypes,
-  eventTypesLogic = 'AND',
+  eventTypesLogic = 'OR',
   eventTypesExclude,
   sort,
 }: {
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -42,7 +42,7 @@ import type {
   CircularMetadata,
 } from './circulars.lib'
 import { sendEmail, sendEmailBulk } from '~/lib/email.server'
-import { origin } from '~/lib/env.server'
+import { feature, origin } from '~/lib/env.server'
 import { truncateJsonMaxBytes } from '~/lib/utils'
 import { closeZendeskTicket } from '~/lib/zendesk.server'
 
@@ -146,6 +146,9 @@ export async function search({
   limit,
   startDate,
   endDate,
+  eventTypes,
+  eventTypesLogic = 'AND',
+  eventTypesExclude,
   sort,
 }: {
   query?: string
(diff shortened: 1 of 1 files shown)


app/routes/circulars/circulars.server.ts
```diff
@@ -146,13 +146,19 @@ export async function search({
   limit,
   startDate,
   endDate,
+  eventTypes,
+  eventTypesLogic = 'AND',
```
tylerbarna:
```suggestion
  eventTypesLogic = 'OR',
```

app/routes/circulars/circulars.server.ts
```diff
@@ -185,20 +191,56 @@ export async function search({
...
+        createdOn: {
+          gte: startTime,
+          lte: endTime,
+        },
+      },
+    },
+  ]
+
+  if (feature('EVENTTYPE')) {
+    if (eventTypes && eventTypes.length > 0) {
```
dakota002:
```suggestion
    if (eventTypes?.length > 0) {
```
You can use optional chaining here and in general when checking `variable && variable.something`

dakota002 (changes requested):
Looks generally good. One small change listed.
Not a requirement, but these fairly complex search objects may be better constructed by function, which we could have a simple unit test for to ensure that the passed in args return the expected filter/body/etc

app/routes/circulars/circulars.server.ts
tylerbarna: broke

dakota002 (approved):
My previous advice was bad, this is good

PR #3731, opened 2026-07-22
Separate Circulars archive header from index in failure mode

### LINKED ISSUE #3690: Improve UI for Circulars page when handling OpenSearch unavailability
# Description
When OpenSearch is failing or otherwise unavailable, the whole Circulars page appears to be unavailable even though submission and distribution are still functioning. We should keep the top portion of the page (new, etc.) visible and show the relevant errors in place of the archive.

# Acceptance criteria
- [x] page header still displays under error conditions
- [x] error appears where the Circulars archive usually appears with message along the lines of "Archive is currently unavailable"
- [ ] expand tests to check for error cases (e.g. hitting Circulars archive page before OpenSearch spins up)

tylerbarna:
# Description

This separates out the header of the archive and the index so the page will still show the header in the event that the index fails to load (such as in the case of opensearch failing). This is accomplished using an [error boundary.](https://v2.remix.run/docs/route/error-boundary)

Screenshot 2026-07-23 at 13 24 49

# Related Issue(s)

Resolves #3690 

# Testing

1. Navigate to circulars archive in local development environment
2. Initiate a search for any query
3. This should load
4. Initiate a search for `crash_test` (note: this is an artificial test and the code for this will be removed prior to merging)
5. This will artificially raise an error similar to an opensearch error and show the error above while still rendering the header


tylerbarna:
In 062209b, having trouble identifying an easy way to simulate crashing opensearch

Additionally, this may or may not have to do with the above, but when I have crashed without crashing the entire local dev, the generic crash page is shown. Not sure if it is superseding the crash and needs to be intercepted

lpsinger:
There is a much, much simpler way to do this: an error boundary. Please look it up in the Remix documentation.

lpsinger (changes requested):
Again, it is my advice to use an _error boundary_. This is not an error boundary.

tylerbarna:
> Again, it is my advice to use an _error boundary_. This is not an error boundary.

Apologies, I was trying to construct an implementation similar to the one shown in the [documentation for error boundaries](https://v2.remix.run/docs/route/error-boundary).

Courey:
So I was fiddling with this and I tried putting an <Outlet /> in the circulars._archive._index but if I then create a file called circulars._archive._index._header it won't build because you can't have a child route in an index. I haven't dug in to see if we can just change the index to something else and then give it a child, but this is the error I got when I tried the above. 
Screenshot 2026-07-23 at 4 57 13 PM

Courey:
I did some research and fiddling and I may have progress on enabling this. Since the file is named _index you can't put a child in because it's a terminal leaf thus it can't have children. But since you want a component level error boundary, that is something that react error boundary supports. https://www.npmjs.com/package/react-error-boundary
I ran a basic test on this. I added react-error-boundary to the package.json and imported it into the _index/route.tsx
I created a simple fallback error:
Screenshot 2026-07-24 at 10 50 06 AM
I wrapped the components I wanted to include in the error boundary:
Screenshot 2026-07-24 at 10 49 55 AM
I just put a throw in so that the group index would error out. This shows that the boundary around the group index and circulars index is where the error is caught. 
Screenshot 2026-07-24 at 10 49 33 AM

lpsinger:
Please permit me time to review this.

jracusin:
Try inserting another hidden route level and put the error boundary there.

Issue #3532, opened 2026-03-09
Create frontend for event type email selection


tylerbarna:
# Description
Update user preferences page so they can select which event types to subscribe to circulars for (see https://github.com/nasa-gcn/gcn.nasa.gov/issues/3523)

# Acceptance criteria

- [ ] In the email notifications section of the user preferences, include an expandable menu to select which event types to subscribe to
- [ ] By default, have all selected
- [ ] Include options to select all or clear selections (clearing selections should also switch off circular subscriptions in general)
- [ ] (for discussion) include a confirmation button and/or highlight any changes in a user's selections from their existing preferences

PR #3735, opened 2026-07-23
Circulars by event type emails backend

### LINKED ISSUE #3523: Create backend for users to select which event types to receive emails for
# Description
Add a feature that allows users to select which circular Kafka topics to subscribe to via email

# Acceptance criteria
Add:
- [ ] New Table for tracking relation between user and event type
  - [ ] Fields: eventType, sub, email, ...?

Update:
- [x] `app\table-streams\circulars-email\index.ts` to check table for users who have matching subscriptions for an event type, reduce list to remove duplicates, then send email
- [ ] `user/email` page to allow selecting specifics
- [ ] Selecting all circular types maintains row in `circulars_subscriptions`
- [ ] When a user selects any subset of circular types, remove their subscription from the previous table, add row to new table 
- [ ] By default, users signed up for circular emails should still receive all circulars (should not require any changes, as this will continue to behave as expected)
- [ ] Users should be able to select a subset of event types
- [ ] Users should be notified by email when their email preferences have been updated

dakota002:
- Adds new tables for circular email subscriptions associated to specific event types
- Refactors a reusable `EmailSubscription` type based on the notices emails
- Adds functions to `app/routes/circulars/circulars.server.ts` to send event type email and get addresses from the new list
- Updates table stream lambda to send event type emails after main send behind feature flag

### DIFF
--- a/app/routes/circulars/circulars.server.ts
+++ b/app/routes/circulars/circulars.server.ts
@@ -6,12 +6,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { tables } from '@architect/functions'
-import type { DynamoDB } from '@aws-sdk/client-dynamodb'
-import {
-  type DynamoDBDocument,
-  paginateQuery,
-  paginateScan,
-} from '@aws-sdk/lib-dynamodb'
+import { type DynamoDB } from '@aws-sdk/client-dynamodb'
+import { type DynamoDBDocument, paginateScan } from '@aws-sdk/lib-dynamodb'
 import { search as getSearch } from '@nasa-gcn/architect-functions-search'
 import {
   DynamoDBAutoIncrement,
@@ -27,10 +23,10 @@ import {
   manageSynonymOnVersionUpdates,
   tryInitSynonym,
 } from '../synonyms/synonyms.server'
+import { send } from '../user.email/email_circulars.server'
 import {
   bodyIsValid,
   formatAuthor,
-  formatCircularText,
   formatIsValid,
   parseEventFromSubject,
   parseEventTypeFromSubject,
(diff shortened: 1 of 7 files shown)


tylerbarna:
@dakota002 maybe split the the code migration for the existing email functionality as a separate PR and rebase off of that?

dakota002:
I have reverted any refactoring changes from this PR

app/routes/circulars/circulars.server.ts
```diff
@@ -814,3 +828,83 @@ export async function send(circular: Circular) {
...
   })
 }
+
+export async function sendEventTypedEmails(circular: Circular) {
+  if (!circular.eventType) throw new Response(null, { status: 500 })
+  const to = await getEmailsForEventTypes(circular.eventType)
+  await sendBulkCircularsTo(circular, to)
+}
+
+export async function createEventBasedCircularEmailSubscription(
```
tylerbarna: may want to change so it's clear this isn't referring to event groups
```suggestion
export async function createEventTypeBasedCircularEmailSubscription(
```

tylerbarna (approved):
Other than a minor comment on the name of a function, this seems to cover all the requirements we've listed in the associated issue. Not sure if there's a straightforward way to test it until we implement #3532 and mess around with it on `test`, but provided that it doesn't touch any of the functionality for general circular subscriptions, that shouldn't really be an issue

app/routes/circulars/circulars.server.ts
dakota002: Makes sense to me to avoid confusion

tylerbarna (approved):

PR #3727, opened 2026-07-22
Fix edit page back button

tylerbarna:
# Description

Currently, pressing the in-page navigation button on the edit request form kicks users back to the circulars archive page. This is counterintuitive, as users navigate to the edit page by pressing the edit button on a specific circular. The button has been updated to return users to the specific circular.

# Related Issue(s)

n/a

# Testing

1. Navigate to a specific circular in local development environment
2. Click on the edit button
3. On the edit page, press the back button on the page. This should now redirect back to the specific circular page

PR #3736, opened 2026-07-23
Handle case where new circular is navigated to from index for back button

tylerbarna:
# Description

Fixes the back button in the new circular form to handle when there is not a circular to navigate back to

# Related Issue(s)

https://github.com/nasa-gcn/gcn.nasa.gov/pull/3727#issuecomment-5061605625

# Testing

1. Navigate to archive locally
2. Press the new button 
3. Navigate back with the in page navigation
4. It should not return an error

### DIFF
--- a/app/routes/circulars.new.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.new.$circularId/CircularEditForm.tsx
@@ -305,7 +305,7 @@ export function CircularEditForm({
         </CollapsableInfo>
         <ButtonGroup>
           <Link
-            to={`/circulars/${circularId}${searchString}`}
+            to={`/circulars/${circularId ?? ''}${searchString}`}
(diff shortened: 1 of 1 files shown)


Courey (approved):

app/routes/circulars.new.$circularId/CircularEditForm.tsx
```diff
@@ -305,7 +305,7 @@ export function CircularEditForm({
         </CollapsableInfo>
         <ButtonGroup>
           <Link
-            to={`/circulars/${circularId}${searchString}`}
+            to={`/circulars/${circularId ?? ''}${searchString}`}
```
lpsinger, line 308: I am not sure if `/circulars/` and `/circulars` are backed by the same cache entry in CloudFront.

Please add Playwright tests.

PR #3738, opened 2026-07-24
Add event type to header with links to event type pages

### LINKED ISSUE #3528: Show associated eventType values on circular pages
# Description
List the tagged eventType values for a given circular 

# Acceptance criteria

- [ ] Show a list of the associated eventType values in the header of the circular page for an individual circular
- [ ] Values should be distinguishable as separate values
- [ ] Potentially make appearance something like the colored blocks for tags in github
- [ ] (optional) make tags clickable, which would direct to the archive with a filter for that eventType applied

tylerbarna:
# Description

This adds the event type as an item in the header. Each event type is also a hyperlink that directs users to the circulars tagged as that event type

Screenshot 2026-07-24 at 11 26 10

# Related Issue(s)

Resolves #3528 

# Testing

1. Navigate to a circular locally
2. The associated event types should be listed
3. They should be clickable and show the search results for that type


tylerbarna:
Messing around with a UI that uses the tag component. Having an odd thing where the text of all tags are automatically in all caps and I'm not seeing an obvious way to prevent that styling

Screenshot 2026-07-24 at 11 42 17

tylerbarna:
#3760 is related to this PR but does not necessarily have to be contained within it or resolved prior to merging this PR

tylerbarna:
Perhaps we should order it such that the eventType isn't stacked adjacent to the EventId so people don't accidentally hit the wrong thing 

image

tylerbarna:
This should be rebased once #3770 is merged and incorporate the human readable event type names using the singular form

PR #3742, opened 2026-07-26
Add rehype-sanitize to Markdown rendering pipeline

lpsinger:


dakota002 (approved):

medamineelhatimi:
Confirmed! the fix looks correct. Placing rehypeSanitize right after remarkRehype (and before rehypeClassNames/rehypeAstro) strips the javascript:/data: schemes from hrefs while preserving the custom classes/<data> nodes those later plugins add, so styling is unaffected. That closes the circular-body vector.
Thanks for the quick turnaround!

Issue #3732, opened 2026-07-23
Slugify event types


tylerbarna:
# Description
Per https://github.com/nasa-gcn/gcn.nasa.gov/pull/3728#issuecomment-5057754907, we should slugify the kafka subtopics such that users can visit a page and see all the circulars for a given event type

# Acceptance criteria

- [x] Event type topics should be slugified in a manner similar to event IDs
- [x] Users should be able to navigate to a url like gcn.nasa.gov/circulars/[eventType] and see all relevant circulars

tylerbarna:
Blocked by #3523 because that will establish the table stream structure

lpsinger:
This is not blocked.

PR #3747, opened 2026-07-29
Relocate Circulars Archive header into header file, rename files to better distinguish from individual circular routes

tylerbarna:
# Description

This moves the header content out of the route file for the circulars archive and into the existing header.tsx file, which has thus far been used just for the introductory paragraph rather than the whole header. This improves the readability of the route file by relocating ~150 lines of header code.

Once merged, this will make #3731 (or a revised, rebase PR) easier to review, as that PR was also doing this relocation.

# Related Issue(s)

Part of #3690 

# Testing

1. Navigate to circulars archive in local dev
2. All standard usage behavior should be unaffected


tylerbarna:
This will also be helpful for #3732 as there will be less code duplication with the primary index and the slugified index pages

lpsinger (commented):
Is this a refactoring only, i.e., only moves of existing code, and no functional changes?

tylerbarna:
> Is this a refactoring only, i.e., only moves of existing code, and no functional changes?

Yes, this is intended to be a refactor

lpsinger (approved):
This is fine with me but I would like someone else to test the functionality. Referring to @Courey and @dakota002.

dakota002 (approved):

dakota002 (approved):

PR #3751, opened 2026-07-31
Package lock bump

dakota002:


lpsinger:
What is the goal of this PR? This looks like version bumps. Was there something wrong with dependabot?

dakota002:
Not sure about dependabot. I was trying to dig into what change may have affected my circulars page, and in doing so I ran npm run clean, deleted caches and reinstalled npm packages. I just saw this was changed after doing so

lpsinger:
Does this fix anything that is broken?

PR #3752, opened 2026-07-31
Catch Unknown s3 error

dakota002:
There seems to be some Unknown Error being thrown in the function that checks for an object size. @Courey tried to replicate this locally, but did not see the error.
Other than the stack trace, the only content of the error is:

```
400: UnknownError
    at throwDefaultError (C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@smithy\smithy-client\dist-cjs\index.js:327:22)
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@smithy\smithy-client\dist-cjs\index.js:336:9
    at de_CommandError (C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@aws-sdk\client-s3\dist-cjs\index.js:5202:14)
    at processTicksAndRejections (node:internal/process/task_queues:104:5)
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@smithy\middleware-serde\dist-cjs\index.js:9:24
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@aws-sdk\middleware-sdk-s3\dist-cjs\index.js:488:18
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@smithy\middleware-retry\dist-cjs\index.js:300:46
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@aws-sdk\middleware-sdk-s3\dist-cjs\index.js:110:22
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@aws-sdk\middleware-sdk-s3\dist-cjs\index.js:137:14
    at C:\Users\ddutko\repos\gcn.nasa.gov\node_modules\@aws-sdk\middleware-logger\dist-cjs\index.js:33:22 
```
The error seems to come from:
```js
    const { ContentLength } = await s3.send(
      new HeadObjectCommand({
        Bucket,
        Key,
      })
    )
```

### DIFF
--- a/app/routes/circulars._archive.tsx
+++ b/app/routes/circulars._archive.tsx
@@ -34,7 +34,9 @@ async function getFileSize(format: string) {
       process.env.NODE_ENV !== 'production' &&
       e instanceof Object &&
       'name' in e &&
-      e.name === 'InvalidBucketName'
+      // TODO: This may have to do with a change to the error response from s3
+      // Look inot this and get a better fix, I noticed it on main as well
+      (e.name === 'InvalidBucketName' || e.name === '400')
     ) {
       console.warn('making up fake object size')
       return 1024 * 1024 * 50


lpsinger:
What was the error message (the content of the HTTP 400 response)?

dakota002:
Nothing useful:
```js
{
  '$fault': 'client',
  '$retryable': undefined,
  '$metadata': {
    httpStatusCode: 400,
    requestId: 'VGZ1JVGTV3GD10XT',
    extendedRequestId: 'syeZhOh8R5eLxNbhWkjog5X9VdqyjaCCKhj+HXQlaXf2O3JW/szTZ5StLTj8hsb91JM8CRA+KNAozGM7xRkFKj1AvjE6gHUP',
    cfId: undefined,
    attempts: 1,
    totalRetryDelay: 0
  }
}
```

dakota002:
And in the browser it just gets a 500:
image

lpsinger:
What does the request look like?

dakota002:
The s3 request? its the `HeadObjectCommand` listed in the description, called in `app/routes/circulars._archive.tsx`

Did something change recently with how we get the Bucket value? This is just strange because it hasn't been an issue, but now seems like its probably a windows path issue. 

I changed my Bucket from `C:\Users\ddutko\repos\gcn.nasa.gov\build\static` to `C:/Users/ddutko/repos/gcn.nasa.gov/build/static` and it worked. So maybe this is a non-issue, but still strange that it was working even just last week

lpsinger:
Does Architect have some local emulation of the S3 interface which is configured by the `ARC_STATIC_BUCKET` environment variable? If so, I would suggest submitting this as an issue to Architect and, if it is easy to do so, make a PR to fix it. If this is causing you a great inconvenience, then we could depend on your PR fork, as we do for several other Architect dependencies. If not, then you could address this by just documenting the path issue in our contributing section.

I'd rather not add an error handling special case that only affects sandbox mode but that would also potentially modify behavior on production.

PR #3764, opened 2026-08-05
Plugin search update

dakota002:
Bump the search plugin to try to fix post-deployment issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "@babel/preset-typescript": "^7.27.0",
         "@gitlab/svgs": "^3.83.0",
         "@nasa-gcn/architect-plugin-dynamodb-local": "^2.0.0",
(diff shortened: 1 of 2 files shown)


Courey (approved):

lpsinger (changes requested):
This introduced dependencies on version of @aws-sdk/* packages that do not match those that we use in production.

lpsinger:
> This introduced dependencies on version of @aws-sdk/* packages that do not match those that we use in production.

See https://github.com/nasa-gcn/architect-plugin-search/pull/354.

lpsinger (changes requested):
It's still installing a much wider mix of @aws-sdk/* package versions. Would you please try undoing the changes to package.json and package-lock.json, then bumping our package again, and running `npm i` again, to see if that changes the package resolution?

tylerbarna (approved):

PR #3780, opened 2026-08-14
Create dummy zendesk response for local development

tylerbarna:
# Description

To facilitate local development of circular moderation features, it would be useful to have a way to test locally without creating real zendesk tickets. This adds a feature flag that, when enabled, returns a dummy response when attempting to fetch zendesk credentials. As a result, any testing of edit requests and moderation will not actually submit to zendesk, but it is still possible to test the zendesk connection locally by removing the feature flag. 

# Related Issue(s)

n/a

# Testing

Note: you will need real values for `ZENDESK_CLIENT_ID` and `ZENDESK_CLIENT_SECRET` as well as access to the zendesk dashboard to properly test this.
1. Access https://nasa-gcn.zendesk.com
2. Add necessary zendesk info to `.env` along with `DISABLE_ZENDESK` feature flag
3. Start the local environment
4. Navigate to a circular and request an edit
5. The edit should be in the local moderation queue without creating a zendesk ticket
6. Accept the request. This should update the local instance of the circular
7. Kill local dev and remove the zendesk token and client from `.env` before restarting
8. Repeat steps 1 through 4. These should all still function and demonstrate the same behavior

### DIFF
--- a/app/lib/zendesk.server.ts
+++ b/app/lib/zendesk.server.ts
@@ -8,7 +8,7 @@
 import memoizee from 'memoizee'
 import { Issuer } from 'openid-client'
 
-import { getEnvOrDie } from './env.server'
+import { feature, getEnvOrDie } from './env.server'
 import { getBearerAuthHeaders } from './headers.server'
 import { throwForStatus } from './utils'
 
@@ -28,6 +28,7 @@ interface RequestComment {
 }
 
 const zendeskDomain = 'https://nasa-gcn.zendesk.com'
+const disableZendesk = feature('DISABLE_ZENDESK')
 
 const getAccessToken = memoizee(
   async () => {
@@ -53,6 +54,12 @@ const getAccessToken = memoizee(
 )
 
 async function fetchZendesk(url: string | URL, method: string, body: any) {
+  if (disableZendesk) {
+    return new Response(JSON.stringify({ request: { id: 1 } }), {
+      headers: { 'Content-Type': 'application/json' },
+    })
+  }
+
   const accessToken = await getAccessToken()
   const response = await fetch(url, {
     method,


tylerbarna:
There are also options like https://github.com/moltenform/visual-mock-zendesk-api and https://github.com/moltenform/visual-mock-zendesk-api that would provide a slightly more robust simulation of zendesk, but I've not come across a widely-used and maintained tool for this purpose in my initial research. Additionally, our zendesk usage is fairly basic and I'm unsure how much we would benefit from a more complicated solution compared to just using zendesk when the need arises

dakota002 (changes requested):
This is a good idea, but I think we should avoid adding a new env variable. 
Instead, it may be better to check whether or not the 2 that we are already using (line 40), `ZENDESK_CLIENT_ID` and `ZENDESK_CLIENT_SECRET`, are present.

tylerbarna:
> This is a good idea, but I think we should avoid adding a new env variable. Instead, it may be better to check whether or not the 2 that we are already using (line 40), `ZENDESK_CLIENT_ID` and `ZENDESK_CLIENT_SECRET`, are present.

I thought about that, but I didn't know if we wanted to alter the existing logic for the getEnvOrDie in the moderation tools

dakota002:
We should not change the `getEnvOrDie` function, but you may be able to replace it with `getEnvOrDieInProduction`. That way, you could try to get the var, and if it fails, you can just skip  locally

app/lib/zendesk.server.ts
```diff
@@ -28,17 +28,22 @@ interface RequestComment {
 }
 
 const zendeskDomain = 'https://nasa-gcn.zendesk.com'
+const disableZendesk = feature('DISABLE_ZENDESK')
```
dakota002, line 31: Please remove this part using a new env variable.

app/lib/zendesk.server.ts
```diff
@@ -53,7 +58,19 @@ const getAccessToken = memoizee(
 )
 
 async function fetchZendesk(url: string | URL, method: string, body: any) {
+  if (disableZendesk) {
+    return new Response(JSON.stringify({ request: { id: 1 } }), {
+      headers: { 'Content-Type': 'application/json' },
+    })
+  }
+
```
dakota002, line 66:
```suggestion
```
This can also be removed, as you have it doing the same thing below

dakota002 (changes requested):
Looks like you're implementing the genEnvOrDieInProduction function correctly, just need to remove code from your new Env variable

app/lib/zendesk.server.ts
lpsinger: There's no need to add a new environment variable. Please just use genEnvOrDieInProduction.

tylerbarna: yep, sorry, just forgot to cut out the old version

dakota002 (approved):

PR #3783, opened 2026-08-19
Add Errata to retraction keywords

tylerbarna:
# Description

This adds 'Errata' and its plural form as a keyword matched to the Retraction event type. I noticed there were just over [30 circulars](https://gcn.nasa.gov/circulars?view=index&query=subject%3A%22Errata%22+OR+subject%3A%22erratum%22&startDate=&endDate=&sort=circularID) that could accurately be described as retractions but would not be tagged as such

# Related Issue(s)

Updates #3518 

# Testing

n/a

### DIFF
--- a/app/routes/circulars/circulars.lib.ts
+++ b/app/routes/circulars/circulars.lib.ts
@@ -346,6 +346,7 @@ const eventTypeMatchers: Record<EventType, RegExp[]> = {
     /\bInvalid\b/i,
     /\bBogus\b/i,
     /\bFake\b/i,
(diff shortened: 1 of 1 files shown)


Vidushi-GitHub (approved):

jracusin:
Not all errata are actually retractions, sometimes they are modifications to some measured quantity (magnitude, position, etc.). I think these errata should receive their natural otherwise event type classification.

jracusin:
We have decided to not treat errata as retractions.

PR #3791, opened 2026-08-28
Update topicToFormatAndNoticeType to catch circulars  and heartbeats

dakota002:
Fixes an issue where topicToFormatAndNoticeType wouldn't return the correct values when selecting circulars and/or heartbeats, causing the UI to be blank

### DIFF
--- a/app/lib/utils.ts
+++ b/app/lib/utils.ts
@@ -21,7 +21,12 @@ export function topicToFormatAndNoticeType(topic: string): {
   noticeFormat: string
   noticeType: string
 } {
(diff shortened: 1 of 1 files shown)


tylerbarna (approved):

jracusin:
Is this an issue for gcn.notices.svom.voevent*? I think your logic would catch this, but it is not JSON.

dakota002:
Hm, I will check, probably though

tylerbarna (approved):

jracusin (approved):

PR #3793, opened 2026-09-01
Add admin page Open search 

### LINKED ISSUE #3681: Add functionality to repopulate all OpenSearch indices
# Description

In an emergency in which the OpenSearch instance becomes corrupted and we need to ask AWS Support to restore it from a backup, we should assume that the OpenSearch indices no longer accurately reflect the source of truth, which is DynamoDB.

Create an administrative function to reindex all of the data (Circulars, synonyms) from DynamoDB.

# Acceptance criteria

- A script or an admin endpoint.
- That, when activated, wipes all of the records in OpenSearch.
- And then posts/puts new records for all Circulars and Synonyms in DynamoDB (plus whatever data sources we are indexing).
- Such that the end result is the same as if the data had been indexed through the normal course of operations all along (i.e., as a result of indexing that is normally performed by our Table Streams lambdas etc.)
- Maximize reuse of our existing code for normal implementation of indexing of records under steady-state operations.

dakota002:
# Description

Adds:
 - New page for Admins: `/admin/opensearch`
 - New Table: `reindex_logs` to list last update trigger and status (COMPLETE/RUNNING)
 - `queues` pragma and associated lambda with longer timeout for potentially long running reindexing tasks

Updates:
 - Refactors reused `putIndex` code into `app/lib/opensearch.server.ts`
 - `esbuild.config.js` to include "queues" paths in build
 - (implemented in another PR, but the info here is still relevant) ~~Bumps `tsconfig.json` lib to "ES2024" to fix an issue where properties from synonym groups were not being correctly mapped to their types (see below)~~

A card view lists details about each index including health, status, and document count.
Each index can be manually reindexed, triggering a lambda defined at the `@queues` pragma section in `app.arc`. This will only UPDATE the existing index, but there is the potential for a future update where we create an entirely new index and make use of aliases, but this is out of the scope of this issue/PR. There are some TODO notes about that included in the code.

image

Clicking the button pops up a modal before confirming the trigger.

image

### ES2024 Bump info:
The specific code that was throwing errors is in `buildIndexData`:
```ts
items.push(
      ...Object.entries(
        Object.groupBy(synonyms, ({ synonymId }) => synonymId)
      ).flatMap(([synonymId, values]) => [
        {
          synonymId,
          eventIds: values?.map(({ eventId }) => eventId),
          slugs: values?.map(({ slug }) => slug),
          initialDate: min(values?.map(({ initialDate }) => initialDate)),
        },
      ])
    )
```
ES2022 wouldn't parse values correctly and doesn't support `Object.groupBy`,
image

 but ES2024 handles these correctly

# Related Issue(s)
Resolves #3681

# Testing

Before running, edit `sandbox-search.js` to filter out some items from one of the data sources. I used circulars, and just changed:
```ts
    ...circulars.flatMap((item) => [
      item,
    ]),
```
to 
```ts
    ...circulars
      .filter((x) => x.circularId % 2)
      .flatMap((item) => [
        item,
      ]),
```
to remove ~half the items.
- Go to the circulars page and see only the odd numbered circulars
- Sign in as "admin"
- go to `http://localhost:3333/admin/opensearch`
- click the reindex button and button in the popup for circulars index
- See the updated number of records and all missing even circulars back at circulars page


jracusin (approved):

app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,165 @@
...
+  health: 'green' | 'yellow' | 'red'
+  status: string
+  index: string
+  uuid: string
+  pri: number
+  rep: number
+  'docs.count': number
+  'docs.deleted': number
+  'store.size': string
+  'pri.store.size': string
```
lpsinger, line 33: This response type is defined in @opensearch-project/opensearch. Please use the existing types.

app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,165 @@
...
+  await queues.publish({
+    name: 'reindex-opensearch',
+    payload: { indexName },
+  })
+}
+
+export async function listIndexes() {
+  const client = await getSearchClient()
+  const response = await client.cat.indices({ format: 'json' })
+  const result = response.body as OpenSearchIndex[]
```
lpsinger, line 117: A type assertion should not be necessary. The response return value should be correctly typed. If not, then you may need to update @opensearch-project/opensearch.

app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,165 @@
...
+  const db = await tables()
+  for (const item of result) {
+    const reindex_log = await db.reindex_logs.get({ indexName: item.index })
+    if (reindex_log) {
+      item.reindexStatus = reindex_log.status
+      item.reindexTriggerTime = reindex_log.triggerTime
+    }
+  }
```
lpsinger, line 125: Execute these requests concurrently.

Are you scanning the entire table? If so, then please just do it in a single scan request.

app/lib/opensearch.server.ts
```diff
@@ -0,0 +1,165 @@
...
+        },
+      ])
+    )
+  } else {
+    const TableName = db.name(indexName)
+    const pages = paginateScan({ client }, { TableName })
+    for await (const page of pages) {
+      items.push(...(page.Items as unknown[]))
+    }
+  }
```
lpsinger, line 163: You read every single page into memory and then start issuing requests for each page. Is that necessary? Can you issue the requests as you scan each page?

lpsinger (changes requested):
Interesting design

app/lib/opensearch.server.ts
dakota002: Which one is it? The function is only defined with TResponse = Record<string, any

lpsinger: Then please try updating that package.

app/lib/opensearch.server.ts
dakota002: I likely can for circulars or any future tables if the data is 1-to-1 with the index, but the tough one is synonyms, since the searchable index requires mapping the data

dakota002:
@Courey I had to update opensearch, I verified that the search is still working for circulars and synonyms, but I wanted to ping you in case there are nuances i've missed

PR #3794, opened 2026-09-02
Clear search when switching archive views

### LINKED ISSUE #3769: Inconsistent behavior in archive when switching between circular and event view after submitting a search
# Current behavior
When switching between circulars and event view, the index does not load if a search has been initiated prior to switching.

# Expected behavior
The query value should be cleared and the clean const set to true so the index renders

# Steps to reproduce
1. Navigate to archive and submit a search
2. Without clearing the search, switch to event view
3. The page will not load the event view index
4. Switching back to circulars view will also not render the index
# Environment
prod

weewoo14:
# Use of AI Disclosure
GitHub CoPilot was used to help me navigate through the codebase and identify key areas to help resolve the issue. All physical code was handwritten by me.

# Description

When clicking on the buttons to navigate between Circular and Event view, I added an onClick to clear the inputQuery between the two views. In doing so, defaultValue had to be changed to value because defaultValue does not change during re-mounts of the component.

# Related Issue(s)

Resolves #3769 

# Testing

This was tested in local development.
1) Type a search query into the Circulars/Events search bar and press enter.
2) Navigate to the other view using its respective button.
3) The search bar should be cleared and the results should be displayed.

### DIFF
--- a/app/routes/circulars._archive._index/ArchiveHeader.tsx
+++ b/app/routes/circulars._archive._index/ArchiveHeader.tsx
@@ -136,7 +136,7 @@ export default function ArchiveHeader({
             id="query"
             name="query"
             type="search"
-            defaultValue={inputQuery}
+            value={inputQuery}
             placeholder={searchText}
             aria-describedby="searchHint"
             onChange={({ target: { form, value } }) => {
@@ -158,6 +158,9 @@ export default function ArchiveHeader({
             to={`/circulars?view=index&limit=${limit}`}
             preventScrollReset
             className={getSelection('index')}
+            onClick={() => {
+              setInputQuery('')
+            }}
           >
             <Icon.List role="presentation" />
             Circulars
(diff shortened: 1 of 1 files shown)


dakota002 (approved):
I tested locally and it behaves as expected. @tylerbarna if you also approve we can merge it

tylerbarna (approved):
Seems to solve the issue, thanks @weewoo14 !

lpsinger (commented):
Usually we don't like to have the _value_ of a field provided by a React state variable. I believe that we should revert this.

weewoo14:
Hi @lpsinger, I apologize for the inconvenience. 

Would you suggest using something like a useRef onto the TextInput element and then set it to empty (along with the inputQuery) in the onClick function? I believe this would still maintain the defaultValue parameter instead of having to use value.

lpsinger:
This is beside the point. I don’t think that `Clear search when switching archive views` is desirable behavior.

weewoo14:
Instead, when navigating between Circulars and Events view, if an ```inputQuery``` is present, it should mimic what a search query looks like, instead of what is presently hardcoded in the navigation between the two views?

PR #3796, opened 2026-09-02
Add hidden input field to prevent stripped eventType

### LINKED ISSUE #3792: eventTypes are stripped from edited Circulars while eventType is behind a feature flag
# Current behavior
When a Circular is edited while eventTypes front end is still behind a feature flag, the eventTypes get removed as part of the edit.

# Expected behavior
The eventTypes should remain in place even while the frontend is still under development.

# Steps to reproduce
A user submitted an edit to a Circular on production. I saw in the moderator interface that the eventTypes were removed. Screenshot shown below.

# Environment

tylerbarna:
# Description

This adds a hidden input field to the edit request page that is only set when the event type feature flag is not present (as is the case in production at the moment). This ensures existing event types are not stripped from circulars when unrelated edits are requested.

# Related Issue(s)

Fixes #3792

# Testing

1. Start up local _without_ `EVENTTYPE` as a feature flag in `.env`
2. Go to a circular; check the existing event types in the json
3.  Request any edit of the circular 
4. Go to the moderation queue
5. The event types should be preserved

### DIFF
--- a/app/routes/circulars.new.$circularId/CircularEditForm.tsx
+++ b/app/routes/circulars.new.$circularId/CircularEditForm.tsx
@@ -285,6 +285,15 @@ export function CircularEditForm({
                 setEventId(derivedEventId)
               }}
             />
+            {!showEventTypes &&
+              defaultEventTypes.map((eventType) => (
+                <input
+                  key={`preserved-event-type-${eventType}`}
+                  type="hidden"
+                  name="eventTypes"
+                  value={eventType}
+                />
+              ))}
             {showEventTypes && (
               <>
                 <fieldset className="margin-top-2">


dakota002 (approved):

jracusin (approved):

lpsinger:
This is a security risk. You cannot depend on the value of a hidden field to enforce an invariant, because the client can trivially submit a different request that has a different value for that field.

tylerbarna:
@lpsinger Could you expand on what you mean? The field isn't hidden in the moderation review, so any change would be apparent when reviewing the request

lpsinger:
Yes, but this is not supposed to be editable at all by users, is it?

dakota002:
> Yes, but this is not supposed to be editable at all by users, is it?

The only edit is a request for changes, not an actual change to the circular, so its no more or less valid than allowing a text box where users can request any change to the body. 

By keeping it hidden it just avoids making a request which look like it is trying to delete all the event types, when in reality they just don't have the UI for it anyway.

And @tylerbarna can correct me here if i am wrong, but this shouldn't have an effect on new circular submissions

lpsinger:
In the final design, is the event type supposed to be editable in change requests, at all?

dakota002:
Yes, it is already defined in the file, if you expand it down a few more lines you will see it in the section when the same feature flag is turned on

lpsinger:
Alright, never mind.

PR #3798, opened 2026-09-03
Install @nasa-gcn/architect-functions-search from NPM, not Dakota's git repo

lpsinger:


dakota002 (approved):
Yes this should have been caught when it was updated in the package.json too, must have missed it

PR #3799, opened 2026-09-03
Preserve all search fields when switching between index views

lpsinger:


dakota002:
This is just reverting the previous change correct? Can you make it such that when the user switches the view, it will trigger the new search? The current behavior is that if there is a query in circulars and you switch to events, it can maintain the query, but the search results area is just blank. 

It should show results for the same query against events instead

lpsinger:
> This is just reverting the previous change correct? Can you make it such that when the user switches the view, it will trigger the new search? The current behavior is that if there is a query in circulars and you switch to events, it can maintain the query, but the search results area is just blank.
> 
> It should show results for the same query against events instead

No, it reverts the previous change _and_ preserves all search fields, not just the search term.

dakota002 (approved):

PR #3804, opened 2026-09-10
Add Team management pages behind feature flag

dakota002:
# Description

This adds the functions and pages to allow users to manage their team memberships behind the "TEAMS" feature flag. 

This update does NOT add any functionality that would affect the Kafka broker.
I will have a separate PR that addresses App Clients/Token related stuff

### New link to the teams pages in the header UI

image

### Team page listing the teams a user belongs to and any pending invites for them to join a team

image

### View that a user with Admin permissions for their team sees

image

This allows the team admin to manage users permission level, add or remove users, and update the description field for the team.

image

### View that a non-admin user sees

image


app/routes/user.teams._index/route.tsx
```diff
@@ -0,0 +1,198 @@
...
+        New teams can only be created by an administrator as part of the Kafka
+        Producer onboarding process. If you do not see your team here, please{' '}
+        <Link to="/contact">contact us</Link>.
+      </p>
+      {userIsAdmin && (
+        <Link to="/user/teams/new" className="usa-button usa-button--outline">
+          Create New Team
+        </Link>
+      )}
+      <SegmentedCards>
```
tylerbarna, line 84: Could we get text for users that tell them explicitly that they are not currently in any teams?

app/routes/user.teams._index/route.tsx
```diff
@@ -0,0 +1,198 @@
...
+              View
+            </Link>
+          </ToolbarButtonGroup>
+        </div>
+      </Grid>
+    </>
+  )
+}
+
+function InviteCard({ invite }: { invite: TeamInvite & { teamName: string } }) {
```
tylerbarna, line 140: Same request as above; some explicit info that the user does not have any pending team invites

tylerbarna (changes requested):
Looks good other than those two small points of feedback

tylerbarna:
Oh, also, should we have some UI element that directs users to request team admin access if they believe they need it for the creation of new teams or will we be handling those hands-on enough that team admins will always be in contact with us for the onboarding process

